PR #39, opened 2022-12-07
Clean-up the remote_execution client for gRPC.

aloiscochard:
Also fix intrinsic properties for remote execution platforms, as previously the values provided were not compatible with the specification:
- https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/platform.md

### DIFF
--- a/app/buck2_core/src/fs/fs_util.rs
+++ b/app/buck2_core/src/fs/fs_util.rs
@@ -12,6 +12,7 @@
 use std::fs;
 use std::fs::File;
 use std::io;
+use std::io::Read;
 use std::io::Write;
 use std::ops::Deref;
 use std::path::Path;
@@ -332,6 +333,17 @@ pub fn remove_all<P: AsRef<Path>>(path: P) -> anyhow::Result<()> {
         return Ok(());
     }
     r
+}
+
+pub fn read_to_bytes<P: AsRef<Path>>(path: P) -> anyhow::Result<Vec<u8>> {
+    let path_display = P::as_ref(&path).display();
+    let metadata =
+        fs::metadata(&path).context(format!("When reading metadata `{}`", path_display))?;
+    let mut file = File::open(&path).context(format!("When openning file `{}`", path_display))?;
+    let mut buffer = vec![0; metadata.len() as usize];
+    file.read_exact(&mut buffer)
+        .with_context(|| format!("read_to_bytes({})", path_display))?;
+    Ok(buffer)
 }
 
 pub fn read_to_string<P: AsRef<Path>>(path: P) -> anyhow::Result<String> {
--- a/buck2_execute_impl/src/executors/re.rs
+++ b/buck2_execute_impl/src/executors/re.rs
@@ -62,14 +62,12 @@ pub enum RemoteExecutorError {
 impl ReExecutionPlatform {
     pub fn intrinsic_properties(&self) -> SortedMap<String, String> {
         match self {
-            Self::Linux => {
-                SortedMap::from_iter([("platform".to_owned(), "linux-remote-execution".to_owned())])
-            }
+            Self::Linux => SortedMap::from_iter([("OSFamily".to_owned(), "linux".to_owned())]),
             Self::MacOS { xcode_version } => SortedMap::from_iter([
                 ("platform".to_owned(), "mac".to_owned()),
                 ("subplatform".to_owned(), format!("xcode-{}", xcode_version)),
             ]),
-            Self::Windows => SortedMap::from_iter([("platform".to_owned(), "windows".to_owned())]),
+            Self::Windows => SortedMap::from_iter([("OSFamily".to_owned(), "windows".to_owned())]),
         }
     }
 }
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -7,12 +7,12 @@
  * of this source tree.
  */
 
-#![allow(unused_variables)] // Because a lot of these are stubbed out
 use std::future;
 use std::pin::Pin;
 use std::sync::Mutex;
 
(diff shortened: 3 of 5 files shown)


app/buck2_core/src/fs/fs_util.rs
```diff
@@ -334,6 +335,17 @@ pub fn remove_all<P: AsRef<Path>>(path: P) -> anyhow::Result<()> {
     r
 }
 
+pub fn read_to_bytes<P: AsRef<Path>>(path: P) -> anyhow::Result<Vec<u8>> {
```
ndmitchell, line 338: You might want to do `let path = P::as_ref(&path)` or similar on the first line, and then avoid repeating it below. But your choice.

buck2_execute_impl/src/executors/re.rs
```diff
@@ -63,13 +63,15 @@ impl ReExecutionPlatform {
     pub fn intrinsic_properties(&self) -> SortedMap<String, String> {
         match self {
             Self::Linux => {
-                SortedMap::from_iter([("platform".to_owned(), "linux-remote-execution".to_owned())])
+                SortedMap::from_iter([
+                    ("OSFamily".to_owned(), "linux".to_owned()),
+                ])
             }
             Self::MacOS { xcode_version } => SortedMap::from_iter([
-                ("platform".to_owned(), "mac".to_owned()),
```
ndmitchell, line 69: Can we ditch this from the diff

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -258,120 +210,126 @@ impl REClient {
...
+                                .with_context(|| "Tree digest not defined.")?,
+                        );
+                        Ok(TDirectory2 {
+                            path: output_directory.path,
+                            tree_digest: digest.clone(),
+                            root_directory_digest: digest,
+                            _dot_dot_default: (),
+                        })
+                    })
+                    .collect();
```
ndmitchell, line 270: Usually you would deal with the error immediately - so `.collect::<anyhow::Result<Vec<_>>>()?` instead of having an anyhow float around for a while.

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -380,16 +338,28 @@ impl REClient {
...
+            .unwrap_or_default()
+            .into_iter()
+            .map(|x| {
+                Ok(Request {
+                    digest: Some(tdigest_to(x.digest)),
+                    data: fs_util::read_to_bytes(&x.name)?,
+                    compressor: compressor::Value::Identity as i32,
+                })
+            })
+            .collect();
```
ndmitchell, line 362: Ditto - error out sooner.

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -380,16 +338,28 @@ impl REClient {
...
         use re_grpc_proto::google::rpc::Code;
 
-        let mut grpc_clients = self.grpc_clients.lock().await;
-        let client = grpc_clients.cas_client().await?;
+        let mut client = self.grpc_clients.cas_client.clone();
+
+        let files_with_digest: anyhow::Result<Vec<Request>> = request
+            .files_with_digest
+            .unwrap_or_default()
+            .into_iter()
```
ndmitchell, line 354: You might want to use `gazebo::prelude::*` which then gives you `into_try_map` which should simplify and avoid the need for type annotations

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -461,22 +415,22 @@ impl REClient {
...
     pub async fn upload_blob(
         &self,
-        blob: Vec<u8>,
-        metadata: RemoteExecutionMetadata,
+        _blob: Vec<u8>,
+        _metadata: RemoteExecutionMetadata,
     ) -> anyhow::Result<TDigest> {
-        stub("upload_blob")
+        // TODO(aloiscochard)
+        Err(anyhow::anyhow!("Not implemented"))
```
ndmitchell, line 422: Say "RE upload_blob" in the message somewhere, so when it fails this way, we know what to fix.

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -495,37 +449,26 @@ impl REClient {
...
+
+        let blobs: anyhow::Result<Vec<InlinedDigestWithStatus>> =
+            response.into_inner().responses.into_try_map(|r| {
+                Ok(InlinedDigestWithStatus {
+                    // TODO(aloiscochard): Here we should check if r.status is ok!
+                    digest: tdigest_from(r.digest.with_context(|| "Response digest not found.")?),
+                    status: tstatus_ok(),
+                    blob: r.data,
+                })
+            });
```
ndmitchell, line 461: ? on this line rather than later

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -559,8 +502,7 @@ impl REClient {
     }
 
     pub fn get_network_stats(&self) -> anyhow::Result<NetworkStatisticsResponse> {
-        // TODO(aloiscochard): Is that unwrap safe? could we avoid it by making the caller async?
-        let state = self.state.lock().unwrap();
+        let state = self.state.lock().unwrap_or_else(|e| e.into_inner());
```
ndmitchell, line 505: What does this do? The error goes into_inner? I think just unwrap here is what we'd recommend, no "or_else"

aloiscochard: As far as I understand, this code means that if the locking fails, we take the data and return it anyway (https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.into_inner).

This is probably the best thing we should do (instead of crashing), as we are only reading the state in this case and not writing to it.

_Please note that change was suggested during previous code review feedback._

PR #42, opened 2022-12-09
Fix protoc version typo

philipcraig:


ndmitchell:
From reading the protoc changelog 3.12 looked sufficient (I just linked to 3.21 since it was the latest). Do you believe we require higher than that?

philipcraig:
agreed, closing

PR #46, opened 2022-12-16
Extend example

shayne-fletcher:


arlyon:
TY for this! You may consider using the `assert_output` rule to make sure the output of the rules is as you expect as well

https://github.com/facebookincubator/buck2/blob/main/examples/prelude/python/hello_world/BUILD

PR #58, opened 2022-12-21
Conan integration to import third-party C/C++ dependencies

aherrmann:
This PR adds support for importing third-party C/C++ dependencies provided by the [Conan package manager](https://docs.conan.io/en/latest/introduction.html).

Detailed motivation and design are contained in the module doctring of the main toolchain module under `prelude/toolchains/conan/defs.bzl`. In brief, some C/C++ third-party dependencies are difficult to import into a Buck2 project, e.g. the current self-hosted build of Buck2 [vendors a binary openssl](https://github.com/facebookincubator/buck2/pull/26) and [skips jemalloc](https://github.com/facebookincubator/buck2/pull/31#issue-1416513233) for that reason.

The approach proposed here is to instead use the existing package build definitions for the Conan package manager and build these third-party dependencies with Conan. This approach is similar to [how rules_haskell integrates with Stack and Cabal](https://github.com/tweag/rules_haskell/blob/84cb0c5210116632819224620168f2ddd8eb1a76/haskell/cabal.bzl#L2404) to import third-party Haskell dependencies, or how [JVM packages are imported](https://github.com/bazelbuild/rules_jvm_external) with the help of Coursier in Bazel.

Conan was chosen for its relatively large community package index, compatibility with Linux, MacOS, and Windows, configurability and extensibility, and its support for cross-compilation. Alternatively, [vcpkg](https://vcpkg.io/en/index.html) was considered. However, its choice of CMake as implementation and configuration language makes it harder to integrate. Other interesting comparisons can be found [here](https://matgomes.com/vcpkg-vs-conan-for-cpp/) and [here](https://github.com/XRPLF/rippled/issues/4125).

This integration aims to leave Buck2 in charge of the overall build, when which targets are built, and which configurations and toolchains are used.

**Code map:**

- `prelude/toolchains/conan/defs.bzl` the main Starlark module for the Conan integration. Defines the relevant rules.
- `prelude/toolchains/conan/*.py` Python scripts implementing the various build actions.
- `prelude/toolchains/conan/buckler/` a Conan extension that defines a generator to generate Buck2 import targets for Conan built packages.
- `examples/prelude/toolchains/BUILD` an example Conan toolchain configuration.
- `examples/prelude/cpp/conan/` an example Conan use-case.

This PR adds tests of the Conan example to the Linux and MacOS CI pipelines. Windows support will be added separately.

**Remaining work for future PRs**

- [ ]  Better cross-platform support.
    Use Conan lock-file bundles to define cross-platform import targets.
- [ ]  Windows support in particular.
    Currently builds fail on long-path issues, Conan provides workarounds for these issues that need to be investigated.
- [ ]  Import Buck2’s third-party C/C++ dependencies with Conan.
- [ ]  Cross-compilation support.
    Conan has builtin cross-compilation support that needs to be made use of.
- [ ]  Test some more complex packages.
    Builds of some more complex packages failed when attempted with a `zig cc` toolchain, e.g. `libmysqlclient` fails to find `libresolv.so`. Investigate these issues. An example setup with the Zig toolchain can be found [here](https://github.com/aherrmann/buck2-conan-toolchain-example).


arlyon:
Excellent, looking forward to reading through this :)

prelude/toolchains/conan/buckler/conanfile.py
```diff
@@ -0,0 +1,240 @@
...
+        self.requires = [_Requirement.parse(req) for req in component_info.requires]
+
+    def generate(self):
+        """Generate Buck2 target definitions for the component."""
+        name = _Requirement(None, self.component_name).to_name(self.package_name)
+        deps = [req.to_label(self.package_name) for req in self.requires]
+        return """\
+
+conan_component(
+    name = {name},
```
arlyon, line 135: you could consider using the `!r` [conversion flag](https://peps.python.org/pep-3101/#explicit-conversion-flag) for these

prelude/toolchains/conan/buckler/conanfile.py
```diff
@@ -0,0 +1,240 @@
...
+)
+""".format(
+            name = repr(self.name),
+            components = repr({
+                name: _Requirement(None, name).to_label(self.name)
+                for name in self.components.keys()
+            }),
+            visibility = repr(["PUBLIC"]) if self.public else repr([]))
+
+        for component in getattr(self, "components", {}).values():
```
arlyon, line 205: what is the requirement for getattr here?

prelude/toolchains/conan/defs.bzl
```diff
@@ -0,0 +1,791 @@
...
+            shared_lib = extract_shared_libs[lib][0]
+        else:
+            shared_lib = None
+        if lib in static_libs:
+            static_lib = extract_static_libs[lib][0]
+        else:
+            static_lib = None
+        native.prebuilt_cxx_library(
+            name = name,
+            deps = deps,  # TODO[AH] Do we need exported_deps?
```
arlyon, line 328: is exported_deps not needed for transitive deps to work correctly? @ndmitchell please confirm

prelude/toolchains/conan/defs.bzl
```diff
@@ -0,0 +1,791 @@
...
+    represents only the `crypto` component.
+    """
+    native.cxx_library(
+        name = "_bundle_" + name,
+        exported_deps = components.values(),
+    )
+    _conan_cxx_libraries(
+        name = name,
+        main = ":_bundle_" + name,
+        components = components,
```
arlyon, line 399: I love we support this. Some context about components for others: https://docs.conan.io/en/latest/creating_packages/package_information.html#using-components

prelude/toolchains/conan/defs.bzl
```diff
@@ -0,0 +1,791 @@
...
+        DefaultInfo(default_outputs = [output]),
+        ConanProfileInfo(config = output, inputs = content.hidden(args_inputs)),
+    ]
+
+conan_profile = rule(
+    impl = _conan_profile_impl,
+    attrs = {
+        "arch": attrs.string(doc = "The target architecture"),
+        "os": attrs.string(doc = "The target operating system"),
+        "build_type": attrs.string(doc = "The Conan build-type, e.g. Release or Debug"),
```
arlyon, line 710: Nice!

prelude/toolchains/conan/conan_common.py
```diff
@@ -0,0 +1,172 @@
...
+    if trace_log is not None:
+        # Enable Conan debug trace.
+        env["CONAN_TRACE_FILE"] = os.path.abspath(trace_log)
+
+    # TODO[AH] Enable Conan revisions for reproducibility
+    # Enable Conan revisions for reproducibility
+    #env["CONAN_REVISIONS_ENABLED"] = "1"
+
+    # Prevent over-allocation.
+    env["CONAN_CPU_COUNT"] = "1"
```
arlyon, line 155: I take it this means conan build libraries single-threaded

examples/prelude/cpp/conan/BUILD
```diff
@@ -0,0 +1,54 @@
...
+    conanfile = "conanfile.txt",
+    visibility = ["//cpp/conan/import:"],
+)
+
+lock_generate(
+    name = "lock-generate",
+    lockfile = ":lock",
+)
+
+# TODO[AH] Prevent double build of packages.
```
arlyon, line 21: This is interesting. I take it this cost only appears when actually generating the lockfile itself? That _seems_ like an ok trade-off for now.

prelude/toolchains/conan/conan_init.py
```diff
@@ -0,0 +1,58 @@
...
+
+def conan_profile(
+        conan,
+        user_home,
+        trace_log):
+    env = conan_common.conan_env(
+            user_home=user_home,
+            trace_log=trace_log)
+
+    # TODO[AH] Allow users to define additional remotes.
```
arlyon, line 18: A conan_remote rule would be cool. You could pass it in similar to the profile similar to deps.

arlyon:
It would be cool to consider this option as well: https://docs.conan.io/en/1.46/reference/conanfile/tools/system/package_manager.html#configuration-properties-that-affect-how-system-package-managers-are-invoked

arlyon:
This is great. I would love to see the package component limitation lifted though it seems like simply a matter of implementing it. Otherwise, the code is succinct and clear. The python scripts seem solid, and I've managed to run `:update` for boost and glm, glfw, and glad and the output file looks correct! It's great that components are included (and I love the syntax) and deps are mapped, though I couldn't test the transitive deps due to the aforementioned issue.

I will let Neil dispense buck2 related wisdom but I am excited that this is ready to land.

https://github.com/facebookincubator/buck2/blob/76c18df1fe5a0ef8e611779eb080bdcc606b1595/prelude/toolchains/conan/defs.bzl#L344

prelude/toolchains/conan/buckler/conanfile.py
aherrmann: Good point, done.

prelude/toolchains/conan/buckler/conanfile.py
aherrmann: Good catch, that's a left over from an earlier state where that attribute didn't exist for packages without components. Fixed.

prelude/toolchains/conan/conan_common.py
aherrmann: > I take it this means conan build libraries single-threaded.

The intention is to avoid over-allocation due to both Buck2 and Conan parallizing their builds. 

> Edit: make is still invoked w/ -j16 on my machine (16 threads)

Thanks for testing! IIUC Conan recipes still [need to actively use this setting](https://docs.conan.io/en/latest/reference/env_vars.html#conan-cpu-count):
> Conan recipes can use the cpu_count() tool to build the library using more than one core.

Which package were you building there?

examples/prelude/cpp/conan/BUILD
aherrmann: Correct, well, when generating the targets file, i.e. when invoking `conan install --generator ...`. But, yes, that's the idea.

prelude/toolchains/conan/conan_init.py
aherrmann: Yes, a `conan_remote` rule sounds like a good way to do this.

aherrmann:
> [...] I've managed to run `:update` for boost and glm, glfw, and glad [...] though I couldn't test the transitive deps due to the aforementioned issue.

Thanks for testing!

> This is great. I would love to see the package component limitation lifted though it seems like simply a matter of implementing it.

I looked into these Conan central and tried to build boost. It looks like the issue is that boost contains some header-only components for which `len(libs) == 0`. I've added a case to handle this and update the PR. With that I was able to build boost locally.

For the multiple library components case, i.e. `len(libs) > 1`, I haven't found an example yet to test this on, so haven't attempted implementing it, yet. Are you aware of such a Conan package? I looked into `ffmpeg`, which [seems to have been one of the packages motivating components](https://github.com/conan-io/conan/issues/3173), but even that seems to use single library components by now as far as I can tell.

shim/toolchain.bzl
```diff
@@ -68,7 +68,7 @@ def _cxx_toolchain(ctx):
                 nm = RunInfo(args = ["nm"]),
                 # objcopy = ctx.attrs.objcopy_for_shared_library_interface[RunInfo],
                 # ranlib = ctx.attrs.ranlib[RunInfo],
-                ranlib = RunInfo(args = ["raninfo"]),
+                ranlib = RunInfo(args = ["ranlib"]),
```
ndmitchell, line 71: We no longer use shim/toolchain, since we now use the system toolchains in the examples

prelude/toolchains/cxx/zig/defs.bzl
```diff
@@ -295,7 +295,7 @@ def _cxx_zig_toolchain_impl(ctx: "context") -> ["provider"]:
     return [ctx.attrs.distribution[DefaultInfo]] + cxx_toolchain_infos(
         platform_name = dist.arch,
         c_compiler_info = CCompilerInfo(
-            compiler = RunInfo(args = cmd_args([zig, "cc"])),
+            compiler = RunInfo(args = cmd_args([zig, "cc"], delimiter = " ")),
```
ndmitchell, line 298: Why is this necessary?

prelude/toolchains/cxx.bzl
```diff
@@ -41,12 +41,12 @@ def _system_cxx_toolchain_impl(ctx):
             linker_info = LinkerInfo(
                 linker = RunInfo(args = ["clang++"]),
                 linker_flags = ["-fuse-ld=lld"],
-                archiver = RunInfo(args = ["ar", "rcs"]),
+                archiver = RunInfo(args = ["ar"]),
```
ndmitchell, line 44: Why is this necessary?

prelude/toolchains/cxx.bzl
```diff
@@ -41,12 +41,12 @@ def _system_cxx_toolchain_impl(ctx):
...
                 archiver_type = archiver_type,
                 type = linker_type,
                 link_binaries_locally = True,
                 archive_objects_locally = True,
-                use_archiver_flags = False,
+                use_archiver_flags = True,
```
ndmitchell, line 49: Why is this necessary?

.circleci/config.yml
```diff
@@ -209,6 +212,8 @@ jobs:
           command: |
             cd examples/prelude
             cmd.exe /c mklink /D prelude C:\Users\circleci\project\prelude
+            # Skip Conan tests on Windows. TODO[AH] Implement Windows support.
+            Remove-Item -Recurse -Force cpp/conan
```
ndmitchell, line 216: We usually do this with host_info if's in the prelude - see where we do it in other places.

prelude/toolchains/conan/TARGETS
```diff
@@ -0,0 +1,57 @@
+native.export_file(
```
ndmitchell, line 1: This file should be named TARGETS.v2, since otherwise internal stuff gets confused

ndmitchell (commented):
Looks good - just a few notes/questions.

prelude/toolchains/conan/defs.bzl
cjhopman: For c++, exported_deps is required when your dependents need to include your dependency's headers. This could be because your headers that your dependencies include then further directly include ones from your dependency's or if you just otherwise expect your dependents to do so (like if your interfaces are going to require things from your dependency's). Often this isn't required.

Now, when looking at doing something like we're doing here, we're kinda limited by only being able to understand what the underlying specification encodes (and how its other interpretations behave). If the specification doesn't distinguish between public and private dependencies, it is almost certainly treating them all as public (it could be treating them as private, but then everything we sort of need to individually specify their transitive deps and that is rather gross). In that case, I'd recommend encoding them as public deps (i.e. exported_deps) in buck2 as well.

prelude/toolchains/conan/conan_common.py
cjhopman: > The intention is to avoid over-allocation due to both Buck2 and Conan parallizing their builds

There is a weight on actions that you could try tweaking here. It can basically be thought of as telling buck that the action requires that many cores. it's also been used to limit parallelism due to memory requirements, not cores, but its modeled with just the one limit (trivia note: that's actually different than buck1 which had ability to configure several different weights for different resources)

prelude/toolchains/conan/defs.bzl
aherrmann: @cjhopman Thanks for the context! I've changed the code to set `exported_deps` instead.

prelude/toolchains/conan/conan_common.py
aherrmann: @cjhopman Thanks! That looks like a useful feature to address this. Given that seemingly not all Conan recipes make use of the `CONAN_CPU_COUNT` knob, and that the load a parallelized package build will also depend on the size of the package, this does indeed seem like something that will require a bit of tuning. In light of that I'd prefer to leave this to a follow-up PR. I've added a TODO comment to the code to keep track of it.

prelude/toolchains/cxx/zig/defs.bzl
aherrmann: The Conan toolchain configures Conan to use the C/C++ toolchain provided by Buck2. It does so by [setting the relevant variables](https://github.com/facebookincubator/buck2/pull/58/files#diff-636b3a34913c4a58b72558b8ef0f48f441101379984f3828fc7697b0c88c98c9R704) in the Conan profile. Some build tools assume that these contain a single path, instead of commands with arguments, so the toolchain [generates wrapper scripts](https://github.com/facebookincubator/buck2/pull/58/files#diff-636b3a34913c4a58b72558b8ef0f48f441101379984f3828fc7697b0c88c98c9R656-R667) that execute the required command. Without this change the command would be written on multiple lines despite the explicit [`delimiter = " "` at a higher level](https://github.com/facebookincubator/buck2/pull/58/files#diff-636b3a34913c4a58b72558b8ef0f48f441101379984f3828fc7697b0c88c98c9R639). Perhaps there is a better way to do this? As the comment there indicates I also tried setting `quote = "shell"`, but that had other unwanted effects.

prelude/toolchains/cxx.bzl
aherrmann: Defining `"rcs"` in the `archiver`'s `RunInfo` breaks with Conan when building a package from source. The problem is that we configure `ar rcs` as the archiver programm and the build system of the Conan package again passes similar flags, e.g. `cs`, to the archiver, resulting in a command line of the form `ar rcs cs ...`. This fails with an error of the form:

```
ar: libz.a: No such file or directory
```
As explained in https://github.com/madler/zlib/issues/588#issuecomment-992694506 .

If we set `ar` alone as the archiver then this problem goes away. However, then we need to make sure that Buck2's archive actions pass the correct arguments to `ar`, which is why this sets `use_archiver_flags = True`, see [here](https://github.com/facebookincubator/buck2/blob/6ea6a5e1cc836fc057e5d28e24d1c89a229259b8/prelude/cxx/archive.bzl#L17-L18).

Perhaps the core issue is that the `archiver` field conflates the archiver binary and the flags to be passed to the archiver in an action to create a static archive. If instead it followed the pattern of e.g. `CCompilerInfo` and there were separate `archiver` and `archiver_flags` fields then we could extract the archiver even in a configuration that explicitly configures `archiver_flags` and sets `use_archiver_flags = False`.

See also related [commit message](https://github.com/facebookincubator/buck2/pull/58/commits/5f04d41640730baedaeed98cbb247da567bba389).

prelude/toolchains/cxx.bzl
aherrmann: See response to https://github.com/facebookincubator/buck2/pull/58/files#r1061631371.

.circleci/config.yml
aherrmann: The difficulty here is that some of those targets are generated, namely those in `examples/prelude/cpp/conan/import/BUILD`. See also the internal discussion we had on this from 21/12/2022. Removing the folder seemed like the best option at the time. But, I'm happy to do something else if there's a better option.

prelude/toolchains/conan/TARGETS
aherrmann: Done.

yeswalrus:
Just curious, has this PR been abandoned? This functionality would be tremendously useful for me.

f0rmiga:
@aherrmann Interesting work here. Is there something I could do to help push it across the finish line?

aherrmann:
Thank you for the positive feedback!

In terms of implementation work this PR is ready. However, the main branch has moved on a fair bit since then, so, it will need a rebase and likely a fair bit of work to adapt to any relevant changes to main since.

@ndmitchell IIRC last we spoke about this PR the intention was to get this merged. Apart from a rebase, what would it take to get this in?

ndmitchell:
Let's rebase, and do the minimum required to merge it. Basically focus on the changes to the existing toolchain. If users want the new code, and it's off to the side so can't have any negative effects, we should err on the side of merging it.

f0rmiga:
That's amazing. Thanks for the quick reply.

aherrmann:
@ndmitchell I've rebased this on main and made the necessary fixes to get CI green again. I no longer seem to have push access to update the branch of this PR, so I've opened #440 instead. PTAL

PR #63, opened 2023-01-01
clippy fixes

philipcraig:


ndmitchell:
Not always a fan of all clippy warnings, but these all look like sensible ones to include. Thanks!

philipcraig:
> Not always a fan of all clippy warnings, but these all look like sensible ones to include. Thanks!

I skipped the silly ones!

PR #67, opened 2023-01-03
Improve setup.sh

shayne-fletcher:
Two small improvements:
- remove `set -euo pipefail` from the top of the script since it plays poorly with zsh (and after consideration have reached the conclusion was never correct in that the script is intended to be sourced so "exiting" the shell even in the event of an error is not the right thing to do);
- add `--set-switch` to the `opam env` command in order to silence the notice "`[NOTE] To make opam select the switch default in the current shell, add --set-switch or set OPAMSWITCH`"

### DIFF
--- a/examples/prelude/README.md
+++ b/examples/prelude/README.md
@@ -17,9 +17,21 @@ ln -s $(realpath ../../prelude) prelude
 
 If you are using any other prelude directory, make sure the prelude points to it in `.buckconfig`
 
+On Linux and macOS, an alternative to symlinking the prelude manually in this way is to `source setup.sh` (see the next section for details).
+
+## Support for building the example OCaml targets
+
+The information in this section is (at this time) Linux and macOS specific.
+
+The commands in `setup.sh` assume an [opam](https://opam.ocaml.org/) installation.
+
+Some of these commands are intended to affect the current shell thus `setup.sh` is to be invoked via the `source` builtin e.g `source setup.sh`.
+
+In addition to symlinking the prelude directory [as described above](#add-in-prelude-into-the-project), sourcing `setup.sh` activates the 'default' opam switch and creates symlinks in the 'third-party/opam' directory. These symlinks are neccessary to support building the example OCaml targets. If any of the symlinks are found to already exist, they will not be overwritten.
+
 ## Sample commands
 
-**_NOTE:_** These are currently only supported on Linux
+**_NOTE:_** These commands are currently only supported on Linux and macOS.
 
 ```sh
 $BUCK2 build //...
--- a/examples/prelude/setup.sh
+++ b/examples/prelude/setup.sh
@@ -1,4 +1,3 @@
-#!/bin/bash
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 #
 # This source code is licensed under both the MIT license found in the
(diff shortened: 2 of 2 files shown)


ndmitchell:
Why is this script running with zsh? We have a bin/bash shebang on the first line? Those fail lines are quite important as otherwise when something goes wrong we don't fail and get an error message.

Alternative is just to make this script a Python script instead, which tends to eliminate all the other issues.

shayne-fletcher:
> Why is this script running with zsh? We have a bin/bash shebang on the first line?

the default (standard, expected) shell on macOS is zsh. it is built on top of bash and mostly bash compatible. iiuc, the shebang is used by the kernel to determine a script's interpreter if the script were invoked as in `./setup.sh`. here the intended usage is to pass it as argument to the `source` command as in `source ./setup.sh` and therefore i expect it is being executed by a zsh interpreter (i could be wrong).

> Those fail lines are quite important as otherwise when something goes wrong we don't fail and get an error message.

pretty confident that if something goes wrong you'll get an error message. the difference is that execution won't halt on the first error.

> 
> Alternative is just to make this script a Python script instead, which tends to eliminate all the other issues.

this is a reasonable idea and i did consider it (assuming one can affect the executing shell's environment from a Python script which i imagine one can).

the current behavior i'm seeing on my mac however is that on completion of the script the shell from which it is invoked is terminated. that for sure is not what we want and is a result of the `-e`. i spent an hour or so looking for info about this but haven't turned up anything satisfactory about what to do about it. maybe `set -o errrexit nounset pipefail` is an equivalent that works with both but in truth, i figured it would be ok to just remove the `set` line because the chances of a failure in the commands that follow are super remote or, if something were to fail, they won't fail silently as i explain above.

shayne-fletcher:
> > Why is this script running with zsh? We have a bin/bash shebang on the first line?
> 
> the default (standard, expected) shell on macOS is zsh. it is built on top of bash and mostly bash compatible. iiuc, the shebang is used by the kernel to determine a script's interpreter if the script were invoked as in `./setup.sh`. here the intended usage is to pass it as argument to the `source` command as in `source ./setup.sh` and therefore i expect it is being executed by a zsh interpreter (i could be wrong).

in fact, by that reasoning (that the script is intended to be "sourced"), the shebang should probably not be there.

shayne-fletcher:
> the shebang should probably not be there

i'm going to make that modification in this PR too while you make your call as to whether you want these changes or not.

shayne-fletcher:
> > the shebang should probably not be there
> 
> i'm going to make that modification in this PR too while you make your call as to whether you want these changes or not.

i note that the script does not have it's executable bit set (preventing the potentially confusing `./setup.sh` command from working already as we'd hope) so that's good, nothing to do there.

shayne-fletcher:
@ndmitchell 

> Can we have a block of doc comment at the top of this script saying that we expect you source it, and it is required to get OCaml working? Maybe it should be referenced in the adjacent README as well?

no worries. i've implemented both of these suggestions. if this is now looking ok, so that i might import it, might you put a stamp on this please? thanks.

examples/prelude/README.md
```diff
@@ -17,9 +17,21 @@ ln -s $(realpath ../../prelude) prelude
 
 If you are using any other prelude directory, make sure the prelude points to it in `.buckconfig`
 
+On Linux and macOS, an alternative to symlinking the prelude manually in this way is to `source ./setup.sh` (see the next section for details).
+
+## Support for building the example OCaml targets
+
+The information in this section is (at this time) Linux and macOS specific.
+
+The commands in 'setup.sh' assume an [opam](https://opam.ocaml.org/) installation.
```
ndmitchell, line 26: Usually you would do backticks to make it teletype in markdown

examples/prelude/setup.sh
```diff
@@ -1,23 +1,27 @@
...
 #
 # This source code is licensed under both the MIT license found in the
 # LICENSE-MIT file in the root directory of this source tree and the Apache
 # License, Version 2.0 found in the LICENSE-APACHE file in the root directory
 # of this source tree.
 
-set -euo pipefail
+# The commands in this script are to be executed in the current shell
+# and so invoke it via the builtin 'source' command e.g. `source
+# ./setup.sh`. It assumes an opam installation. In addition to
```
ndmitchell, line 10: Normally the `./` bit is only to run things - just `source setup.sh` should be sufficient

ndmitchell (commented):
Looks good (few nits, but import and I'll stamp)

examples/prelude/README.md
shayne-fletcher: ok

shayne-fletcher:
> Looks good (few nits, but import and I'll stamp)

stand by, i'll fix up those infelicities and push again. ping you when it's done (2 mins)

shayne-fletcher:
@ndmitchell , ok this is ready to go.

shayne-fletcher:
> Looks good (few nits, but import and I'll stamp)

i don't know why, but i'm not presented an option to import. is it because it needs an approving PR review here first?

PR #114, opened 2023-04-07
Fix typo

qrilka:
I hope I understood the intent of the text correctly, the current version is broken because of a hanging round bracket


ndmitchell:
Thanks for spotting this. Actually, I think this is a case of losing the suffix of the sentence. Let me put up a diff that fills it out :)

ndmitchell:
This got broken in 68919d3ea63584fceb807dff439cc8461973fa34. I'll put it back.

ndmitchell:
Fixed in f9755e80863a1b9ca4c1358575a7a317a4707919. Thanks for the report!

Issue #120, opened 2023-04-07
Building Buck on Illumos


steveklabnik:
Hey there!  I'm excited about Buck, and want to give it a try generally. I think it *could* solve some problems that I have at work. But we use illumos for various things, which means sometimes stuff gets weird. I personally use Windows, so I'm not an expert on illumos either, so I am ultimately trying to get a program I don't know running on a platform I don't know well; fun times :)

So, here's where I'm at at the time of putting this together:

https://github.com/steveklabnik/buck2/commit/199021929656cb23449c6ee5e271a84b50f38a07

Let's talk about how I got there.

So, the first thing that goes wrong is:

```
error: could not compile `fs2` due to previous error
warning: build failed, waiting for other jobs to finish...
error[E0432]: unresolved import `self::os`
 --> /home/sklabnik/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.19.1/src/features.rs:2:15
  |
2 | pub use self::os::*;
  |               ^^ could not find `os` in `self`

```

Okay, illumos is a weird unix, but I thought nix supported it. Let's check in on nix: https://crates.io/crates/nix

Oh, the latest release is 0.26, not 0.19. Let's see if we can update it:

```console
$ cargo update -p nix
error: There are multiple `nix` packages in your project, and the specification `nix` is ambiguous.
Please re-run this command with `-p <spec>` where `<spec>` is one of the following:
  nix@0.19.1
  nix@0.22.3
  nix@0.23.2
  nix@0.24.3
```

Oh my. 

Turns out 0.19 is being brought in by the dependency on fs2. I haven't heard of that package before. 

https://github.com/danburkert/fs2-rs last commit in 2018. Uh oh. Here's a comment from 3 weeks ago on a PR: https://github.com/danburkert/fs2-rs/pull/42#issuecomment-1470397839

> for what it is worth, I submitted a PR to fs4 and they fixed it. I haven't seen any activity on fs2 for >5 years, but fs4 is a maintained fork that provides all of the same functionality.

Okay! Let's sub in fs4! If we do that... 

```console
error[E0425]: cannot find function `statvfs` in module `rustix::fs`
  --> src/unix.rs:55:23
   |
55 |     match rustix::fs::statvfs(path.as_ref()) {
   |                       ^^^^^^^ not found in `rustix::fs`
   |
help: consider importing this function
   |
45 | use crate::statvfs;
   |
help: if you import `statvfs`, refer to it directly
   |
[...]

PR #121, opened 2023-04-07
Don't immediately unwrap protoc vendored paths

steveklabnik:
The intention of using environment variables is to override these paths. With the code as written, this will unconditionally panic if protoc_bin_vendored doesn't know where the protobuff compiler/includes are, even if we're using an environment variable to set these paths.

With this change, we only unwrap when actually attempting to use the path from `protoc_bin_vendored`.

Found as part of #120.


steveklabnik:
(I signed the CLA, I'm assuming the bot will be along shortly, no rush of course.)

PR #144, opened 2023-04-11
Generate a .gitignore with buck2 init --git

### LINKED ISSUE #140: `buck2 init --git` may not be producing a correct .gitignore
By default, no `.gitignore` is produced, but I believe that you'd want to produce one for `buck-out`, because I don't believe (I may be wrong!) that you don't want to include this directory into version control.

steveklabnik:
This creates a gitignore that will ignore the default output directory.

Fixes #140

I didn't see a test where adding this behavior to it would make sense, if you'd like a test for this, happy to add one, and/or make any other changes you'd prefer.

### DIFF
--- a/app/buck2_client/src/commands/init.rs
+++ b/app/buck2_client/src/commands/init.rs
@@ -210,6 +210,14 @@ fn set_up_prelude(path: &Path, git: bool) -> anyhow::Result<()> {
     Ok(())
 }
 
+fn set_up_gitignore(path: &Path) -> anyhow::Result<()> {
+    let mut gitignore = std::fs::File::create(path.join(".gitignore"))?;
+
+    writeln!(gitignore, "buck-out")?;
+
+    Ok(())
+}
+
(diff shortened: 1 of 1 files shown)


app/buck2_client/src/commands/init.rs
```diff
@@ -210,6 +210,14 @@ fn set_up_prelude(path: &Path, git: bool) -> anyhow::Result<()> {
     Ok(())
 }
 
+fn set_up_gitignore(path: &Path) -> anyhow::Result<()> {
+    let mut gitignore = std::fs::File::create(path.join(".gitignore"))?;
+
+    writeln!(gitignore, "buck-out")?;
```
ndmitchell, line 216: Should this be `/buck-out` as that will scope it more tightly?

themarwhal:
Hi @steveklabnik! Thank you for taking the time to make the change! It just happens that I made a very similar change a few hours ago (https://github.com/facebook/buck2/commit/18bcd6fe77fbf400107087e5a41f3077c3929339) . The issue should be resolved now.

steveklabnik:
Ah awesome! Thanks!

PR #145, opened 2023-04-11
chore: add boilerplate for hacking on buck2 with nix + direnv

thoughtpolice:
I've been using this to hack on and build `buck2` for a while now, since I often don't use `rustup` (and actually not using NixOS -- but Nix-On-Ubuntu-On-WSL2?!?!!) and figured it might be worth putting upstream at this point, with the official announcement being made. This will at least let any prospective users who have Nix get compiling real quick. It will also allow some future improvements to `flake.nix`, but...

Please note (especially lurking Nix users): this only allows you to use Nix to provision a development environment, one which is capable of running `cargo build`. There is not yet a "nix package" (AKA thing that installs inside `/nix/store`) for buck2 in this PR. I have one as part of `buck2-nix`, but it's out of date because some recent changes in mid/late February started causing a build failure. I'll open up a separate PR for that later and discuss it.

Please see the commit message for a longer version.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,4 @@
+if ! has nix_direnv_version || ! nix_direnv_version 2.2.1; then
+    source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/2.2.1/direnvrc" "sha256-zelF0vLbEl5uaqrfIzbgNzJWGmLzCmYAkInj/LNxvKs="
+fi
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /target/
 Cargo.lock
 buck-out
+/.direnv
 
 # symlinks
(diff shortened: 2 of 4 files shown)


ndmitchell:
We have relatively little ability to support this going forward - but it doesn't look like it will take much maintenance, so might as well merge.

thoughtpolice:
FWIW, I'd be perfectly OK with leaving this as-is, for now. Actually, I could also refactor it to probably exist in another repository, too, I think, without harming the UX too much. So don't feel obligated to do this at the moment.

I do need to file an issue about `prost` though, since that's preventing a Nix build of a more recent version of buck2, but I might just go ahead and open a literal bug report.

ndmitchell:
I see no reason not to land this - it doesn't hurt.

PR #148, opened 2023-04-12
Typo: `rust fmt` -> `rustfmt`

mgeisler:
The space looks like a typo to me.


ndmitchell:
Seems the project itself uses `Rustfmt` all the time, so we should follow.

PR #167, opened 2023-04-14
Make rustc configurable

bigfootjon:
This makes the built-in system rust toolchain configurable, matching `system_python_toolchain`:

https://github.com/facebook/buck2/blob/main/prelude/toolchains/python.bzl#L75

I need this personally because I want to use a toolchain from `rustup` instead of my real system toolchain, but I don't want to manually configure the toolchain


bigfootjon:
I'll just use `rustup override`

PR #172, opened 2023-04-16
Allow to change compiler_type for system_cxx_toolchain

VestniK:
I'm playing with buck2 using codebase which can't be compiled with clang due to difference in C++20 featureset supported in clang and gcc. I've found that the there is only one piece is missing to be able to define gcc cxx toolchain using unmodified prelude. This PR adds this missing thing allowing to use system gcc compiler with the following toolchain declaration:
```
system_cxx_toolchain(
    name = "cxx",
    compiler_type = "gcc",
    compiler = "gcc",
    cxx_compiler = "g++",
    linker = "g++",
    cxx_flags = ["-std=c++20", "-O3"],
    visibility = ["PUBLIC"],
)
```

### DIFF
--- a/prelude/toolchains/cxx.bzl
+++ b/prelude/toolchains/cxx.bzl
@@ -75,21 +75,21 @@ def _system_cxx_toolchain_impl(ctx):
                 compiler = RunInfo(args = [ctx.attrs.cxx_compiler]),
                 preprocessor_flags = [],
                 compiler_flags = ctx.attrs.cxx_flags,
-                compiler_type = "clang",  # one of CxxToolProviderType
+                compiler_type = ctx.attrs.compiler_type,  # one of CxxToolProviderType
             ),
             c_compiler_info = CCompilerInfo(
                 compiler = RunInfo(args = [ctx.attrs.compiler]),
                 preprocessor_flags = [],
                 compiler_flags = ctx.attrs.c_flags,
-                compiler_type = "clang",  # one of CxxToolProviderType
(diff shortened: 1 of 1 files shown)


themarwhal:
Hi @VestniK! Thanks for contributing :) It looks like CI is failing, but it should be fixed past https://github.com/facebook/buck2/commit/a6068af003b55dde4aa24a016c53c52767efd52d , do you mind rebasing?

VestniK:
Only one test fails after rebase and it fails on due to different line endings on windows and linux:
```
---- cli::new::test_render stdout ----
thread 'cli::new::test_render' panicked at 'assertion failed: `(left == right)`
  left: `"load(\"@fbcode_macros//build_defs:rust_library.bzl\", \"rust_library\")\n\nrust_library(\n    name = \"a-rust-library\",\n    srcs = glob([\"src/**/*.rs\"]),\n    deps = []\n)"`,
 right: `"load(\"@fbcode_macros//build_defs:rust_library.bzl\", \"rust_library\")\r\n\r\nrust_library(\r\n    name = \"a-rust-library\",\r\n    srcs = glob([\"src/**/*.rs\"]),\r\n    deps = []\r\n)"`', integrations\rust-project\src\cli\new.rs:174:5
```
The only difference between strings is in line endings style. I can fix it but it looks lke such fix will be irrelevant change in this PR.

themarwhal:
@VestniK, your change has been merged! Thanks again!

Issue #156, opened 2023-04-13
Remote configuration is not reloaded with build, must restart the server.


stagnation:
Hi!

I was tinkering a little bit with the engflow example and a local BuildBarn deployment (https://github.com/buildbarn/bb-deployments)*.
At first I just got errors about the url missing a scheme. 
The same with every modification to the config.
But building the same project with a debug build from source I got a new error.
So it must be that the remote execution configuration is read and cached by the server.

```
WARN buck2_execute::re::client: Failed to connect to RE, retrying after sleeping 4 seconds: Error {
    context: "RE: creating client",
    source: Error {
        context: "Error creating CAS client",
        source: Error {
            context: "Error connecting",
            source: tonic::transport::Error(
                Transport,
                hyper::Error(
                    Connect,
                    "invalid URL, scheme is missing",
                ),
            ),
        },
    },
}
```

I expect `buck2 build //...` to pick up changes.
But restarting the server works fine: `buck2 kill; buck2 build //...`.

---------------------------------

Forgive the nested questions, if you wish I can open a new issue.

Now I have a very minimal config that is used, but does not work for `CorruptMessage` error-reasons.
```
~/tas/buc/exa/rem/engflow $ cat .buckconfig
[repositories]
root = .
prelude = prelude

[buck2]
# Adjust as needed.
digest_algorithms = SHA256

[buck2_re_client]
# Adjust as needed
action_cache_address = grpc://127.0.0.1:8991
engine_address = grpc://127.0.0.1:8991
cas_address = grpc://127.0.0.1:8991

[build]
  execution_platforms = root//platforms:platforms
```
The error:
```
WARN buck2_execute::re::client: Failed to connect to RE, retrying after sleeping 3 seconds: Error {
    context: "RE: creating client",
    source: Error {
        context: "Error creating CAS client",
        source: Error {
            context: "Error connecting",
            source: tonic::transport::Error(
                Transport,
                hyper::Error(
                    Connect,
                    Custom {
                        kind: InvalidData,
                        error: CorruptMessage,
                    },
                ),
            ),
        },
    },
}
```
Grpc service listing:
```
$ grpcurl -plaintext 127.0.0.1:8991 list
build.bazel.remote.execution.v2.ActionCache
build.bazel.remote.execution.v2.Capabilities
build.bazel.remote.execution.v2.ContentAddressableStorage
build.bazel.remote.execution.v2.Execution
google.bytestream.ByteStream
grpc.health.v1.Health
[...]

PR #174, opened 2023-04-17
buck2: don't add tls_config when URI does not call for it

krallin:
Summary:

Like it says in the title. Right now Tonic will try to connect using TLS if you pass `tls_config` even if the URI clearly does not call for it.

See: https://github.com/facebook/buck2/issues/156

Test Plan:

We have tests internally for the HTTPS case. The non-HTTPS scenario is unfortunately one for which we lack test coverage right now, but let's at least unbreak this to start.

I did test that this works locally with non-TLS Build Barn.

### DIFF
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -55,6 +55,7 @@ use tonic::transport::channel::ClientTlsConfig;
 use tonic::transport::Certificate;
 use tonic::transport::Channel;
 use tonic::transport::Identity;
+use tonic::transport::Uri;
 
 use crate::error::*;
 use crate::metadata::*;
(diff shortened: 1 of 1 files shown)


stagnation:
Thank you :) I was waiting for this to push my example of a BuildBarn configuration. With notes about how to use the bb-deployments repository. And some caveats about the not-yet plumbed instance name prefix. I'll rebase onto your example and submit a diff, with mostly documentation.

krallin:
Ah, i actually submitted a separate diff for a build barn example already
(since I was testing with it), but please feel free to suggest improvements
to it, that’d be welcome!

stagnation:
Yeah I saw that :) it's great to have an Buildbarn example: here is the documentation PR: https://github.com/facebook/buck2/pull/179

PR #177, opened 2023-04-18
Set relative command path

dezren39:
If one moves the BUCK file from the initial root location, then buck build fails to find the out file.

This change was needed when I moved the example into a subdirectory because I wanted to keep it. It may be that I'm misunderstanding how to use buck2. By merging this change, it may reduce the number of issues brand new users encounter if they `buck2 init --git` and then move the buck file into a subdirectory immediately afterwards. (For reasons such as to be the first 'app' of a multi-app repo, which all generally reuse the prelude and toolchains.) I want to be able to run `buck2 build //init-app:hello_world` from any child directory of `.buckroot` and it works.


dezren39:
it must have been something local. i will close this. i've been using windows or wsl and this may have added to my issues, retrying in a fresh instance again today and this didnt pose the same issues. sorry i'm not sure what i did wrong 😅

ndmitchell:
If you can reproduce the server getting stuck let us know. I don't know what WSL/Windows does to things like inotify and other notification protocols, but it wouldn't surprise me if something there has corner cases.

PR #179, opened 2023-04-18
Improve buildbarn documentation

stagnation:
Improve the example with instructions for how to deploy buildbarn locally with bb-deploymens.
And a workaround for instance name prefixes.

I will write an issue to track that, and we should refer to that issue in the example README.

### DIFF
--- a/examples/remote_execution/buildbarn/.buckconfig
+++ b/examples/remote_execution/buildbarn/.buckconfig
@@ -2,14 +2,12 @@
 root = .
 prelude = prelude
 
-[buck2]
-digest_algorithms = SHA256
-
 [buck2_re_client]
-action_cache_address = http://localhost:8980
-engine_address = http://localhost:8980
-cas_address = http://localhost:8980
+action_cache_address = grpc://localhost:8980
+engine_address = grpc://localhost:8980
+cas_address = grpc://localhost:8980
 tls = false
+instance_name = fuse
 
 [build]
   execution_platforms = root//platforms:platforms
--- a/examples/remote_execution/buildbarn/README.md
+++ b/examples/remote_execution/buildbarn/README.md
@@ -1,30 +1,63 @@
-## Remote execution integration with Build Barn
+## Remote execution integration with Buildbarn
 
-This project provides a small example of what a project that utilizes [Build Barn](https://github.com/buildbarn) for RE might look like.
+This project provides a small example of what a project that utilizes [Buildbarn](https://github.com/buildbarn).
 
 In this document, we will go over the key configs used in this setup.
+Using a local docker-compose deployment from the [example deployment repo](https://github.com/buildbarn/bb-deployments).
+If you already have a Buildbarn deployment you can skip that.
+
+### Deploy a local Buildbarn
+
+```
+... $ git clone https://github.com/buildbarn/bb-deployments
+... $ cd bb-deployments
+.../bb-deployments $ cd docker-compose
+.../bb-deployments/docker-compose $ ./run.sh
+```
+
+This uses `docker-compose` to spin up the required Buildbarn services.
+Using FUSE based workers, those are generally the fastest as they can load action files on demand
+and avoids the overhead of setting up the full input root up front.
+In practice many actions do not read all the files in the input root.
+
+If you do not want FUSE workers you can instead switch to hardlinking workers
+The example deployments have two worker kinds "fuse", and "hardlinking",
+you can see the queues in the Buildbarn scheduler, http://localhost:7982.
+
+```
+Buildbarn Scheduler
+
+...
+
+Instance name                                    Platform properties                                    Size  Timeo
+   prefix                                                                                               class
+
+"fuse"        OSFamily="Linux" container-image="docker://ghcr.io/catthehacker/                          0     ∞
+              ubuntu:act-22.04@sha256:5f9c35c25db1d51a8ddaae5c0ba8d3c163c5e9a4a6cc97acd409ac7eae239448"
+"hardlinking" OSFamily="Linux" container-image="docker://ghcr.io/catthehacker/                          0     ∞
+              ubuntu:act-22.04@sha256:5f9c35c25db1d51a8ddaae5c0ba8d3c163c5e9a4a6cc97acd409ac7eae239448"
+```
+
+More information is available in the [repo](https://github.com/buildbarn/bb-deployments).
 
 ### Relevant configs in .buckconfig
 
-First, the Build Barn endpoint and certificate should be configured as the following:
+First, the Buildbarn endpoint should be configured as the following:
 
 ```ini
 [buck2_re_client]
-engine_address       = $BB_ENDPOINT
-action_cache_address = $BB_ENDPOINT
-cas_address          = $BB_ENDPOINT
+engine_address       = grpc://localhost:8980
+action_cache_address = grpc://localhost:8980
+cas_address          = grpc://localhost:8980
+tls                  = false
+
+# Select the instance name, "fuse" or "hardlinking", without quotes
+instance_name        = fuse
 ```
 
-If you're using the Docker-compose setup, the endpoint would be `http://localhost:8980` for all 3 of those.
-
-Additionally, set the `digest_algorithm` config to `SHA256`.
-```ini
-[buck2]
-digest_algorithms = SHA256
-```
+TLS is not used in this example.
 
 ### Relevant configs in `ExecutionPlatformInfo`
 
-Build Barn takes in a Docker image and OSFamily  in its RE properties to select a worker.
-
-The execution platform used in this project `root//platforms:platforms` do so.
+Buildbarn takes in a Docker image and `OSFamily` in its RE properties to select a worker.
+This is configured in `root//platforms:platforms`.
--- a/examples/remote_execution/buildbarn/platforms/defs.bzl
+++ b/examples/remote_execution/buildbarn/platforms/defs.bzl
@@ -18,7 +18,7 @@ def _platforms(ctx):
             local_enabled = True,
             remote_enabled = True,
             use_limited_hybrid = True,
-            # Set those up based on what workers you've registered with Build Barn.
+            # Set those up based on what workers you've registered with Buildbarn.
             remote_execution_properties = {
                 "OSFamily": "Linux",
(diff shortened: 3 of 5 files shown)


examples/remote_execution/buildbarn/README.md
```diff
@@ -1,30 +1,51 @@
...
 
 In this document, we will go over the key configs used in this setup.
+Using a local docker-compose deployment from the [example deployment repo](https://github.com/buildbarn/bb-deployments)
 
-### Relevant configs in .buckconfig
+### Deploy Buildbarn
 
-First, the Build Barn endpoint and certificate should be configured as the following:
+This is mostly a turn-key solution, with just one modification.
+Buck2 does not yet support instance-name prefixes for different workers [xxx issue link](buck2/issues/XXX).
```
stagnation, line 11: TODO

examples/remote_execution/buildbarn/README.md
```diff
@@ -1,30 +1,51 @@
-## Remote execution integration with Build Barn
+## Remote execution integration with Buildbarn
 
-This project provides a small example of what a project that utilizes [Build Barn](https://github.com/buildbarn) for RE might look like.
+This project provides a small example of what a project that utilizes [Buildbarn](https://github.com/buildbarn).
 
 In this document, we will go over the key configs used in this setup.
+Using a local docker-compose deployment from the [example deployment repo](https://github.com/buildbarn/bb-deployments)
```
stagnation, line 6: We may want some connective tissue here, to say that with a ready Buildbarn deployment the deploy section can be skipped. But for those who want to test remote execution, I think it's good to have instructions for that.
We can make it more succinct after we fix the instance-name prefix issue. But I would like to help people avoid that speed bump.

thoughtpolice:
I wonder if this should be on the homepage on http://buck2.build, right? Then it can just point to this directory as an example. But the website should have a direct example of how to configure this. It was definitely non-obvious to me at first!

examples/remote_execution/buildbarn/.buckconfig
```diff
@@ -2,13 +2,10 @@
...
 prelude = prelude
 
-[buck2]
-digest_algorithms = SHA256
-
 [buck2_re_client]
-action_cache_address = http://localhost:8980
-engine_address = http://localhost:8980
-cas_address = http://localhost:8980
+action_cache_address = grpc://localhost:8980
```
krallin, line 6: I'd rather keep this as `http://`. 

Neither of those are actually valid as per [the GRPC spec for URIs](https://github.com/grpc/grpc/blob/master/doc/naming.md), but `grpc://` appears to be more used for when TLS is desired, so I intend to change #174 to respect that (at which point this would break)

examples/remote_execution/buildbarn/README.md
```diff
@@ -1,30 +1,51 @@
...
 
-### Relevant configs in .buckconfig
+### Deploy Buildbarn
 
-First, the Build Barn endpoint and certificate should be configured as the following:
+This is mostly a turn-key solution, with just one modification.
```
krallin, line 10: Let's maybe just remove this and wait for #181 to get merged.

examples/remote_execution/buildbarn/README.md
```diff
@@ -1,30 +1,51 @@
...
-[buck2]
-digest_algorithms = SHA256
+[buck2_re_client]
+engine_address       = grpc://localhost:8980
+action_cache_address = grpc://localhost:8980
+cas_address          = grpc://localhost:8980
 ```
 
-### Relevant configs in `ExecutionPlatformInfo`
+No TLS is used in this example (for history see this [issue](https://github.com/facebook/buck2/issues/156)).
```
krallin, line 46: What do we think the user should learn from this history? Can we either omit this entirely, or make that clearer?

examples/remote_execution/buildbarn/README.md
```diff
@@ -1,30 +1,51 @@
...
 
-Build Barn takes in a Docker image and OSFamily  in its RE properties to select a worker.
+### Relevant configs in `ExecutionPlatformInfo`
 
-The execution platform used in this project `root//platforms:platforms` do so.
+The Buildbarn workers take in a Docker image as their execution platform.
```
krallin, line 50: I think using "execution platform" here is a little confusing. To my knowledge, this is not a term Build Barn uses, but it *is* a term Buck2 uses, and that's not what it means in Buck2. I think "worker" did seem more appropriate (since that *is* terminology Build Barn uses)

examples/remote_execution/buildbarn/platforms/defs.bzl
```diff
@@ -21,6 +21,8 @@ def _platforms(ctx):
             # Set those up based on what workers you've registered with Build Barn.
             remote_execution_properties = {
                 "OSFamily": "Linux",
+                # This could be any Docker image.
```
krallin, line 24: I don't think it can actually be any Docker image? It's got to be the one you did start your Build Barn cluster with, no?

examples/remote_execution/buildbarn/.buckconfig
```diff
@@ -2,13 +2,10 @@
 root = .
 prelude = prelude
 
-[buck2]
-digest_algorithms = SHA256
```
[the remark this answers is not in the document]
krallin: Possibly, though I think it's a bit unlikely that we'd make this a priority (we've got no use for it internally, but also that would be a pretty substantial overhaul of how we start up things, since we need this config very early on, which is not true of the RE client).

krallin (changes requested):

examples/remote_execution/buildbarn/.buckconfig
krallin: Actually, let's keep this as-is, and add a separate config field for TLS. I'll update #156 with that.

examples/remote_execution/buildbarn/README.md
stagnation: Yes! Great point. I originally wrote this last week, when we had no timeline for a solution.
And then an in-review Pull Request would be a good location for other confused developers to find the information. I planned to remove this after the solution, so the final artifact would not have it.

I'm fine with wating for 181, so I can remove the mention from here.

examples/remote_execution/buildbarn/README.md
stagnation: Yepp, with your work to solve it we do not need any mention of the problems.

I will remove it.

examples/remote_execution/buildbarn/README.md
stagnation: yes!

examples/remote_execution/buildbarn/platforms/defs.bzl
stagnation: No, you are right, and your base example did not have the comment. I just copy-pasted it from the Engflow example.

I'll remove it.

examples/remote_execution/buildbarn/.buckconfig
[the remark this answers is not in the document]
meroton-benjamin: At the moment that's true, there's not much gain in consuming GetCapabilities only to look at which digest function to use when SHA256 is pretty much universally supported.

But looking forward a little, the RE api recently adopted [SHA256TREE](https://github.com/bazelbuild/remote-apis/pull/235) as a digest function which is a precursor to add new functionality to the remote api [ConcatenateBlobs and SplitBlobs](https://github.com/bazelbuild/remote-apis/pull/233). Those will allow partial transfer of larger blobs and ultimately become a more efficient replacement of bytestream.write.

There is of course no hurry, I just thought it might be worth poking the project on a trajectory that meshes well with how the protocol is moving.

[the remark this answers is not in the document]
krallin: > But looking forward a little, the RE api recently adopted https://github.com/bazelbuild/remote-apis/pull/235 as a digest function which is a precursor to add new functionality to the remote api https://github.com/bazelbuild/remote-apis/pull/233. Those will allow partial transfer of larger blobs and ultimately become a more efficient replacement of bytestream.write.

Sure, my point there is primarily that actually configuring the hashing scheme you use is rather the trivial bit, there's a lot more to actually implementing describing what you're doing than inferring a configuration :)

examples/remote_execution/buildbarn/.buckconfig
```diff
@@ -2,13 +2,10 @@
...
-
 [buck2_re_client]
-action_cache_address = http://localhost:8980
-engine_address = http://localhost:8980
-cas_address = http://localhost:8980
+action_cache_address = grpc://localhost:8980
+engine_address = grpc://localhost:8980
+cas_address = grpc://localhost:8980
```
krallin, line 8: Let's add `tls = false` here

examples/remote_execution/buildbarn/README.md
krallin: That landed, can you add it?

krallin (changes requested):

examples/remote_execution/buildbarn/.buckconfig
stagnation: Yes, though you already did in the commit that added it.

examples/remote_execution/buildbarn/README.md
stagnation: Yes!

examples/remote_execution/buildbarn/.buckconfig
```diff
@@ -2,14 +2,12 @@
...
 tls = false
+instance_name = fuse
```
stagnation, line 10: For the record: inline toml comments are parsed as part of the string.
I presume this is a toml file, but you can disregard this if inline comments are not meant to work.

```
Internal error (stage: remote_call_error): Remote Execution Error (GRPC-SESSION-ID): RE: execution with digest 9a8d
b5f5c0ba0b07866989dafa06b9914c25483dfedba2b0f53d06251191bdfc:142: Failed to start remote execution: status: Unavail
able, message: "No workers exist for instance name prefix \"fuse # or \\\"hardlinking\\\"\" platform {\"properties\
":[{\"name\":\"OSFamily\",\"value\":\"Linux\"},{\"name\":\"container-image\",\"value\":\"docker://ghcr.io/catthehac
ker/ubuntu:act-22.04@sha256:5f9c35c25db1d51a8ddaae5c0ba8d3c163c5e9a4a6cc97acd409ac7eae239448\"}]}", details: [], me
tadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```

Issue #180, opened 2023-04-18
Instance Name Prefix is not plumbed through


stagnation:
Hi

I recently set up buck2 to build with Buildbarn through its[ deployment repository](https://github.com/buildbarn/bb-deployments), and found that the workers with instance name prefixes could not be reached. There is just a hardcoded empty string set in the REAPI calls.

```
remote_execution/oss/re_grpc/src/client.rs
64:const INSTANCE_NAME: &str = "";
284:                    instance_name: INSTANCE_NAME.into(),
319:            instance_name: INSTANCE_NAME.into(),
443:            instance_name: INSTANCE_NAME.into(),
663:        instance_name: INSTANCE_NAME.into(),
```

This is a valuable configuration to keep and target different worker (queues) with a single Buildbarn scheduler.
And this is demonstrated in the example deployment.

For buck2 to successfully schedule action we need to remove that from the deployment at the moment.
See the caveats here: https://github.com/facebook/buck2/pull/179/files#diff-7de77dbbbc29297627446111dde92536735bc8598b81a10520d8ce02fffcb9a7R10-R11
(which will back-link to this discussion)

I think this would be a natural config for the toolchain.
I did some rudimentary hacking to have the toolchain accept this, but I did not proceed to actually plumb it into `re_grpc/src/client.rs`.

So a user could specify the instance name.

```
+++ b/examples/remote_execution/buildbarn/platforms/defs.bzl
@@ -20,6 +20,7 @@ def _platforms(ctx):
         executor_config = CommandExecutorConfig(
             local_enabled = True,
             remote_enabled = True,
+            instance_name_prefix = "fuse",  # or "hardlinking"
             use_limited_hybrid = True,
             remote_execution_properties = {
                 "container-image": image,
```
-------------------------------------------------------------------------------------------------
The prefix is shown in Buildbarn's scheduler 
```
Buildbarn Scheduler                                                                                                 

Build queue
Total number of operations:  0
Platform queues


Instance name                                    Platform properties                                    Size  Timeout   Queued
   prefix                                                                                               class         operatio

"fuse"        OSFamily="Linux" container-image="docker://ghcr.io/catthehacker/                          0     ∞       0
              ubuntu:act-22.04@sha256:5f9c35c25db1d51a8ddaae5c0ba8d3c163c5e9a4a6cc97acd409ac7eae239448"
"hardlinking" OSFamily="Linux" container-image="docker://ghcr.io/catthehacker/                          0     ∞       0
[...]

Issue #170, opened 2023-04-15
Materializer chokes on uploading action cache artifacts that exceed default 4MB gRPC limit


thoughtpolice:
If you follow along along with the saga in https://github.com/thoughtpolice/buck2-nix/issues/12, I have almost got remote execution working on buildbarn with Nix! Except that the tarball I download seems to be a tad bit too large (formatted for legibility in the github UI):

```
Action failed: prelude//toolchains:nixpkgs-overlay-rust (download_tarball https://github.com/oxalica/rust-overlay/archive/1373567ffd13719f6b7522737b010bfc514d49b4.tar.gz)

Internal error (stage: download): action_digest=c8b86439a7e637e7c7d98cf2addf516a771ed82f:93:
Failed to declare in materializer: status: InvalidArgument, message:

"Attempted to read a total of at least 23508808 bytes, while a maximum of 16777216 bytes is permitted", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```

I can see that the action itself completed successfully in the buildbarn action cache UI, and that the file downloaded (given in the error above, from `rust-overlay`) is 23508808 bytes in size:

<img width="1185" alt="screenshot-buildbarn" src="https://user-images.githubusercontent.com/3416/232237290-2c9f1434-a7f5-4933-8f34-602a129cf5ea.png">

I'm not sure how to approach this. Maybe it's a gRPC limit? The `CacheCapabilities` don't exactly specify anything about this:

```
austin@GANON:~/src/reapi-server$ grpcurl -plaintext '127.0.0.1:8980' build.bazel.remote.execution.v2.Capabilities/GetCapabilities
{
  "cacheCapabilities": {
    "digestFunctions": [
      "MD5",
      "SHA1",
      "SHA256",
      "SHA256TREE",
      "SHA384",
      "SHA512"
    ],
    "actionCacheUpdateCapabilities": {
      "updateEnabled": true
    },
    "symlinkAbsolutePathStrategy": "ALLOWED"
  },
  "executionCapabilities": {
    "digestFunction": "SHA256",
    "execEnabled": true,
    "executionPriorityCapabilities": {
      "priorities": [
        {
          "minPriority": -2147483648,
          "maxPriority": 2147483647
        }
      ]
    },
    "digestFunctions": [
      "MD5",
      "SHA1",
      "SHA256",
      "SHA256TREE",
      "SHA384",
      "SHA512"
    ]
  },
  "lowApiVersion": {
    "major": 2
[...]

PR #181, opened 2023-04-18
Fixes for remote execution

thoughtpolice:
These three commits add three features:

- `buck2_re_client.address` to serve as a "catch all" address in `.buckconfig`, useful to make the config shorter and less redundant
- instance name support #180
- an upload path that doesn't choke on the 4MB gRPC decode limit in e.g. buildbarn (#170

I don't know if these commits should be made into a stacked PR instead to make the Meta import easier, but let me know and I can do that.

The first two patches are quite simple. The third commit is a bit more detailed. TL;DR it probably performs worse on small files and file batches for now, but with these changes I can at least get to the point where I can upload `nixpkgs.tar.gz` and the un-tar'd files as well! (~160MB, 34k files.) So I think that's good for now, but I'll keep prodding at it as I find things...

### DIFF
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -61,8 +61,6 @@ use crate::metadata::*;
 use crate::request::*;
 use crate::response::*;
 
-const INSTANCE_NAME: &str = "";
-
 fn tdigest_to(tdigest: TDigest) -> Digest {
     Digest {
         hash: tdigest.hash,
@@ -188,7 +186,7 @@ impl REClientBuilder {
             ),
         };
 
-        Ok(REClient::new(grpc_clients))
+        Ok(REClient::new(opts.instance_name.clone(), grpc_clients))
     }
 }
 
@@ -252,6 +250,7 @@ pub struct REState {
 }
 
 pub struct REClient {
+    instance_name: Option<String>,
     grpc_clients: GRPCClients,
     state: Mutex<REState>,
 }
@@ -264,8 +263,9 @@ impl Drop for REClient {
 }
 
 impl REClient {
-    pub fn new(grpc_clients: GRPCClients) -> Self {
+    pub fn new(instance_name: Option<String>, grpc_clients: GRPCClients) -> Self {
         REClient {
+            instance_name,
             grpc_clients,
             state: Mutex::new(REState::default()),
         }
@@ -281,7 +281,7 @@ impl REClient {
         let res = client
             .get_action_result(with_internal_metadata(
                 GetActionResultRequest {
-                    instance_name: INSTANCE_NAME.into(),
+                    instance_name: self.instance_name.clone().unwrap_or("".to_string()),
                     action_digest: Some(tdigest_to(request.digest)),
                     ..Default::default()
                 },
@@ -316,7 +316,7 @@ impl REClient {
         let action_digest = tdigest_to(execute_request.action_digest.clone());
 
         let request = GExecuteRequest {
-            instance_name: INSTANCE_NAME.into(),
+            instance_name: self.instance_name.clone().unwrap_or("".to_string()),
             skip_cache_lookup: false,
             execution_policy: None,
             results_cache_policy: Some(ResultsCachePolicy { priority: 0 }),
@@ -440,7 +440,7 @@ impl REClient {
             })?;
 
         let re_request = BatchUpdateBlobsRequest {
-            instance_name: INSTANCE_NAME.into(),
+            instance_name: self.instance_name.clone().unwrap_or("".to_string()),
             requests: [
                 request
                     .inlined_blobs_with_digest
@@ -507,7 +507,7 @@ impl REClient {
         metadata: RemoteExecutionMetadata,
         request: DownloadRequest,
     ) -> anyhow::Result<DownloadResponse> {
-        download_impl(request, |re_request| async {
+        download_impl(self.instance_name.clone(), request, |re_request| async {
             let mut client = self.grpc_clients.cas_client.clone();
             Ok(client
                 .batch_read_blobs(with_internal_metadata(re_request, metadata))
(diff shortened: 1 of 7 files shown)


remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -422,75 +428,134 @@ impl REClient {
...
+        for blob in inlined_blobs.unwrap_or_default() {
+            let hash = blob.digest.hash.clone();
+            let size = blob.digest.size_in_bytes;
+            let uuid = uuid::Uuid::new_v4().to_string();
+            let data = blob.blob;
+
+            let resource_name = format!(
+                "{}uploads/{}/blobs/{}/{}",
+                instance_prefix, uuid, hash, size
+            );
```
krallin, line 460: Can we point to the underlying docs for this? And ideally I think you might be supposed to actually send the digest type here IIRC (though not in all cases if I read the spec right)?

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -422,75 +428,134 @@ impl REClient {
...
+            }
+            let resp = bs_client.write(stream::iter(rs)).await?;
+
+            if resp.into_inner().committed_size != size as i64 {
+                return Err(anyhow::anyhow!(
+                    "Failed to upload file: invalid committed_size from WriteResponse"
+                ));
+            }
+
+            blob_hashes.push(hash);
```
krallin, line 553: Note: this buffers the whole file in memory, which isn't worse than what was here before, but perhaps could be improved (no big deal if you'd rather not).

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -422,75 +428,134 @@ impl REClient {
...
-            .requests
-            .iter()
-            .map(|x| x.digest.as_ref().unwrap().hash.clone())
-            .collect::<Vec<String>>();
-        let response = client
-            .batch_update_blobs(with_internal_metadata(re_request, metadata))
-            .await?;
+        // upload inlined blobs
+        let inlined_blobs = request.inlined_blobs_with_digest;
+        for blob in inlined_blobs.unwrap_or_default() {
```
krallin, line 450: This has got no concurrency, so we upload all the files in sequence. Can you add that in? It should be a matter of putting all those futures into a Vec, then using `futures::future::try_join_all` on it to execute them concurrently?

You'll need to clone the `bs_client` to be able to use it in all those futures (because it wants  `&mut self`), but that's fine, that's how you're supposed to use it. 

(in an ideal world we would probably have some limits on concurrency via the use of a semaphore, but that can come later)

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -651,7 +716,11 @@ fn convert_action_result(action_result: ActionResult) -> anyhow::Result<TActionR
     Ok(action_result)
 }
 
-async fn download_impl<F, Fut>(request: DownloadRequest, f: F) -> anyhow::Result<DownloadResponse>
+async fn download_impl<F, Fut>(
```
krallin, line 719: FWIW, I imagine downloads probably should get the same treatment as uploads (i.e. use the bytestream API)

krallin (changes requested):
Thanks! I left a number of comments inline.

benbrittain:
I checked out this pull request and attempted using it with `buildbarn`, but just a heads up I'm getting errors. 

```
2023-04-21T19:34:22.517029Z  WARN buck2_execute::re::client: Failed to connect to RE, retrying after sleeping 4 seconds: Error {
    context: "RE: creating client",
    source: Error {
        context: "Error creating CAS client",
        source: Error {
            context: "Error connecting to `grpc://infradev7:8980`",
            source: tonic::transport::Error(
                Transport,
                hyper::Error(
                    Connect,
                    Custom {
                        kind: InvalidData,
                        error: CorruptMessage,
                    },
                ),
            ),
        },
    },
}
```

krallin:
You probably dont have TLS enabled on that server — add `tls = falseˋ to
your options.

benbrittain:
I already have that set. I get (sort of) successful RBE on `main`, but not this branch.

benbrittain:
I can confirm that a rebase to include `tls` fixes it the big error. However I get a different error now:

```
Action failed: root//third_party:artifact (artifact_install)
Internal error (stage: remote_upload_error): Remote Execution Error (GRPC-SESSION-ID): RE: upload: status: Unknown, message: "EOF", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
stdout:
stderr:
Build ID: f04de820-e50a-4273-99aa-9151d3a893fe
RE: GRPC-SESSION-ID
```

thoughtpolice:
Yeah, I saw the EOF message but I thought I fixed it. Maybe not. I should be able to get back to this by Monday, though, so I'll try to get it posted.

remote_execution/oss/re_grpc/src/client.rs
mostynb: You must not specify digest_function here if you're not using one of MD5, MURMUR3, SHA1, SHA256, SHA384, SHA512, or VSO hashes.

If you're using SHA256TREE or BLAKE3 (and probably any hashes added to the spec from now on), then you must specify digest_function here.

https://github.com/bazelbuild/remote-apis/blob/35aee1c4a4250d3df846f7ba3e4a4e66cb014ecd/build/bazel/remote/execution/v2/remote_execution.proto#L227

krallin:
I'm going to close this since #193 is quite a bit more complete and closer to being ready to ship (I included the missing instance name bit there) so it'll make more sense to merge that one instead. Sorry we had duplicate PRs for this! It sounds like there was quite a bit of interest in implementing this.

thoughtpolice:
No worries, I meant to close this myself actually; I'll rebase the other changes and split them out.

krallin:
@thoughtpolice  do you have bits you still needed here? Note that I included instance name in the other patch since we needed this for testing internally

thoughtpolice:
Yes, I would still like 00ed67d6c7928c503515692a32e12bdc0d8b08d2 to be included too, I'll just submit a separate change for that.

krallin:
Ah, yes, go for it

PR #182, opened 2023-04-18
chore(cargo): un-fork and update tonic to 0.9.2

thoughtpolice:


krallin:
FYI, it will likely take a bit of time to land this. We'll want to keep the Tonic version we use in the OSS build in sync with our internal build, but this breaking change means we'll have to do quite a bit of updating first:

https://github.com/hyperium/tonic/pull/1274

krallin:
FYI, it will likely take a bit of time to land this. We'll want to keep the Tonic version we use in the OSS build in sync with our internal build, but this breaking change means we'll have to do quite a bit of updating first:

https://github.com/hyperium/tonic/pull/1274

thoughtpolice:
Ah, in that case I'll close this for now; it can just happen later!

PR #193, opened 2023-04-23
Remote Build Execution upload batching features

benbrittain:
* Queries the RBE Capabilities service to get maximum message size.
* Batches small blob updates into several BatchUploadBlobs requests.
* If file is larger than maximum message size, use Bytestream service for upload.

### DIFF
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -37,6 +37,7 @@ use re_grpc_proto::build::bazel::remote::execution::v2::ActionResult;
 use re_grpc_proto::build::bazel::remote::execution::v2::BatchReadBlobsRequest;
 use re_grpc_proto::build::bazel::remote::execution::v2::BatchReadBlobsResponse;
 use re_grpc_proto::build::bazel::remote::execution::v2::BatchUpdateBlobsRequest;
+use re_grpc_proto::build::bazel::remote::execution::v2::BatchUpdateBlobsResponse;
 use re_grpc_proto::build::bazel::remote::execution::v2::Digest;
 use re_grpc_proto::build::bazel::remote::execution::v2::ExecuteOperationMetadata;
 use re_grpc_proto::build::bazel::remote::execution::v2::ExecuteRequest as GExecuteRequest;
@@ -50,6 +51,7 @@ use re_grpc_proto::google::bytestream::byte_stream_client::ByteStreamClient;
 use re_grpc_proto::google::bytestream::ReadRequest;
 use re_grpc_proto::google::bytestream::ReadResponse;
 use re_grpc_proto::google::bytestream::WriteRequest;
+use re_grpc_proto::google::bytestream::WriteResponse;
 use re_grpc_proto::google::longrunning::operation::Result as OpResult;
 use re_grpc_proto::google::rpc::Code;
 use re_grpc_proto::google::rpc::Status;
@@ -384,16 +386,16 @@ enum BatchUploadRequest {
 
 #[derive(Default)]
 /// Builds up a vector of batch upload requests based upon the maximum allowed message size.
-struct BatchUploader {
+struct BatchUploadReqAggregator {
     max_msg_size: i64,
     curr_req: Vec<BatchUploadRequest>,
     requests: Vec<Vec<BatchUploadRequest>>,
     curr_request_size: i64,
 }
 
-impl BatchUploader {
+impl BatchUploadReqAggregator {
     pub fn new(max_msg_size: usize) -> Self {
-        BatchUploader {
+        BatchUploadReqAggregator {
             max_msg_size: max_msg_size as i64,
             ..Default::default()
         }
@@ -414,7 +416,9 @@ impl BatchUploader {
     }
 
     pub fn done(mut self) -> Vec<Vec<BatchUploadRequest>> {
-        self.requests.push(std::mem::take(&mut self.curr_req));
+        if self.curr_req.len() > 0 {
+            self.requests.push(std::mem::take(&mut self.curr_req));
+        }
         self.requests
     }
 }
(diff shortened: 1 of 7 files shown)


benbrittain:
This is a subset of the improvements made in #181 but includes batching of smaller messages into a reduced number of `BatchUpdateBlobs` requests instead of exclusively using on `ByteStream.Write`.

benbrittain:
I tried running this on my "production" codebase at work as a lazy E2e test and the size of the protobuf messages is not quite right since the blob hashes aren't the only thing in a message. Putting together a fix in a minute.

benbrittain:
Fixed and rebased on `main`.

benbrittain:
I've also added support for not uploading blobs already present on the remote

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -466,77 +577,240 @@ impl REClient {
...
+        // For small file uploads the client should group them together and call `BatchUpdateBlobs`
+        // https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto#L205
+        let mut batched_blob_updates = BatchUploader::new(max_msg_size);
+
+        // Create futures for any blobs that need uploading.
+        for blob in request.inlined_blobs_with_digest.unwrap_or_default() {
+            if !needed_blobs.contains(&blob.digest) {
+                // Blob is already in remote, skip upload.
+                continue;
+            }
```
krallin, line 642:
```suggestion
```

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -466,77 +577,240 @@ impl REClient {
...
+                        resource_name: resource_name.to_owned(),
+                        write_offset,
+                        finish_write: length == 0,
+                        data: data.to_owned(),
+                    });
+
+                    write_offset += length as i64;
+                    reader.consume(length);
+                    if length == 0 {
+                        break;
```
krallin, line 718: Is the use of a buffered reader all that useful here? This adds the complexity of lining up the `consume` calls but without real benefit. We are just going to issue reads for `max_msg_size` anyway:

```suggestion
            let fut = async move {
                let file = tokio::fs::File::open(&name).await?;
                let mut buff = vec![0; max_msg_size];
                
                let mut write_offset = 0;
                let mut upload_segments = Vec::new();
                
                loop {
                    let length = file.read(&mut buff).await.context("Error reading")?;

                    upload_segments.push(WriteRequest {
                        resource_name: resource_name.to_owned(),
                        write_offset,
                        finish_write: length == 0,
                        data: data.to_owned(),
                    });

                    write_offset += length as i64;

                    if length == 0 {
                        break;
```

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -764,17 +1081,26 @@ where
...
+            // If the data is small enough to be transfered in a batch
+            // blob update, just write it directly to file.
+            if req.named_digest.digest.size_in_bytes < max_msg_size as i64 {
+                let data = get(&req.named_digest.digest)?;
+                file.write_all(&data).await.context("Error writing")?;
+            } else {
+                let mut responses = big_f(req.named_digest.digest.clone()).await?;
+                while let Some(ReadResponse { data }) = responses.message().await? {
+                    file.write_all(&data).await.context("Error writing")?;
+                }
```
krallin, line 1098: Can this get some `context` on most calls so that if this errors out we know what happened?

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -764,17 +1081,26 @@ where
...
+            }
             file.flush().await.context("Error flushing")?;
             anyhow::Ok(())
-        }
-        .await
-        .with_context(|| {
+        };
+        fut.await.with_context(|| {
             format!(
                 "Error writing digest `{}` to `{}`",
```
krallin, line 1105: This message should probably change since in the "large blob" case the error might actually occur while downloading, not writing.

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -553,13 +827,32 @@ impl REClient {
...
+            |digest| async move {
+                let mut client = self.grpc_clients.bytestream_client.clone();
+
+                let resource_name = format!("blobs/{}/{}", digest.hash, digest.size_in_bytes);
+                let response = client
+                    .read(ReadRequest {
+                        resource_name,
+                        read_offset: 0,
+                        read_limit: 0,
+                    })
```
krallin, line 850: Can this get the `with_internal_metadata` wrapping as well?

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -697,44 +990,70 @@ fn convert_action_result(action_result: ActionResult) -> anyhow::Result<TActionR
...
+    let mut bytestream_digests = vec![];
+    for digest in file_digests
+        .iter()
+        .map(|req| &req.named_digest.digest)
+        .chain(inlined_digests.iter())
+        .map(|d| tdigest_to(d.clone()))
+        .filter(|d| d.size_bytes > 0)
+    {
+        if digest.size_bytes as usize > max_msg_size {
+            bytestream_digests.push(digest);
```
krallin, line 1020: I don't actually see you doing anything with this (but see my comment below)

krallin (changes requested):
Thank you! I added some inline comments for you review.

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -466,77 +577,240 @@ impl REClient {
...
+                    .await?;
+                let resp: FindMissingBlobsResponse = missing_blobs.into_inner();
+                for blob in resp.missing_blob_digests {
+                    needed_blobs.insert(tdigest_from(blob));
+                }
+            }
+        }
+
+        anyhow::Ok(needed_blobs)
+    }
```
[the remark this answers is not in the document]
benbrittain: How would `Uploader` have context on what other clients have uploaded already? Additionally, I'll note that without this snippet, remote execution is currently unusable for my workload.

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -764,17 +1081,26 @@ where
...
 
-        async {
+        let fut = async {
             let mut file = opts
                 .open(&req.named_digest.name)
                 .await
                 .context("Error opening")?;
-            file.write_all(&data).await.context("Error writing")?;
+            // If the data is small enough to be transfered in a batch
+            // blob update, just write it directly to file.
```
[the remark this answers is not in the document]
benbrittain: I was trying to get at the fact that one is buffered, will make clearer :+1:

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -466,77 +577,240 @@ impl REClient {
...
-                    .into_map(|x| Request {
-                        digest: Some(tdigest_to(x.digest)),
-                        data: x.blob,
-                        compressor: compressor::Value::Identity as i32,
-                    }),
-                files_with_digest,
-            ]
-            .concat(),
-        };
+        // NOTE if we stop recording blob_hashes, we can drop out a lot of allocations.
```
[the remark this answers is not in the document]
benbrittain: Yeah, but also it don't seem to be using the blob hash information except for logging, and we can drop a whole `Vec<Vec<String>>`, so worth thinking about

remote_execution/oss/re_grpc/src/client.rs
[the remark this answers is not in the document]
krallin: > How would Uploader have context on what other clients have uploaded already? Additionally, I'll note that without this snippet, remote execution is currently unusable for my workload.

It calls `get_digests_ttl`, here:

https://github.com/facebook/buck2/blob/f3b83eeba0f395359456c250bfddc951757ddec4/app/buck2_execute/src/re/uploader.rs#L102-L107

It then only uploads things that are about to expire.

However, the problem is that in this client, this is implemented as ... a stub:

https://github.com/facebook/buck2/blob/f3b83eeba0f395359456c250bfddc951757ddec4/remote_execution/oss/re_grpc/src/client.rs#L566-L578

(I didn't realize, my bad, but _that_ is the bit that needs changing)

Specifically, what would need to be changed here is getting this to make a call to find_missing, then return a more realistic TTL for blobs that are found to not be missing. At the very least, that TTL would have to be one that Uploader would be happy with (right now, it expects a TTL of 600 seconds). Then, the upload would not attempt to re-upload things that currently exist.

The RE API doesn't actually provide a way to discover the TTL of blobs (the `get_digests_ttl` method we use internally does, but of course that's not something you can use here). Production RE backends should normally extend the TTL of blobs when they are queried (otherwise you've got a race window no matter what when you call find_missing), but it's also not possible to discover this.

The current design does force you to make a decision as to what you think is the TTL when your backend says the blob isn't missing. I think a reasonable way to solve this would be make both the ttl_wanted in Uploader and the ttl returned by `get_digests_ttl` configurable, at which point we can probably set them to rather low defaults for OSS builds.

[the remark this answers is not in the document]
benbrittain: That makes sense, and explains my confusion! I'll add some logic there.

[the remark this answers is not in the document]
benbrittain: I moved the logic, seems to work :+1:

benbrittain:
I'm get to fix the missed large blobs, but it's on my TODO list. I'd also like to make it so if for some reason a request fails in the batch it gets retried with the bytestream service

benbrittain:
This is ready for another look whenever you have time :)

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -471,72 +582,183 @@ impl REClient {
...
+            let metadata = metadata.clone();
+            let fut = async move {
+                // Number of complete (non-partial) messages
+                let num_msgs = size as usize / max_msg_size;
+                let mut segments = vec![];
+                for (i, chunk) in data.chunks(max_msg_size).enumerate() {
+                    segments.push(WriteRequest {
+                        resource_name: resource_name.to_owned(),
+                        write_offset: (i * max_msg_size) as i64,
+                        finish_write: i == num_msgs,
```
krallin, line 616: If size is a multiple of `max_msg_size`, I think this will not work (e.g. if they're equal, then `i == 0` on the last message but `num_msgs` is 1). It's a little tricky to actually exercise this, because inlined blobs are used only for directories, which are normally rather small (but that's not *required*).

I think it would be simpler to always set this to `false` and simply, in the block below, do:

```
segments.last_mut().unwrap().finish_write = true;
```

This leads me to a separate comment, which is: this function has a lot of logic, but little in the way of testing. Do you think we could add some tests similar to what download has?

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -471,72 +582,183 @@ impl REClient {
...
-                    digest: Some(tdigest_to(x.digest)),
-                    // FIXME: This could do a lot of blocking reads
-                    data: fs_util::read(&x.name)?,
-                    compressor: compressor::Value::Identity as i32,
-                })
-            })?;
+        // NOTE if we stop recording blob_hashes, we can drop out a lot of allocations.
+        let mut upload_futures: Vec<BoxFuture<anyhow::Result<Vec<String>>>> = vec![];
+        let max_msg_size = self.capabilities.max_msg_size;
+        let client_uuid = uuid::Uuid::new_v4().to_string();
```
krallin, line 588: We don't really validate (or expect) that we don't have duplicate blobs in the input here, I think you should generate a new UUID for each upload (i.e. in the place where you generate `resource_name`). This way, we don't have to worry about potentially using the same UUID for two separate uploads of the same blob, and it's not really all that expensive anyway.

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -471,72 +582,183 @@ impl REClient {
...
+            let size = file.digest.size_in_bytes;
+            let name = file.name.clone();
+            let metadata = metadata.clone();
+            if size < max_msg_size as i64 {
+                batched_blob_updates.push(BatchUploadRequest::File(file));
+                continue;
+            }
+            let resource_name = format!("uploads/{}/blobs/{}/{}", client_uuid, hash.clone(), size);
+            let fut = async move {
+                let mut file = tokio::fs::File::open(&name).await?;
```
krallin, line 650: It would be good to indicate what we failed to open here in a `with_context` block.

krallin (changes requested):
Thanks! I included some minor suggestions inline, and asked if you think you could include some unit tests as well?

On our end, I did test this and it'll need a few adjustements to pass our e2e tests, but those are primarily related to our internal RE backend and I'll just include them in your changes.

benbrittain:
I have not forgotten about this, I'll add some upload testing soon :+1:

krallin:
No problem, thanks! :)

benbrittain:
Added a few upload test cases (which, of course, caught a couple of minor problems)!

krallin:
Merged this today (after making a certain number of adjustements), thanks for your contribution to Buck2!

PR #234, opened 2023-05-05
chore: rework the README a bit

thoughtpolice:
chore: rework the README a bit

Summary: The README as it stands doesn't really "sell" Buck2 very well but more
than that, it barely even expresses what it is or why it exists! It has links,
but generally you should expect people to not follow them if the frontmatter
isn't selling it.

This reworks the frontmatter of the README to drill in three major points about
why Buck2 is interesting as a build system.

- It's Fast(TM)
- It's Hermetic
- It's Multi-Language

There are *other* things that are important too, like what makes it different
from Bazel -- but instead just direct people to look at the "Why Buck?" page on
the site, which goes into more detail. And you don't have unlimited space to
extoll the virtues of the project. So these three seem like the major taglines
to stick too. Plus, Bazel users will likely be willing to look in more detail
since they don't need to be convinced about points 2 and 3, at least.

Furthermore, just point people to the binary releases to get started. For the
vast majority of potential users, there's very little point to listing the
Cargo build instructions when it's not relevant to their immediate interests.

This also moves some of the coding conventions stuff, and the build
instructions, to a new file: `HACKING.md` which is more or less self
explanatory.


thoughtpolice:
[Rendered link](https://github.com/thoughtpolice/buck2/tree/sapling-pr-archive-thoughtpolice#readme)

HACKING.md
```diff
@@ -0,0 +1,78 @@
...
+
+Build uses prebuilt `protoc` binary from
+[protoc-bin-vendored](https://crates.io/crates/protoc-bin-vendored) crate. If
+these binaries to do not work on your machine, the path to `protoc` binary and
+protobuf include path can be specified via `BUCK2_BUILD_PROTOC` and
+`BUCK2_BUILD_PROTOC_INCLUDE` environment variables.
+
+### Using [Nix] to compile the source
+
+Most [Nix] users provision tools directly with Nix itself, rather than rustup;
```
ndmitchell, line 33: I think we probably want one or two words here to make it clear that Nix is not the recommended way to do this, but a possible alternative. We don't want people trying to install Buck2 to accidentally think they need Nix before they continue.

HACKING.md
```diff
@@ -0,0 +1,78 @@
...
+Or alternatively install it directly from GitHub:
+
+```sh
+rustup install nightly-2023-03-07
+cargo +nightly-2023-03-07 install --git https://github.com/facebook/buck2.git buck2
+```
+
+Build uses prebuilt `protoc` binary from
+[protoc-bin-vendored](https://crates.io/crates/protoc-bin-vendored) crate. If
+these binaries to do not work on your machine, the path to `protoc` binary and
```
ndmitchell, line 27: Do we have a good idea where these won't work? Should we start by saying that in most cases it will work?

HACKING.md
```diff
@@ -0,0 +1,78 @@
+# Tips and tricks for hacking on Buck2
+
```
ndmitchell, line 2: Perhaps a link to contributing here? Since it seems contributing is about "yes, we'd love you to contribute" and Hacking is how to actually do it?

ndmitchell (commented):
This is really good - thanks for taking the time to make these improvements. Sorry for the delay in the review (I was travelling, and then sprained my ankle, which limited the amount of time I could sit at a computer for...)

A few notes - happy for you to revise it and I'll merge as is, or happy to make the edits myself. Your call.

thoughtpolice:
I can fix these later tonight. Hope your ankle is feeling better!

HACKING.md
thoughtpolice: Because this is a detail handled in the Nix case and should be handled on most systems with static exes, I just removed this bit.

thoughtpolice:
@ndmitchell Neil, I've updated this with the changes you suggested. I also touched up the front-matter on the README a little, and added two footnotes for the sake of honesty (and to avoid weird internet tendencies where people will think these were like, bad faith, or something):

- Performance notes like "it's 2x faster" come from Buck1 versus Buck2 at Meta. Practical comparisons to Bazel are omitted and haven't been performed, not out of spite but because, well obviously, Buck1 was what you had to replace and be compatible with!
- Hermeticity requires Remote Execution, but that should be fixed Soon(TM).

HACKING.md
```diff
@@ -0,0 +1,83 @@
...
+Or alternatively install it directly from GitHub:
+
+```sh
+rustup install nightly-2023-03-07
+cargo +nightly-2023-03-07 install --git https://github.com/facebook/buck2.git buck2
+```
+
+Build uses prebuilt `protoc` binary from
+[protoc-bin-vendored](https://crates.io/crates/protoc-bin-vendored) crate.
```
steveklabnik, line 30: Not sure if this is important enough to mention here, but you can also override this with an environment variable, which was important to get it working on my platform.

thoughtpolice: I've also reworked this into a side note about alternative operating systems and the need for `protoc`. Could be more organized but I think this should help.

ndmitchell:
Thanks!

PR #239, opened 2023-05-07
Add a basic system_go_toolchain example

sluongng:
Provide a basic example with all the attributes of the GoToolchainInfo provider populated.

Users are expected to bring their own GO_ROOT and Windows is not yet supported.

### DIFF
--- a/prelude/toolchains/go.bzl
+++ b/prelude/toolchains/go.bzl
@@ -1,0 +1,62 @@
+load("@prelude//go:toolchain.bzl", "GoToolchainInfo")
+
+def _system_go_toolchain_impl(ctx):
+    go_root = ctx.attrs.go_root
+    go_binary = go_root + "/bin/go"
+
+    arch = host_info().arch
+    if arch.is_aarch64:
+        go_arch = "arm64"
+    elif arch.is_x86_64:
+        go_arch = "amd64"
+    else:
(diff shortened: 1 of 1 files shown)


themarwhal:
This looks great, thank you! 😄 Would it be possible to add a Go example to `example/prelude`? It would be great to have some basic CI coverage on it.

sluongng:
@themarwhal The truth is that in the current stage, the go rules in `prelude//go/...` are not functional at all.

When trying to compile with the system toolchain provided in this PR, you would get no Go standard library to compile 
and link against. There are several things needed to fix `prelude//go/...` before it could be used in Open Source and currently, I wonder how they work as-is inside Meta 🤔 . Happy to discuss this further in https://github.com/facebook/buck2/issues/240

Another difficulty in testing this would be that it's a system toolchain, so it would require the CI system on Meta's CircleCI instance to have Go installed for any tests to pass.

In my branch over at https://github.com/facebook/buck2/compare/main...sluongng:buck2:sluongng/remote-go-toolchain, I have a remote toolchain that would be easier to set up tests against. I could send that as a follow-up after this PR gets merged.

ndmitchell:
Let's keep the discussion on how to get Go working in #455 (closed PR's tend to not show up in discussion topics for a lot of peopl)

PR #270, opened 2023-05-20
nix: add rust-src to dev shell

stolyaroleh:
..in order to fix the following `rust-analyzer` error:
```
[ERROR project_model::workspace] Failed to find sysroot for Cargo.toml file /home/oleh/src/buck2-upstream/./Cargo.toml. Is rust-src installed? e=can't load standard library from sysroot
/nix/store/d92bh46p85l310vg43xfai43xd0gmfp8-rust-default-1.70.0-nightly-2023-03-07
(discovered via `rustc --print sysroot`)
try installing the Rust source the same way you installed rustc
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,7 +21,7 @@
 
     rust-version = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain;
     my-rust-bin = rust-version.override {
-      extensions = [ "rust-analyzer" ];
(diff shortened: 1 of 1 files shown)


thoughtpolice (commented):
This looks correct to me, FWIW. I haven't used `rust-analyzer` on this project in VSCode so it slipped by me.

PR #279, opened 2023-05-30
Set protobuf environment variables in Nix shell

kmicklas:
Without this building on NixOS (in the Nix shell) will hit the issue described in the README. This add environment overrides to use the protobuf toolchain from Nixpkgs instead of `protoc-bin-vendored`.


ndmitchell:
Thanks @kmicklas and thanks @thoughtpolice for the review.

PR #302, opened 2023-06-18
Fix some grammatical errors in readme

TzviPM:
Just some issues I found while reading through the project readme.


thoughtpolice:
Note that I'm overhauling the README a bit in another patch, so this will conflict; see #234. The text is the same in some places, though, would love any grammar feedback on that one too!

ndmitchell:
Thanks! Some of these changes got applied in #234, and I've rebased the remaining ones over the top of it, so will land both shortly.

PR #303, opened 2023-06-19
http_archive: allow remote execution

sluongng:
In OSS, SHA256 is a lot more popular than SHA1.

### DIFF
--- a/prelude/http_archive/http_archive.bzl
+++ b/prelude/http_archive/http_archive.bzl
@@ -111,7 +111,7 @@ def http_archive_impl(ctx: "context") -> ["provider"]:
 
     # The HTTP download is local so it makes little sense to run actions
     # remotely, unless we can defer them.
-    local_only = ctx.attrs.sha1 == None
+    local_only = ctx.attrs.sha1 == None and ctx.attrs.sha256 == None
 
     ext_type = _type(ctx)
 


ndmitchell:
@krallin - given the support for Blake3 as well, what should be the logic here?

krallin:
@ndmitchell Perhaps we should expose the supported hash algorithms via `host_info` and make this dependent on whether the available hash is supported?

ndmitchell:
No objection to returning the hash algorithms on host-info. Instead of exposing a way to find out if you can defer it, why not have an option to `download_file` that says defer if you can, else don't?

krallin:
> Instead of exposing a way to find out if you can defer it, why not have an option to download_file that says defer if you can, else don't

We do have that actually, this is trying to do something a bit different: it wants to figure out if we should run the action locally vs. remotely based on whether we expect the download to be deferrable.

TBH, maybe that code should just be deleted now that we have hybrid execution

krallin:
Sorry for the delay here ... I landed https://github.com/facebook/buck2/commit/186f296227b09d00da54a8c945c459cd228d6662 to expose the digest config as I was mentioning

Now that we have that, I'll land the changes to have http_archive look at it once it ships internally (next Monday)

PR #305, opened 2023-06-25
Bump rust toolchain to nightly 2023-06-23

benbrittain:
* Fixes ICE in the Github CI
* NodeKey enum size decrease

### DIFF
--- a/app/buck2_build_signals_impl/src/lib.rs
+++ b/app/buck2_build_signals_impl/src/lib.rs
@@ -97,7 +97,7 @@ assert_eq_size!(DeferredResolve, [usize; 4]);
 assert_eq_size!(ConfiguredTargetNodeKey, [usize; 2]);
 assert_eq_size!(InterpreterResultsKey, [usize; 1]);
 assert_eq_size!(BuildArtifact, [usize; 6]);
-assert_eq_size!(NodeKey, [usize; 8]);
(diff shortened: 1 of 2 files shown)


thoughtpolice:
Is there any insight into exactly what is triggering this ICE? I have a nightly build of buck2 on Linux that gets updated every day [in this repository](https://github.com/thoughtpolice/buck2-nix/), but I haven't seen any ICEs at all in my builds, despite the fact this seems completely reproducible and happens across multiple OSs/architectures.

benbrittain:
I honestly have no idea. I tried permuting a few things before attempting a toolchain upgrade which seems to have removed the ICE entirely. (looks like I should fix the other builders though too)

stepancheg:
Bumping is not that easy, because we synchronize the version with the version used in internal non-cargo builds (in particular, size changes will break internal builds).

dtolnay:
We have an upgrade to 1.71.0-beta.3 in progress, because of the ICEs with 1.70.0-stable. That will fix the GitHub CI &mdash; hopefully by early next week. @diliop

dtolnay:
23efaf53ecf6737cb215686e244ff99874f647be

dtolnay:
CI builds are working again:

Screenshot from 2023-06-28 13-28-47

Issue #256, opened 2023-05-13
Use remote_execution.RequestMetadata


sluongng:
Currently requests to remote execution platforms do not come with a RequestMetadata https://github.com/facebook/buck2/blob/aebff4791208ab8f2df0787f9147196ab392c91d/remote_execution/oss/re_grpc_proto/proto/build/bazel/remote/execution/v2/remote_execution.proto#L1883. Having this would help RBE services to differentiate a Buck2 remote build from others.

Right now, there is only internal metadata attached to the request https://github.com/facebook/buck2/blob/aebff4791208ab8f2df0787f9147196ab392c91d/remote_execution/oss/re_grpc/src/client.rs#L1252-L1257.

ndmitchell:
See #310 where this is being discussed.

alexlian:
Should be addressed with: https://github.com/facebook/buck2/pull/502

PR #310, opened 2023-06-29
feat(remote_execution): attach RequestMetadata for non-fbcode builds

thoughtpolice:
Summary: When doing requests to a RBE service, it's useful to attach the 'RequestMetadata' type defined by Bazel onto gRPC requests, using a specific header.

This is actually already done, more or less, but only for the internal fbcode build service which uses a different and simpler Metadata structure, and a different header. Therefore attaching a new header isn't so complicated.

This also ensures that non-fbcode builds of Buck2 don't bother attaching fbcode Metadata, since it's not useful and can only be tested internally there, anyway. And likewise, we don't attach OSS Bazel RequestMetadata to fbcode_build configurations, either.

Test Plan: NIH.

Change-Id: Irrsqpmptqmzzoukzuzrqmkxsoqtxvuvx

### DIFF
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -46,7 +46,9 @@ use re_grpc_proto::build::bazel::remote::execution::v2::FindMissingBlobsRequest;
 use re_grpc_proto::build::bazel::remote::execution::v2::FindMissingBlobsResponse;
 use re_grpc_proto::build::bazel::remote::execution::v2::GetActionResultRequest;
 use re_grpc_proto::build::bazel::remote::execution::v2::GetCapabilitiesRequest;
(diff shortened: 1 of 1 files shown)


thoughtpolice:
This will eventually fix #256, but it isn't ready yet. Just wanted to file it to get it out of the way.

krallin:
Could we preferably expose this via a config option rather than a build flag? We do use our internal RE in order to test the OSS build, and that doesn't build with fbcode_build, so this would break that testing.

thoughtpolice:
Exactly why I drafted this instead of just submitting! I figured there would be some catch I didn't understand. I can rework this a little later.

JakobDegen:
#502 has landed which incorporates all of this and then some, so I'm gonna close this. Feel free to reopen if you disagree

PR #315, opened 2023-06-30
feat(github): test buck2-on-buck2 build with GitHub actions

thoughtpolice:
This is just a starter and a follow up from #313. ~~This is based on and cannot be merged until #314 is merged, and I rebase it.~~

I'm using Nix for this workflow right now &mdash; but just because it's a super easy way to provision `reindeer`, `clang` and `lld` without having to install them otherwise. The magic nix cache should make this pretty fast and easy, and it's the best place I know to get quick binaries.

macOS is busted in the build matrix, but might not be too hard to fix, I think. Eventually this will probably want to include Windows though, so we'll have to do something about that...

Builds clock in at about 9m with this.

### DIFF
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -1,0 +1,34 @@
+name: Buck2 with Buck2
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        target:
+          - os: 'ubuntu-22.04'
+            triple: 'x86_64-unknown-linux-gnu'
+          # TODO FIXME (aseipp): broken with nix, needs some tweaks in the cxx_toolchain for c++abi i think...
+          #- os: 'macos-12'
+          #  triple: 'x86_64-apple-darwin'
+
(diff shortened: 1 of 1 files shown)


thoughtpolice:
@ndmitchell @stepancheg This is ready for review since #314 was merged

thoughtpolice:
Sorry, I re-stacked some patches and resynchronized my branches, so this also includes #320. I've re-drafted this; we can merge #320 first and then I'll rebase again.

One day GitHub will support real stacked diffs, we can hope...

thoughtpolice:
This is ready for review again.

themarwhal:
Hello! Sorry for a delay in getting to this. Adding bootstrapping coverage in CI is a great idea, I added it as a part of the CircleCI pipeline as that's where we currently have our tests. (https://github.com/facebook/buck2/pull/327). It's still missing Windows coverage, which I'll take a look at in a bit.

thoughtpolice:
Cool! I don't really know much about CircleCI and the Nix approach wouldn't have worked anyway for Windows, so the only real advantage of this patch is "maybe 10 less lines of code." Closing; thanks!

PR #321, opened 2023-07-01
fix(buck2_common): be less aggressive about HTTP retry warnings

### LINKED ISSUE #316: Warnings that occur when HTTP2 downloads are throttled are a little bit aggressive
https://github.com/thoughtpolice/buck2/actions/runs/5418361387/jobs/9850380998

If you check the `Build Buck2 with Buck2` step, you'll a duplicate number of rather aggressive warnings &mdash; I believe these originate from being too aggressive when downloading packages from https://crates.io, and getting rate limited:

```
WARN buck2_common::http::retries: Retrying a HTTP error after 2 seconds: HTTP Client error: HTTP: Error sending request: http2 error: stream error received: refused stream before processing any application logic
```

However, it's actually a bit worse when superconsole is in use in your terminal &mdash; here it just uses a more typical log stream that streams once &mdash; because it gets spammed a bunch during refresh of the active targets being built. Looks like it happens dozens of times in Windows Terminal for me, at least. I can attach a small asciinema of this but it should be easy to reproduce, I think.

Honestly, I don't think this warning as it stands now is really *that* important, a little rate limiting is expected if you're doing a fresh build with no cache, I think. You're downloading like 100+ dependent crates pretty quickly in this scenario. The download phase of the crates being extended by a second or two is probably OK if you end up getting a big set of them in batches, it may only add a few seconds to the initial download actions.

I think better logic would be if Buck2:

- Downloads with typical exponential backoff during throttling (I assume this is the case already), and
- Only throws an error after a number of retries results in, say, a backoff window of 60 seconds or more occurs.

60 seconds is arbitrary but seems like it's egregious enough to make that an error. A hair you'd have to split is whether or not you would only do one warning or just warn on every single action. Maybe you only want one warning per HTTP domain; e.g. if two download_file actions to download two separate files from http://foo.example.com get rate limited, only report once for the foo.example.com domain &mdash; and if another download_file call to http://bar.example.com *also* gets rate limited, report that, so you'd only have two warnings total.

thoughtpolice:
Summary: When downloading from sites like crates.io, we can pretty easily get rate-limited when a lot of our crates are uncached. Don't be so aggressive about warning; most small backoffs won't be too noticeable.

As a note, we _don't_ emit the error message we get from the callstack here in the message; the logic is that it's already a retriable error, so it doesn't make much sense to spam an error that we'd probably get again anyway, and that we ultimately are going to re-attempt. This should just make the UX a little more polished.

Also emit the domain name we're hitting problems with.

Test Plan: Change the new warning conditional from 30 seconds to zero seconds. Build buck2 with buck2. Observe the new error message, which helpfully points out we're connecting to crates.io. Change the conditional back to 30 seconds. Rebuild buck2 with buck2. Note that no warnings appear, because the exponential backoff only results in a few small 2-second retries.

GitHub Issue: Fixes #316

### DIFF
--- a/app/buck2_common/src/http/retries.rs
+++ b/app/buck2_common/src/http/retries.rs
@@ -45,7 +45,11 @@ pub trait AsHttpError {
     fn as_http_error(&self) -> Option<&HttpError>;
 }
 
-pub async fn http_retry<Exec, F, T, E>(exec: Exec, mut intervals: Vec<Duration>) -> Result<T, E>
+pub async fn http_retry<Exec, F, T, E>(
+    dl_url: &str,
+    exec: Exec,
+    mut intervals: Vec<Duration>,
+) -> Result<T, E>
 where
     Exec: Fn() -> F,
     E: AsHttpError + std::fmt::Display,
@@ -53,6 +57,13 @@ where
 {
     intervals.insert(0, Duration::from_secs(0));
     let mut backoff = intervals.into_iter().peekable();
+
+    let host = url::Url::parse(dl_url)
+        .unwrap()
+        .host_str()
+        .unwrap()
+        .to_owned();
+    let mut retries = 0;
 
     while let Some(duration) = backoff.next() {
         tokio::time::sleep(duration).await;
@@ -64,11 +75,18 @@ where
         if let Some(http_error) = http_error {
             if http_error.is_retryable() {
                 if let Some(b) = backoff.peek() {
-                    tracing::warn!(
-                        "Retrying a HTTP error after {} seconds: {:#}",
-                        b.as_secs(),
-                        http_error
-                    );
+                    let secs = b.as_secs();
+                    if secs > 30 {
+                        // NOTE (aseipp): 30 seconds is arbitrary for this, but
+                        // at that point things will be pretty annoying!
+                        tracing::warn!(
+                            host = host,
+                            retries = retries,
+                            "Egregious HTTP backoff - retrying slow request in {} seconds",
+                            secs,
+                        );
+                    }
+                    retries += 1;
                     continue;
                 }
             }
@@ -143,29 +161,33 @@ mod tests {
 
     #[tokio::test]
     async fn test_http_retry_success() {
+        let url = "http://example.com";
         let mock = Mock::new(vec![ok_response()]);
-        let result = http_retry(|| mock.exec(), retries(0)).await;
+        let result = http_retry(url, || mock.exec(), retries(0)).await;
         assert!(result.is_ok());
     }
 
     #[tokio::test]
     async fn test_http_retry_retryable() {
+        let url = "http://example.com";
         let mock = Mock::new(vec![retryable(), ok_response()]);
-        let result = http_retry(|| mock.exec(), retries(1)).await;
+        let result = http_retry(url, || mock.exec(), retries(1)).await;
         assert!(result.is_ok());
     }
 
     #[tokio::test]
     async fn test_http_retry_exhaust_retries() {
+        let url = "http://example.com";
         let mock = Mock::new(vec![retryable(), ok_response()]);
-        let result = http_retry(|| mock.exec(), retries(0)).await;
+        let result = http_retry(url, || mock.exec(), retries(0)).await;
         assert!(result.is_err());
     }
 
     #[tokio::test]
     async fn test_http_retry_non_retryable() {
+        let url = "http://example.com";
         let mock = Mock::new(vec![non_retryable(), ok_response()]);
-        let result = http_retry(|| mock.exec(), retries(1)).await;
+        let result = http_retry(url, || mock.exec(), retries(1)).await;
         assert!(result.is_err());
     }
 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -270,6 +270,7 @@ triomphe = "0.1.8"
 trybuild = "1.0.56"
 twox-hash = "1.6.1"
 unicode-segmentation = "1.7"
+url = "2.4.0"
 uuid = { version = "1.2", features = ["v4"] }
 walkdir = "2.3.2"
 which = "4.3.0"
--- a/app/buck2_client_ctx/src/manifold.rs
+++ b/app/buck2_client_ctx/src/manifold.rs
@@ -467,6 +467,7 @@ impl ManifoldClient {
         }
 
         let res = http_retry(
+            &url,
             || async {
                 self.client
                     .put(&url, buf.clone(), headers.clone())
@@ -499,6 +500,7 @@ impl ManifoldClient {
         );
 
         let res = http_retry(
+            &url,
             || async {
                 self.client
                     .post(&url, buf.clone(), vec![])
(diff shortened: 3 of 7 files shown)


krallin:
Not a huge fan of trying to parse the host of out of the url and doing so with an unwrap here. Can we just remove that?

Internally, we also wouldn't actually want this as-is, since hiding the URL makes the message pretty useless as a whole.

app/buck2_common/src/http/retries.rs
```diff
@@ -64,11 +75,18 @@ where
         if let Some(http_error) = http_error {
             if http_error.is_retryable() {
                 if let Some(b) = backoff.peek() {
-                    tracing::warn!(
-                        "Retrying a HTTP error after {} seconds: {:#}",
-                        b.as_secs(),
-                        http_error
-                    );
+                    let secs = b.as_secs();
```
krallin, line 78: Could we maybe check elapsed + this backoff, and then use that and the 30 second threshold? I'm a bit concerned about not showing anything until we start facing a very large backoff.

app/buck2_common/src/http/retries.rs
```diff
@@ -64,11 +75,18 @@ where
...
+                    if secs > 30 {
+                        // NOTE (aseipp): 30 seconds is arbitrary for this, but
+                        // at that point things will be pretty annoying!
+                        tracing::warn!(
+                            host = host,
+                            retries = retries,
+                            "Egregious HTTP backoff - retrying slow request in {} seconds",
+                            secs,
```
krallin, line 86: - Can we avoid calling the backoff "egregious"? 
- Also, do note we are not retrying a slow request here we're rather retrying a failed one
- We shouldn't hide the error here

app/buck2_common/src/http/retries.rs
krallin: I also note that I can' find any places that retry for 30 seconds so this really just amounts to just deleting the log message.

krallin (changes requested):
I'm a bit concerned that we're really far away from what you're looking for here. Its gonna be hard to reconcile those two things with small adjustements.

--

I think I'd go a different route here. I'd suggest adding an event (in buck2.proto) for this and then just dispatching that to the log file but not producing any user-visible output. This way we'd retain debug capabilities without actually being verbose.

This is unfortunately a bit tricky to do because this code gets invoked (mostly) in places that don't have an event dispatcher set. There is a bit of prior art to this, which is to just dispatch to all connected commands (which is mostly fine as far as debugging is concerned):

https://github.com/facebook/buck2/blob/main/app/buck2/src/panic.rs#L163-L176

Unfortunately this still not great because some places don't have that either (the log upload places, mostly)

So I think what I'd suggest doing is:

- Create a trait for dispatching this warnng
- Have an implementation that fires off an event like noted above,
  - Use it in materialize/http.rs
- Have an implementation that just logs
  - Use it in manifold.rs

thoughtpolice:
> Not a fan of trying to parse the host of out of the url and doing so with an unwrap here. Can we just remove that?
> 
> Internally, we also wouldn't actually want this as-is, since hiding the URL makes the message pretty useless as a whole.

I'm confused, do you mean the whole URL or just the *host*? Because in the open source version (I don't know if error propagation is different inside Meta), there was NEVER any indication of what the URL *or* the host was when this warning popped up, even before I wrote this patch; the only thing you saw was typically just the Debug/display usage on a big `http::streaming_error` of some kind, which is frankly completely and totally useless.

Actually, the only reason I knew that my initial throttling was from crates.io wasn't because of the message; it was because there *weren't any other hosts* I was downloading from when bootstrapping a buck2, so it was pretty clear there was only one culprit.

I also don't think the URL actually *is* important, and I deliberately chose to only show the host here. Why? Well, a few reasons:

- The URL can actually be long, and superconsole doesn't seem to handle wide stuff like this very well. What will actually happen is it will overflow at something like 80 columns in windows terminal, pushing an extra line onto the screen, and then causing the whole UI to "violently" repaint the entire build UX. So it's jarring to have a really wide message; part of the motivation was that the display'd error was *hundreds* of characters long and would repeat 5 or 6 times...
  - Yes, this is strictly a separate issue, but I'd like not to make a recursive stack of tasks that might become 20 extra tasks long to just clean up this one thing.
- The *host* is what is throttling you, not necessarily the URL! Yes, there is some mixture overlap the two (like when an operator sees a hot resource that might be causing hot key loads, and might put up a rate limiter or load shed on that one key). But generally speaking I don't think this is very weird, and in 99.99% of places you download, it will be a host setting, not a specific item. But more importantly...
- It isn't an error! If the backoff works, because the service allows you to succeed, **then there was no error** and the URL isn't relevant. But if there is a failure due to giving up from retrying too many times, **then** you will get the URL too and a loud explosion when the build fails. The message by itself isn't an error, it's merely an informative tell that the host has throttled you *multiple* times. But that still doesn't mean anything will fail!

This is generally how I approach retriable errors: you should retry silently *until* it is pretty obviously visible, and only fail loudly when you can no longer do that. That said, I know not everyone will agree with that point; some people always want the full context even on warnings and the host may not seem like enough information. So if you disagree with this I understand, but it wasn't entirely random, either.

thoughtpolice:
Also, I guess this doesn't cover the case where a host can shitcan you before 30 seconds even occurs, so yeah, the window is totally arbitrary. I'll take any solution that doesn't spam 10 logs a second into superconsole and make the build UX look awful, though.

thoughtpolice:
> I think I'd go a different route here. I'd suggest adding an event (in buck2.proto) for this and then just dispatching that to the log file but not producing any user-visible output. This way we'd retain debug capabilities without actually being verbose.
> 
> This is unfortunately a bit tricky to do because this code gets invoked (mostly) in places that don't have an event dispatcher set. There is a bit of prior art to this, which is to just dispatch to all connected commands (which is mostly fine as far as debugging is concerned):
> 
> https://github.com/facebook/buck2/blob/main/app/buck2/src/panic.rs#L163-L176
> 
> Unfortunately this still not great because some places don't have that either (the log upload places, mostly)
> 
> So I think what I'd suggest doing is:
> 
> * Create a trait for dispatching this warnng
> * Have an implementation that fires off an event like noted above,
>   
>   * Use it in materialize/http.rs
> * Have an implementation that just logs
>   
>   * Use it in manifold.rs

Alright, this does look like a better plan of attack (I somehow missed this comment in favor of the OG one, I guess since I clicked through on my email to reply.) I'll drop this for now and see if I can get this implementation working

cjhopman:
@thoughtpolice somewhat unrelated but I noticed this comment:

> What will actually happen is it will overflow at something like 80 columns in windows terminal, pushing an extra line onto the screen, and then causing the whole UI to "violently" repaint the entire build UX

That doesn't sound very good, and I don't think we see behavior like that on other platforms.

I know the console rendering works quite different on windows (aiui it uses some actual like control commands instead of encoding them in control codes in the output stream), so it's not entirely surprising that it's different (though aiui, windows terminals may support terminal control codes these days).

Anyway, have you reported this behavior as a separate issues anywhere? If you could do that that'd be great (and if you could get like a screen capture of it, even better).

Issue #292, opened 2023-06-12
Release cadence


benbrittain:
As we get closer to rolling out our internal Buck test, we need a way of keeping our repository and our buck2 versions in-sync. 

https://github.com/facebook/buck2/releases provides enough information (the prelude_hash is a particularly helpful touch), but it would be nice if prior prebuilt binaries were available.

Is there any chance we could get tags/releases placed on the repo at a roughly weekly cadence (similar to how https://github.com/facebook/watchman/releases is released)?

This would make our adoption much easier

thoughtpolice:
I was going to suggest something like this myself at some point soon, though my suggestion would have been two weeks: once on the first of every month and once on the 15th. It doesn't really matter in the end; some thoughts I had:

- These releases should all probably be thoroughly marked as "pre-releases" to set expectations correctly.
  - I don't know what kind of conditions would need to bet met for a first "stable" release, but I sort of doubt we're close to it at the moment?
- For now, calver (i.e. `v23.06.01` followed by `v23.06.15` in my example) is probably the way to do Buck2 releases, based on the way it's rolled out at Meta, and probably how people will consume it downstream. Maybe calver is just the best way to do things in this project, like watchman?
  - Decoupling the rules from the core binary is actually a huge win for this since it allows you much more leeway in rollouts and end-user control.
- Stable cadence is more important than everything being perfect in every release. Having a release every week or two weeks is fast enough to the point that a user regression can be handled by telling someone to wait a week (or two) for an official fix in a new release as a workaround. Also, many bugfixes can probably be done in the rule code itself, avoiding the need for binary updates.
- It would be nice to document changes/have a CHANGELOG but in lieu of that, "Visit the diff between commit `foobar` and `barbaz`" is probably good enough.
- People should (probably?) be encouraged to update regularly and report issues. You're probably going to have a hard time updating once every 6 months due to the pace of change, but every week or two should be pretty manageable.


But this is something that would be great to see and is mostly a GitHub `.workflow/` file away from happening!

benbrittain:
Is this something that a PR would be accepted for?

cjhopman:
Yeah, I think a PR for this could be accepted. CalVer of some sort is roughly what we were thinking of and I think twice a month seems fine if that's what people are generally fine with. I'd love to figure out how we can ensure we consistently get an up-to-date CHANGELOG, but I suspect that'll be a challenge (mostly in figuring out how to ensure that it actually gets updates).

[...]

PR #322, opened 2023-07-01
chore(docs): notes on unstable rustc in HACKING.md

thoughtpolice:
Some notes on `rustc` unstable usage, after talking in https://github.com/facebook/buck2/pull/320. Figured this would be nice to add.

I have a small upcoming change (I'll file it separately after trying to make a few more) that will use the `tagref` syntax; it's nothing to worry about for anyone else except me, I think. (e.g. `int_roundings` can be replaced with a small bit of code, but it would be nice to undo that when it *does* become stable, so keeping a note on it is nice.)

Also rework some changes from Neil's recent commit; everything in HACKING.md uses `-` for bullet points instead of `*`, so it's just a tiny consistency change.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -121,10 +121,57 @@ have written. Some rules:
 Most code is shared as-is between open source and the internal Meta version of
 Buck2. However, there are some exceptions:
 
-* The open-source remote execution client is different, because our internal
-  one works with custom servers/infrastructure that is not publicly available.
-* There are places controlled with `is_open_source()` which change configuration
+- The open-source remote execution client is different, because our internal one
+  works with custom servers/infrastructure that is not publicly available.
+- There are places controlled with `is_open_source()` which change configuration
   between the internal and open source versions.
-* Some places use `@oss-enable` or `@oss-disable` to comment/uncomment lines
-  of code. The internal code is visible, but the comment markers are moved
-  during export/import of code.
+- Some places use `@oss-enable` or `@oss-disable` to comment/uncomment lines of
+  code. The internal code is visible, but the comment markers are moved during
+  export/import of code.
+
+## Unstable `rustc` features
+
+<!-- [tag:unstable-rustc] -->
+
+TL;DR: Buck2 developers **are free to use unstable `rustc` features**. Please
+submit patches to change this if you are interested in building buck2 with a
+stable Rust compiler. Follow issue https://github.com/facebook/buck2/issues/265
+for more information and changes to this policy.
+
+Buck2 currently uses unstable features of `rustc` freely. Actually, building and
+using Buck2 requires an exact version of `rustc` that is checked at build time;
+this required version is identical within Meta and outside. No other version is
+supported or allowed.
+
+However, there is some outside desire to move away from using unstable Rust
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -121,10 +121,57 @@ have written. Some rules:
...
+- The open-source remote execution client is different, because our internal one
+  works with custom servers/infrastructure that is not publicly available.
+- There are places controlled with `is_open_source()` which change configuration
   between the internal and open source versions.
-* Some places use `@oss-enable` or `@oss-disable` to comment/uncomment lines
-  of code. The internal code is visible, but the comment markers are moved
-  during export/import of code.
+- Some places use `@oss-enable` or `@oss-disable` to comment/uncomment lines of
+  code. The internal code is visible, but the comment markers are moved during
+  export/import of code.
```
thoughtpolice, line 130: Ugh. This was supposed to only change `*` to `-` in the bullet points but of course my editor reflowed it on save. I can back this out if desired.

ndmitchell:
Happy enough with your reformat.

For the unstable features, I think we are only interested in removing unstable features if it does not harm the quality of the code too much. In most cases, the use of the unstable feature was to make things better, so normally I am against removing it. In the fullness of time most unstable features become stable, so I think it will happen eventually. E.g. for int_roundings, https://github.com/thoughtpolice/buck2/commit/62d892d3ea70b7b3d82604aadd70c0613575a375 looks like it makes things worse, while int_roundings is likely to be stabilised. In contrast, https://github.com/thoughtpolice/buck2/commit/c69f195e0e85d9de994358723bd1f7a40f0143eb is just clearly cleaner code with less extension.

ndmitchell:
I dropped some thoughts in https://github.com/facebook/buck2/issues/265 so suggest we figure it out there

thoughtpolice:
Yeah, and I didn't put up https://github.com/thoughtpolice/buck2/commit/62d892d3ea70b7b3d82604aadd70c0613575a375 yet actually because I was wondering if an analogue to `div_ceil` should just be put in Gazebo, or something

ndmitchell:
Seems weird to put something that is soon going to be in the standard library in Gazebo. Putting it in buck2_core might be feasible, but as far as I remember, host stuff doesn't depend on buck2_core (for good internal reasons).

thoughtpolice:
Yeah, I expect it will be in the standard library in a few months (there seems to be one or two final nits but the Rust core people are busy). Like I said, a bit of it is just getting a feel for what will be big changes and what will be small changes. The `div_ceil` thing is kind of funny though; it's a good example of where like, if someone had added this function somewhere else, changing it wouldn't be a big deal, but because the addition *and* the usage are happening all at once, it's easy to think whether it's worth it at all or just redundant. But it's not an important patch, either way.

thoughtpolice:
In light of #292 happening, I think my need for this has been lowered a bit, and I don't think it's as big a deal as long as we have statically linked binaries for users to use; the need for distributions etc is a bit different than what this is catering to.

If the Meta team decides to move to rustc stable, it'll be loudly announced anyway, so I doubt it's a big deal.

I might still submit the occasional cleanup for things like this.

PR #337, opened 2023-07-11
feat(cargo): set codegen-units=1 for the release build

thoughtpolice:
Summary: I experimented with this option back when #133 was opened, and after review it can reduce binary size in --release mode by about 25% -- from 101MiB to 79MiB, while having a marginal build time increase (5m45s -> 6m20s) on my 12-core Ryzen 5600X.

While I suspect this will have little impact in practice on overall wall-clock time when running `buck2`, who doesn't love cheap and easy wins?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -368,6 +368,7 @@ buck2_eden = { path = "app/buck2_eden" }
 [profile.release]
 incremental = true
 panic = "abort"
+codegen-units = 1
 # The line below increases build times from 50s to 84s, with no observed impact on runtime speed.
 # But maybe for some benchmarks it will show more noticeable variation.
 # lto = "thin"


thoughtpolice:
I'd be interested to see if this has any impact on the builds at Meta (it's probably easy enough to replicate in your own build I guess), but I suspect the answer is probably "no": from my chats with Neil, if memory is the limiting factor, the relatively marginal reduction in `.text` size probably won't make much difference; any reduction in frontend stalls or I-cache pressure will be masked entirely by DRAM transfer latency. But hey: 30MiB for free!

ndmitchell:
Just tried this on todays version with our complete upload scripts - goes from 108Mb to 80Mb. Landing a change through the internal diff review process. We're in a code freeze right now, so will take a week or so to land internally.

PR #342, opened 2023-07-13
refactor(buck2_common): don't overload the user with HTTP retry backoff warnings 

thoughtpolice:
Summary: Another attempt at GH-316 after feedback given on GH-321 by @krallin. This introduces a new trait to dispatch these warnings, and two trait impls: one used in the materializer codepath that now writes buck events to the log, and another one used in the manifold client path to just warn to the console, like we do now.

Test Plan: Build a project that uses `reindeer` with `buck2`, as that causes a lot of warnings. Observe lots of warnings. Apply patches. Rebuild with new buck2. Observe no warnings. Run `buck2 log show | grep HttpRequestRetried` and see that all the new log events actually did get emitted and go where intended.

Fixes GH-316.

### DIFF
--- a/app/buck2_client_ctx/src/manifold.rs
+++ b/app/buck2_client_ctx/src/manifold.rs
@@ -22,6 +22,7 @@ use buck2_common::http::find_certs::find_tls_cert;
 use buck2_common::http::retries::http_retry;
 use buck2_common::http::retries::AsHttpError;
 use buck2_common::http::retries::HttpError;
+use buck2_common::http::retries::NoopDispatchableHttpRetryWarning;
 use buck2_common::http::HttpClient;
 use buck2_core::fs::paths::abs_path::AbsPath;
 use bytes::Bytes;
@@ -467,6 +468,8 @@ impl ManifoldClient {
         }
 
         let res = http_retry(
+            &url,
+            NoopDispatchableHttpRetryWarning::new(),
             || async {
                 self.client
                     .put(&url, buf.clone(), headers.clone())
@@ -499,6 +502,8 @@ impl ManifoldClient {
         );
 
         let res = http_retry(
+            &url,
+            NoopDispatchableHttpRetryWarning::new(),
             || async {
                 self.client
                     .post(&url, buf.clone(), vec![])
--- a/app/buck2_client_ctx/src/manifold.rs
+++ b/app/buck2_client_ctx/src/manifold.rs
@@ -21,8 +21,8 @@ use buck2_common::http::counting_client::CountingHttpClient;
 use buck2_common::http::find_certs::find_tls_cert;
 use buck2_common::http::retries::http_retry;
 use buck2_common::http::retries::AsHttpError;
+use buck2_common::http::retries::DispatchableHttpRetryWarning;
 use buck2_common::http::retries::HttpError;
-use buck2_common::http::retries::NoopDispatchableHttpRetryWarning;
 use buck2_common::http::HttpClient;
 use buck2_core::fs::paths::abs_path::AbsPath;
 use bytes::Bytes;
@@ -429,6 +429,25 @@ pub struct ManifoldClient {
     manifold_url: Option<String>,
 }
 
+struct ManifoldDispatchableHttpRetryWarning {}
+
+impl ManifoldDispatchableHttpRetryWarning {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl DispatchableHttpRetryWarning for ManifoldDispatchableHttpRetryWarning {
+    fn dispatch(&self, backoff: &Duration, retries: usize, url: &str) {
+        tracing::warn!(
+            "HTTP request to {} failed; {} retries. Retrying in {} seconds",
+            url,
+            retries,
+            &backoff.as_secs(),
+        );
+    }
+}
+
 impl ManifoldClient {
     pub fn new() -> anyhow::Result<Self> {
         Ok(Self {
@@ -469,7 +488,7 @@ impl ManifoldClient {
 
         let res = http_retry(
             &url,
-            NoopDispatchableHttpRetryWarning::new(),
+            ManifoldDispatchableHttpRetryWarning::new(),
             || async {
                 self.client
                     .put(&url, buf.clone(), headers.clone())
@@ -503,7 +522,7 @@ impl ManifoldClient {
 
         let res = http_retry(
             &url,
-            NoopDispatchableHttpRetryWarning::new(),
+            ManifoldDispatchableHttpRetryWarning::new(),
             || async {
                 self.client
                     .post(&url, buf.clone(), vec![])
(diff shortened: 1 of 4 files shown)


app/buck2_data/src/lib.rs
```diff
@@ -220,3 +221,27 @@ pub mod serialize_duration_as_micros {
...
+
+impl NoopDispatchableHttpRetryWarning {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl DispatchableHttpRetryWarning for NoopDispatchableHttpRetryWarning {
+    fn dispatch(&self, _backoff: &Duration, _retries: usize, _url: &str) {}
+}
```
thoughtpolice, line 247: All of this should probably go into `buck2_common` instead of `buck2_data`? I merely put it there since I wrote it when I modified `data.proto`, so it was kind of coincidental.

thoughtpolice:
Actually, I have no idea why this isn't just a higher order function that's passed to `http_retry`? We don't need a trait at all?

thoughtpolice:
Actually, I guess the trait isn't too bad, I just kind of hate the boilerplate-ness of this in particular, which we do three times here:

```rust
struct F {}

impl F {
    fn new() -> Self {
        Self {}
    }
}
```

But surely there's a shorter way to just make this `impl` one-shot on the struct declaration with a `derive` thingy? Rust programmers love those, I think.

I could also remove the `Noop` variant for the dispatch warning trait, though I'd have to rebase the history a bit, which is fine. I just wanted something to use when introducing the API patch-by-patch.

thoughtpolice:
> But surely there's a shorter way to just make this `impl` one-shot on the struct declaration with a `derive` thingy? Rust programmers love those, I think.

This is apparently the `Default` trait, so I can replace all these with that, I think.

app/buck2_common/src/http/retries.rs
```diff
@@ -45,15 +45,46 @@ pub trait AsHttpError {
...
+
+/// No-op implementation of DispatchableHttpRetryWarning. This does nothing at
+/// all when a retry occurs; useful for mock tests.
+pub struct NoopDispatchableHttpRetryWarning {}
+
+impl NoopDispatchableHttpRetryWarning {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
```
krallin, line 66: (same as above)

app/buck2_common/src/http/retries.rs
```diff
@@ -45,15 +45,46 @@ pub trait AsHttpError {
...
+
+impl DispatchableHttpRetryWarning for NoopDispatchableHttpRetryWarning {
+    fn dispatch(&self, _backoff: &Duration, _retries: usize, _url: &str) {}
+}
+
+pub async fn http_retry<Exec, F, T, E, R>(
+    url: &str,
+    dispatch_retry_warning: R,
```
krallin, line 74: (nit) passing `&R` might make a bit more sense considering that' all the API requires

app/buck2_common/src/http/retries.rs
```diff
@@ -45,15 +45,46 @@ pub trait AsHttpError {
...
 
-pub async fn http_retry<Exec, F, T, E>(exec: Exec, mut intervals: Vec<Duration>) -> Result<T, E>
+/// Used to dispatch warnings to the event log in the case in the case of a
+/// retryable HTTP error. This often occurs when some level of exponential
+/// backoff is happening due to a lot of concurrent http_archive() requests for
+/// a single host; see GH-316 and GH-321 for background.
+pub trait DispatchableHttpRetryWarning {
+    /// Fire off a warning. This might go to the console, event log, or
+    /// some other place.
+    fn dispatch(&self, dur: &Duration, retries: usize, url: &str);
```
krallin, line 55: Can we rename `retries` ->  `attempts`? The first time this function is called the value will be 1 so it's more accurately attempts than retries

app/buck2_client_ctx/src/manifold.rs
```diff
@@ -428,6 +429,25 @@ pub struct ManifoldClient {
...
+impl ManifoldDispatchableHttpRetryWarning {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl DispatchableHttpRetryWarning for ManifoldDispatchableHttpRetryWarning {
+    fn dispatch(&self, backoff: &Duration, retries: usize, url: &str) {
+        tracing::warn!(
+            "HTTP request to {} failed; {} retries. Retrying in {} seconds",
```
krallin, line 443: (see my comment below, it's not actually retries, it's attempts)

app/buck2_execute/src/materialize/http.rs
```diff
@@ -99,8 +101,36 @@ impl AsHttpError for HttpDownloadError {
...
+            backoff_duration_secs: dur.as_secs(),
+            url: url.to_owned(),
+        };
+
+        match buck2_events::dispatch::get_dispatcher_opt() {
+            Some(dispatcher) => {
+                dispatcher.instant_event(event);
+            }
+            None => {
+                tracing::warn!("Failed to dispatch HttpRequestRetried event: {:?}", event)
```
krallin, line 124: So here what we actually would like to call is this function:

```rust
pub fn broadcast_instant_event<E: Into<buck2_data::instant_event::Data> + Clone>(
    event: &E,
) -> bool {
    let mut has_subscribers = false;

    for cmd in ACTIVE_COMMANDS.lock().values() {
        cmd.dispatcher.instant_event(event.clone());
        has_subscribers = true;
    }

    has_subscribers
}
```

(as-is unfortunately the logging wouldn't work when using the deferred materializer)

This function lives in `buck2_server`, so what we need to do is allow this to be injected into the callsites. 

Maybe the easiest way to do this is to simply have the `HttpClient` provide its logging method (because this is already threaded through to all the places you need it), via a method like:

```
trait HttpClient {
  ..

  fn retry_logger(&self) -> &dyn DispatchableHttpRetryWarning;
}
```

You probably should actually keep the "log to stderr" implementation as just the default implementation for this method on `HttpClient`. This way all you need to do is just wrap the client being constructed for the server with the event-logging code.

So, putting it all together, what we need to do is:

- Move this code to `buck2_server`
- Have it implement the broadcast
- When we create the server's http client (in `DaemonState::init_data`), we wrap it with some new struct implementing HttpClient that uses the event-logging implementation
- When we create ManifoldClient's http client, you inject the other one in
- When calling http_retry you get the retry logging from `HttpClient`

I can also make some of those changes if you'd like, just let me know

krallin (changes requested):

ndmitchell:
@thoughtpolice - are you planning to follow up on this or should we close it?

thoughtpolice:
Yes, I haven't gotten back to this, but it still is on my radar and I'd like to finish it up at some point. But if someone else gets to it or whatnot, we can close this.

thoughtpolice:
Closing this for now to clean out PR clutter, as it's hopelessly obsolete. There are other things people are doing here anyway, like HTTP ratelimiting, and my own fork has a patch to suppress this.

PR #343, opened 2023-07-14
chore(nix): use mold for linking inside 'nix develop' shell

thoughtpolice:
Summary: when using Linux, and developing with `nix develop` or `direnv`, go ahead and enable mold as the default linker.

We modify `RUSTFLAGS` in the shell environment to do this so that we don't have to fiddle with the internal `.cargo/config.toml` file; and because it's only for development anyway, this is strictly used only while developing, not while running `nix build`.

On my Ryzen 5600X, versus the default linker settings, this takes the link time for the (debug) buck2 executable from 32s to 3s.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1687709756,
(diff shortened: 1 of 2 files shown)


thoughtpolice:
The macOS test failure just seems to be a fluke.

Issue #112, opened 2023-04-07
Feature request: buck2 vscode extension


jmhodges:
(This is totally not meant as a "do this today" thing, but a placeholder here for future possible work.)

It would be nice for there to be a Buck2 vscode extension. Syntax highlighting for Buck2-flavored Starlark, support for running buck2 commands later on, maybe some nice auto-import, that kind of stuff.

The [official Bazel VSCode extension](https://marketplace.visualstudio.com/items?itemName=BazelBuild.vscode-bazel) is similar enough to actually cause interesting problems. Namely, that, right now, popping open a Buck2 Skylark `.bzl` file will trigger the loading of the Bazel extension which will, in turn, emit errors and such about missing WORKSPACE files and so on that are irrelevant in Buck2-land.

It'd be nice to get Starlark syntax highlighting, at least, without the noise, even if we don't get all the rest.

ndmitchell:
There is actually LSP support baked into Buck2 - just run `buck2 lsp`. I'm using that internally with the Starlark extension https://github.com/facebook/buck2/tree/main/starlark-rust/vscode and the options `"starlark.lspPath": "buck2"` and `"starlark.lspArguments": ["lsp"]`. It could certainly be cleaned up to work better out of the box.

MadcapJake:
What kind of lsp functionality is it capable of? It doesn't seem like to Starlark vscode plugin is implementing any vscode other than a debugger that calls [`starlark --dap` ](https://github.com/facebook/buck2/blob/81e93ea71a68d41f09752a97e0e4412557010734/starlark-rust/vscode/package.json#L69-L73).

PR #383, opened 2023-08-10
Initial fixup of the VSCode extension

### LINKED ISSUE #309: The vscode extension should be cleaned up and released
I think it would be nice if the vscode extension was released and polished and usable for Buck2. I have it installed locally, but it's a chore especially when using features like remote vscode.

What would be needed for this? Ideally it would be released through GitHub as well, which would be really nice to help people contribute.

A tricky thing is how much buck2-specific "stuff" should be in it. The extension is distributed with the starlark package, but right now buck2 is the main user. Ideally, buck2 *specific* integration would include things like running or listing build targets...

mempler:
This PR is just fixing up and cleaning up the old VSCode extension with an at least initial attempt to fix #309 and #112 

This will make the extension a bit more useful than it was before.
allthough debugging capabilities do not work right now.

Much more work in the future will be required to properly hook up hovering and debugging.


starlark-rust/vscode/.vscode/extensions.json
```diff
@@ -0,0 +1,5 @@
+{
+  // See http://go.microsoft.com/fwlink/?LinkId=827846
+  // for the documentation about the extensions.json format
+  "recommendations": ["dbaeumer.vscode-eslint", "amodio.tsl-problem-matcher"]
```
ndmitchell, line 4: Is this a good thing to include? Do we really want to recommend it? Or is this normal for a vscode package?

starlark-rust/vscode/package.json
```diff
@@ -49,59 +46,21 @@
...
         "grammars": [
             {
                 "language": "starlark",
                 "scopeName": "source.starlark",
-                "path": "./syntaxes/starlark.tmLanguage.json"
-            }
-        ],
-        "breakpoints": [
-            {
-                "language": "starlark"
```
ndmitchell, line 64: Why delete the debugger/breakpoint features?

starlark-rust/vscode/package.json
```diff
@@ -49,59 +46,21 @@
...
-                ]
+                "path": "./syntax/starlark.tmLanguage.json"
             }
         ],
         "configuration": {
             "properties": {
                 "starlark.lspPath": {
                     "type": "string",
-                    "default": "starlark",
+                    "1fault": "buck2",
```
ndmitchell, line 63: Intentional? `1fault`

ndmitchell (commented):
Thanks for the changes. Sorry we didn't notice them sooner - the notifications went to someone on holiday.

As for the changes, it seems there is a switch from spaces to tabs - I thought the expectation was spaces everywhere now? It also seems to use `buck2` rather than `starlark` as the driver. I'm honestly not sure what to do about that - we deliberately don't want to couple the starlark-rust project to Buck2, but it is the biggest user. Maybe there should be a switch? Maybe it should try and find if buck2 or starlark is on the path?

Is the nix thing entirely optional? By removing the .lock directories, doesn't that risk the breakage as packages upgrade?

starlark-rust/vscode/package.json
mempler: Due to lack of implementation within `buck2 lsp` I noticed later that they were under a different command. Didn't had time yet to reimplement them

mempler:
> Thanks for the changes. Sorry we didn't notice them sooner - the notifications went to someone on holiday.

happens


> As for the changes, it seems there is a switch from spaces to tabs - I thought the expectation was spaces everywhere now? It also seems to use `buck2` rather than `starlark` as the driver. I'm honestly not sure what to do about that - we deliberately don't want to couple the starlark-rust project to Buck2, but it is the biggest user. Maybe there should be a switch?

There were some talks about this in the mentioned issue, thus I thought that was the right direction. I obviously do not have access to any internal communication that happen.

> Maybe it should try and find if buck2 or starlark is on the path?

Right now, it uses the one that is already in the `PATH` variable.

> Is the nix thing entirely optional? By removing the .lock directories, doesn't that risk the breakage as packages upgrade?

The nix thing is indeed entirely optional. I used nixos at the time of implementation.

starlark-rust/vscode/package.json
mempler: Typo

mempler:
Due to lack of interest from my side, I'll close this PR. however I'll keep my fork and branch around in case anyone wants to Pickup on it.

mempler:
I've decided to publish my own third-party extension, it's source available under https://github.com/PeerStudios/buck2-lsp-adapter and published under https://marketplace.visualstudio.com/items?itemName=PeerStudios.buck2-lsp-adapter which requires buck2 itself instead of starlark-rust. similarly to what i did in this PR.

Issue #374, opened 2023-08-01
Merging compilation database entries should should use argfiles


zjturner:
Windows has a command line limit length, and when you try to build a compilation database for a target with many object files, you pass this threshold and the `make_comp_db.py` script fails.

```
PS C:\Users\Zachary Turner> D:
PS D:\> cd .\src\foo\
PS D:\src\foo> buck2 build //Client/Base:Base[compilation-database]
Action failed: root//Client/Base:Base (cxx_compilation_database_merge)
Local command returned non-zero exit code 1
Reproduce locally: `"buck-out\\v2\\gen\\prelude\\23b3e4bdecafe14a\\python_bootstrap\\tools\\__win_python_wrapper__\\win_ ...<omitted>... 23b3e4bdecafe14a\\Client\\Base\\__Base__\\__comp_db__\\src\\Microsoft\\Environment.cpp.comp_db.json" (run `buck2 log what-failed` to get the full command)`
stdout:
stderr:
The command line is too long.
Build ID: 1c6dffb0-30ea-4f2c-924d-e49a3294256f
Network: Up: 0 B  Down: 17 MiB
Jobs completed: 145. Time elapsed: 1.4s.
Cache hits: 0%. Commands: 86 (cached: 0, remote: 0, local: 86)
BUILD FAILED
```

This looks like a similar problem to what I reported in #364, where we need to use argfiles.  Incidentally, `make_comp_db.py` actually already supports argfiles:

https://github.com/facebook/buck2/blob/a8060c162b1c63e56a1b0a13c72971f72a5799ba/prelude/cxx/tools/make_comp_db.py#L36-L43

But the `create_compilation_database()` function does not pass the arguments in as an argfile:

https://github.com/facebook/buck2/blob/a8060c162b1c63e56a1b0a13c72971f72a5799ba/prelude/cxx/comp_db.bzl#L58-L63

leading to the above error.

PR #390, opened 2023-08-16
Use an argsfile to build the compilation database

andyfriesen:
Fixes https://github.com/facebook/buck2/issues/374

I was unable to find any tests for this.  Are there any?

### DIFF
--- a/prelude/cxx/tools/make_comp_db.py
+++ b/prelude/cxx/tools/make_comp_db.py
@@ -54,8 +54,16 @@ def merge(args):
 
     entries = []
     for entry in args.entries:
-        with open(entry) as f:
-            entries.append(json.load(f))
+        if entry.startswith('@'):
+            with open(entry[1:]) as argsfile:
+                sub_entries = map(str.strip, argsfile.readlines())
+
+                for sub_entry in sub_entries:
+                    with open(sub_entry) as g:
+                        entries.append(json.load(g))
+        else:
+            with open(entry) as f:
+                entries.append(json.load(f))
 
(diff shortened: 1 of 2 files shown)


ndmitchell:
Sorry for the delay in responding to this - our internal tasking system sent it to someone on holiday...

We do have internal tests for this, but they aren't yet open source. What's the reason for preferring an args file here? I think the way we currently use these files is to turn them into an args file when the person is consuming them, rather than writing them down, but I'll talk to the people who wrote this feature.

prelude/cxx/tools/make_comp_db.py
```diff
@@ -54,8 +54,16 @@ def merge(args):
 
     entries = []
     for entry in args.entries:
-        with open(entry) as f:
-            entries.append(json.load(f))
+        if entry.startswith('@'):
```
ndmitchell, line 57: Is this actually required, given that we have support for `@` files in `gen`?

andyfriesen: The two modes that this script can run in don't share any code at all, so some argfile handling mechanism is necessary here as well.

We could lift it into a small utility function.  Basically a function that goes from an argv that potentially contains an argsfile reference to one that has them all expanded.

Would that be better?

ndmitchell: Given it's two entirely separate functions that just happen to live in the same file, I'd just leave it for now. I'll land this as is, and we can potentially common up later.

ndmitchell:
Turns out there are tests which caught a small bug - see the small change I had to make.

Issue #368, opened 2023-07-28
Precompiled headers don't seem to be working


zjturner:
Found this in the prelude:

https://github.com/facebook/buck2/blob/776d355f7e6136de88fa3b7c28e775305826ac3f/prelude/cxx/compile.bzl#L471-L474

```
    # Workaround as that's not precompiled, but working just as prefix header.
    # Another thing is that it's clang specific, should be generalized.
    if ctx.attrs.precompiled_header != None:
        args.add(["-include", headers_tag.tag_artifacts(ctx.attrs.precompiled_header[CPrecompiledHeaderInfo].header)])
```

So it doesn't appear to actually be precompiling anything (even on Linux / Clang).

The prelude is pretty complicated, but I'm willing to take a stab at implementing this if someone can offer some pointers.  I need:

1. To create the step that actually runs the compiler to generate the compiled PCH output file
2. To reference the generated PCH output artifact created in step 1.
3. To create a serialization point in the DAG from step 1 to step 2.

I also thought about implementing this by just making a genrule() target that outputs the PCH and referencing it from the `compiler_flags` argument as a workaround.

ndmitchell:
I added #392 which is an internal implementation of PCH that wasn't merged because performance is slower for .h files, faster for .cpp files. I did not write this code, and had to rebase this code over a few months, so I checked I can still compile a C++ file, but haven't checked I didn't break PCH.

zjturner:
I'll test this out for science.  However, just want to mention that to really get the best benefit out of PCH, you need to be able to specify the header list.  Biggest wins from PCH come from compiling a PCH once per library, and being able to explicitly choose what headers go into the PCH.  With PCH you really want headers that are expensive to compile, but don't change frequently because that reduces the impact on .h file iteration while still greatly improving .cpp file iteration.  And if you can limit it to files that change basically never (e.g. third party deps, libc++, boost, webrtc, etc) then it's pure win.  But only the individual library knows that.

My ideal design is something like this:

```
cxx_precompiled_header(name = "foo", headers = [header-list])
cxx_library(name = "bar", pch = ":foo")
```

`buck2 build //foo` should do no work, because it doesn't know what flags to pass when generating the precompiled header.  But `buck2 build //bar` should first run the compiler with the same flags that its using for its own sources but outputting a PCH, and then it should compile the rest of the sources in parallel taking the compiled pch output as input to each object.

ndmitchell:
[...]

PR #392, opened 2023-08-17
Add basic support for per-source auto-PCH

ndmitchell:
Summary:
This adds opt-in support to automatically generate a PCH for each C++
source (based on bmaurer's investigation in this area) -- by extracting
its leading `#include`s into a header and pre-compiling it -- which can
dramatically improve the compile-cycle time when iterating on `.cpp`s
~50% (the subsequent diff in this stack uses `-fpch-codegen`/
`-fpch-debuginfo`, which makes this faster w/ tradeoffs, as it can eagerly
generate more code than w/o PCH).

In basic tests, this improves the `.cpp` iteration workflow ~50% while
regressing the `.h` iteration workflows by up to ~30% (presumably due to
the overhead in serializing/deserializing the AST).  A followup here is to
support some way to "race" both approaches in incremental builds to get
the best of both worlds.

TODOs:
1) We currently re-generate the source to strip the leading `#includes`
   that were put in the PCH, as using the original source fails
   compilation w/ conflicts.  We need to figure out if theres some way
   to avoid this or, if it is needed, make sure we're reporting the original
   `.cpp` in error messages and not breaking cases of source-relative
   includes.
2) The `#include` extractor utility is really basic and likely needs some
   work (e.g. it can't walk past `#if*`s).

Differential Revision: D46329995


ndmitchell:
Exported as part of discussions on #368 with @zjturner - not intended to be merged as is right now

PR #398, opened 2023-08-26
Added support for strip-prefix when unzipping in Linux

Overhatted:
I have this target:
```
http_archive(
    name = "SQLiteFile",
    urls = [
        "https://www.sqlite.org/2021/sqlite-amalgamation-3350300.zip",
    ],
    sha256 = 'a629d0b1cc301347109e8ad211ff46af371b6ef73c41b7698e9cf1fb37bf4b95',
    sub_targets = [
        "sqlite3.c",
        "sqlite3.h",
    ],
    strip_prefix = "sqlite-amalgamation-3350300",
)
```
But it didn't work on Linux because unzip doesn't support strip_prefix. I added support by simply copying the sub-directory after unzipping.
I wasn't able to use symlinks because the subtargets attribute wouldn't work with them.

I wasn't able to find the tests for the rules code. If it exists, please let me know where I can find it.

### DIFF
--- a/prelude/http_archive/http_archive.bzl
+++ b/prelude/http_archive/http_archive.bzl
@@ -43,11 +43,12 @@ def _type(ctx: AnalysisContext) -> str:
         fail("unsupported archive type: {}".format(typ))
     return typ
 
+#Returns a two-element tuple with the first element being the cmd_args with the unarchive command and the second a bool indicating whether the prefix still needs to be stripped (in cases where the too used to uncompress does not support this feature).
 def _unarchive_cmd(
         ext_type: str,
         exec_is_windows: bool,
         archive: Artifact,
-        strip_prefix: [str, None]) -> cmd_args:
+        strip_prefix: [str, None]) -> tuple:
     if exec_is_windows:
         # So many hacks.
         if ext_type == "tar.zst":
@@ -66,7 +67,7 @@ def _unarchive_cmd(
                 "-f",
                 "-",
                 _tar_strip_prefix_flags(strip_prefix),
-            )
+            ), False
         elif ext_type == "zip":
             # unzip and zip are not cli commands available on windows. however, the
             # bsdtar that ships with windows has builtin support for zip
@@ -76,7 +77,7 @@ def _unarchive_cmd(
                 "-f",
                 archive,
                 _tar_strip_prefix_flags(strip_prefix),
-            )
+            ), False
 
         # Else hope for the best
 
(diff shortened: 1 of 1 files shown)


prelude/http_archive/http_archive.bzl
```diff
@@ -47,7 +47,7 @@ def _unarchive_cmd(
         ext_type: str,
         exec_is_windows: bool,
         archive: Artifact,
-        strip_prefix: [str, None]) -> cmd_args:
+        strip_prefix: [str, None]) -> tuple.type:
```
ndmitchell, line 50: I think you can write this as `tuple[cmd_args, bool]`. When there is a complex return type it's often useful to document on the function what the return type means.

prelude/http_archive/http_archive.bzl
```diff
@@ -175,24 +170,32 @@ def http_archive_impl(ctx: AnalysisContext) -> list[Provider]:
         mkdir = "mkdir -p {}"
         interpreter = ["/bin/sh"]
 
-    unarchive_cmd = _unarchive_cmd(ext_type, exec_is_windows, archive, ctx.attrs.strip_prefix)
+    unarchive_cmd, was_prefix_kept = _unarchive_cmd(ext_type, exec_is_windows, archive, ctx.attrs.strip_prefix)
```
ndmitchell, line 173: Is `needs_strip_prefix` a better name than `was_prefix_kept` - since if the prefix was kept because you didn't ask the prefix to be stripped it returns `False`, but the prefix was kept.

prelude/http_archive/http_archive.bzl
```diff
@@ -175,24 +170,32 @@ def http_archive_impl(ctx: AnalysisContext) -> list[Provider]:
...
+
+    output_name = value_or(ctx.attrs.out, ctx.label.name)
+    output = ctx.actions.declare_output(output_name, dir = True)
+    script_output = ctx.actions.declare_output(output_name + "Tmp", dir = True) if was_prefix_kept else output
```
ndmitchell, line 177: Pehraps `output_name + "_tmp"` - we tend not to generate CamelCase names

ndmitchell:
A few very minor notes. The CI is red because `tuple.type` should be `tuple` (it was a migration that probably happened after you started writing your code), but `tuple[cmd_args, bool]` is even better.

Looks good to me. Our internal test suite has not yet been open sourced. That said, the strip feature is woefully under-tested too. But honestly, we can live with that until the real test suite is open source, at which point there would be a good place to add a test. Until then, I wouldn't worry.

prelude/http_archive/http_archive.bzl
Overhatted: I tried it as `tuple[cmd_args, bool]` and it didn't work: https://app.circleci.com/pipelines/github/facebook/buck2/11065/workflows/07b21140-63be-4cd5-82e3-4a6d88c98cfa/jobs/28597

From searching the rest of the code base it seems the only possibility is simply `tuple`.

prelude/http_archive/http_archive.bzl
Overhatted: That is definitely a better name

ndmitchell:
Thanks, I have merged this internally, and added a test case internally. Unfortunately our internal test cases for this function reference internal HTTP servers (they aren't particularly secret, but the tests would have zero value when open sourced). We'll work on cleaning that up, but just so you are aware your added functionality is now tested.

PR #400, opened 2023-08-27
rust-project: Fix un-truncated line endings with the rustup sysroot getter

cormacrelf:
Sysroots from rustup currently include the trailing newline, needs trimming. The .buckconfig-based sysroot setting is also broken without fb-specific tooling (xplat etc).


ndmitchell:
Thanks for the change! I'm going to find the person internally who worked on this to have them take a look.

davidbarsky:
Oh, that’s an oversight on my part. Sorry about that!

Issue #391, opened 2023-08-17
Experience trying to configure a wasm32 target platform


cormacrelf:
Let's say I want to make a hello world Rust/WASM binary. As a really basic start, you can at least compile a _library_ for wasm32, by overriding `rustc_target_triple` in the rust toolchain definition:

```diff
# toolchains/BUCK
 system_rust_toolchain(
     name = "rust",
     default_edition = "2021",
+    rustc_target_triple = "wasm32-unknown-unknown",
     visibility = ["PUBLIC"]
 )
```

But it isn't enough for a binary. Here are the things I had to fix to get a hello world rust wasm binary linked:

1. Obviously there is no wasm32-unknown-unknown in the big target-triple selector for rust. That's why I added the custom triple in the toolchain to begin with. But you could plausibly add more branches to the select.
2. To use with `--target-platforms` I tried adding `platform(name="wasm32", constraint_values=["prelude//cpu/constraints:wasm32", "prelude//os/constraints:none"])` to `platforms/BUCK`. Unfortunately this meant that `toolchains//:python_bootstrap` could no longer be configured, because it was looking for an OS among linux/macos/windows. Adding an explicit `interpreter = "python3"` to `system_python_bootstrap_toolchain` worked for this.
3. Rust linking a wasm32 binary uses the wrong linker. I tried this on macOS and fedora. The linker script just gives you a long list of unsupported linker options. `error: unsupported option '--export'`, etc. I hacked together a solution, basically you need to:
    1. Use `rust-lld` as the cxx toolchain's linker. i.e. `system_cxx_toolchain(..., linker = "rust-lld")`.
    2. add `rustc_flags = ["-C", "linker-flavor=wasm-ld"]` to the rust toolchain; and
    3. prevent the passage of `-fuse-ld=lld` to the linker (because rust-lld does not support that flag). You can do that by by editing `prelude/toolchains/cxx.bzl` to add an `elif ctx.attrs.linker == "rust-lld": pass` before the branch that adds `-fuse-ld=lld`.

Overall this was not that much work, but prelude needs some patches before this is usable beyond a proof of concept.

Here are some issues with all this:

1. The big confusion for me was the constraints -- it seems that the prelude's OS and CPU constraints are used for both the platform buck2 is executing actions on, and the target platforms it's building for.
[...]

PR #407, opened 2023-09-06
Make a bunch of python bootstraps exec deps

cormacrelf:
See https://github.com/facebook/buck2/issues/391#issuecomment-1707865763

Issue consists of:

- using a target platform with `config//os:none`, e.g. webassembly
- the `system_python_bootstrap_toolchain` rule's default value for the interpreter select()ing from `config//os:linux`/macos/windows, but no default value
- Some msvc tools using regular deps for their python bootstrap deps
- Using a system_cxx_toolchain, which does the same
- So all the python bootstrap rules are configured for the target platform which has os:none. And thus you can't configure a python bootstrap toolchain. You can obviously just set `interpreter="python3"` yourself to make the error go away, but you can't select() a proper solution in there because the configs that are set are based on the target platform.

There are more places python bootstraps are used without being exec deps. E.g. [here](https://github.com/facebook/buck2/blob/959555e5f6eed689d88deceb460096ebde590421/prelude/cxx/cxx_toolchain.bzl#L218) which seems very complicated.

I suppose this would also be fixed by using a transition to exec platform on all python bootstrap binaries? Surely nobody ever needs a python bootstrap binary configured for a target platform, it would be pretty useless given the limitations of the python bootstrap rules. Anyway, this works.

### DIFF
--- a/prelude/toolchains/cxx.bzl
+++ b/prelude/toolchains/cxx.bzl
@@ -181,9 +181,9 @@ system_cxx_toolchain = rule(
         "link_flags": attrs.list(attrs.string(), default = []),
         "link_style": attrs.string(default = "shared"),
         "linker": attrs.string(default = "link.exe" if host_info().os.is_windows else "clang++"),
-        "linker_wrapper": attrs.default_only(attrs.dep(providers = [RunInfo], default = "prelude//cxx/tools:linker_wrapper")),
-        "make_comp_db": attrs.default_only(attrs.dep(providers = [RunInfo], default = "prelude//cxx/tools:make_comp_db")),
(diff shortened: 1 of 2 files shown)


ndmitchell:
Thanks - this seems sensible to me. My guess is we managed to get away with this because they are Python which runs basically the same "everywhere". But of course, everywhere is not quite everywhere, e.g. WASM.

ndmitchell:
392c94105afadaddbe13f6871ab5dac35171c5ca undoes some of these because they aren't required - the deps of an exec_dep are already on the exec platform. If you transition again you end up on the exec platform of your exec platform. Often that isn't a problem and will work, but it's not technically correct. Let me know if it breaks your use case (hopefully it either keeps it the same or improves it)

cormacrelf:
It doesn't break it. But I do think that is very funny.

PR #409, opened 2023-09-09
Add a "wasm" LinkerType/linker_type.

cormacrelf:
Why?
1. We want to emit `.wasm` files, so the filenames need configuration. LinkerType is how we do that.

2. Wasm needs to avoid all the `== "darwin"`, `== "gnu"` checks throughout the cxx toolchain code.

3. There are new flags supported by `wasm-ld`, LLVM's wasm linker, which we may wish to support.

4. LLD has `-flavor wasm`. (The other linker_types map 1:1 to LLD flavors.  "windows" is just a much better name than `-flavor link` is. Etc.)

`-flavor wasm` isn't passed anywhere in buck's rules, but rustc natively sends that flag when it's assuming your linker is `rust-lld`. This is a bit of a roundabout way of configuring it; it would be nice to tell rustc "we got this" and configure every flag ourselves. I also got the impression that linker_type should simply be describing how the `linker` executable behaves, so both should be configured in tandem i.e. `cxx_toolchain_override(..., linker_type = "darwin", linker = RunInfo(args = cmd_args("lld", "-flavor", "darwin")))`. Let me know if I'm on the wrong track there.

I have a separate PR coming that improves the zig toolchain + sets the linker_type to a select(). I can't currently get zig to link it, but you can combine it with cxx_toolchain_override that sets the linker to rust-lld from a rustup distribution. (Buckifying rust-toolchain.toml and constructing a sysroot is also done, PR coming soon.)

For linkage, ref #391

### DIFF
--- a/prelude/cxx/cxx_toolchain_types.bzl
+++ b/prelude/cxx/cxx_toolchain_types.bzl
@@ -13,7 +13,7 @@ load("@prelude//cxx:debug.bzl", "SplitDebugMode")
 # executable can access those.
 RichLinkerRunInfo = provider(fields = ["exe", "flags"])
 
-LinkerType = ["gnu", "darwin", "windows"]
+LinkerType = ["gnu", "darwin", "windows", "wasm"]
 
 ShlibInterfacesMode = enum("disabled", "enabled", "defined_only")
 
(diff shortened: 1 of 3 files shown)


ndmitchell:
Thanks - looks good to me!

PR #410, opened 2023-09-10
fix cxx_toolchain_override not handling missing fields on base toolchains

cormacrelf:
The defaults for the `lto_mode` and `as_compiler_info` arguments to `cxx_toolchain_infos` are `None`. The zig cxx toolchain, for example, doesn't set them. These changes let you override the zig toolchain without erroring out in cxx_toolchain_override, which was trying to access fields on those Nones.


cormacrelf:
Hm, it seems there are more of these than I thought. Overriding system_cxx_toolchain is whack-a-mole.

ndmitchell:
Looks good to me! Are there any others that need doing?

PR #440, opened 2023-10-09
feat(conan): add conan toolchain import third-party C/C++ dependencies

aherrmann:
Closes #58

This is a rebase of #58. I no longer have push access to this repository and can't update #58 myself, so I created this separate PR to run CI on the changes. See #58 for a detailed description and motivation.


prelude/toolchains/cxx/zig/defs.bzl
```diff
@@ -307,7 +307,7 @@ def _cxx_zig_toolchain_impl(ctx: AnalysisContext) -> list[Provider]:
     return [ctx.attrs.distribution[DefaultInfo]] + cxx_toolchain_infos(
         platform_name = dist.arch,
         c_compiler_info = CCompilerInfo(
-            compiler = RunInfo(args = cmd_args([zig, "cc"])),
+            compiler = RunInfo(args = cmd_args([zig, "cc"], delimiter = " ")),
```
ndmitchell, line 310: Why are these required?

prelude/toolchains/cxx.bzl
```diff
@@ -27,7 +27,7 @@ def _system_cxx_toolchain_impl(ctx: AnalysisContext):
     """
     A very simple toolchain that is hardcoded to the current environment.
     """
-    archiver_args = ["ar", "rcs"]
```
ndmitchell, line 30: This seems like a good idea independently. I might try and land that separately, so that we have a revertable lump for each piece.

ndmitchell (approved):
This code is beautiful. Minor questions, but will try and import it and see why the error is occuring.

ndmitchell:
The import error is:

```
2023-10-15T10:18:36.176308Z ERROR cli_log: Error in changeset processing: Unknown error: This file path is illegal and cannot be imported: buck2/prelude/toolchains/conan/TARGETS.v2
```

I suspect this is a bug with our import/export configuration. Ours is a little tricky because the files are internally named TARGETS, and then renamed for export (we want to change the source of truth to `BUCK` eventually, once Buck1 dies). Perhaps if you change `prelude/toolchains/conan/BUCK` to `prelude/toolchains/conan/BUCK.v2` it might work?

prelude/toolchains/cxx/zig/defs.bzl
aherrmann: The question also came up on the previous PR :smile: https://github.com/facebook/buck2/pull/58#discussion_r1061631207. The reason is:

> The Conan toolchain configures Conan to use the C/C++ toolchain provided by Buck2. It does so by [setting the relevant variables](https://github.com/facebookincubator/buck2/pull/58/files#diff-636b3a34913c4a58b72558b8ef0f48f441101379984f3828fc7697b0c88c98c9R704) in the Conan profile. Some build tools assume that these contain a single path, instead of commands with arguments, so the toolchain [generates wrapper scripts](https://github.com/facebookincubator/buck2/pull/58/files#diff-636b3a34913c4a58b72558b8ef0f48f441101379984f3828fc7697b0c88c98c9R656-R667) that execute the required command. Without this change the command would be written on multiple lines despite the explicit [`delimiter = " "` at a higher level](https://github.com/facebookincubator/buck2/pull/58/files#diff-636b3a34913c4a58b72558b8ef0f48f441101379984f3828fc7697b0c88c98c9R639). Perhaps there is a better way to do this? As the comment there indicates I also tried setting `quote = "shell"`, but that had other unwanted effects.

aherrmann (commented):
Thank you for the review and trying the import!

> Perhaps if you change `prelude/toolchains/conan/BUCK` to `prelude/toolchains/conan/BUCK.v2` it might work?

I've pushed a commit to that effect. :crossed_fingers:

ndmitchell:
Nothing to be done about the PR failures - we've fixed them on master. I've successfully imported the diff and we are reviewing internally.

aherrmann:
I've tested the merged state and it works. Thank you for pushing this through!

PR #446, opened 2023-10-16
Remove unnecessary unstable feature `named-profiles`

weihanglo:
As of Oct 16 2023, the version of rust in use is `nightly-2023-07-10`. `named-profiles` has been stabilized in 1.57, which predates the toolchain in use.

I assume that we don't need this anymore.


ndmitchell:
Thanks! LGTM

PR #462, opened 2023-10-27
Fix the Zig toolchain and example

aherrmann:
Closes #461

Updates the Zig toolchain example to take recent Buck2 changes into account.
Wrap the Zig C/C++ compiler commands in scripts to avoid quoting errors.
Avoid `ln -r`, BSD `ln` does not support `-r`.
Test the Zig example on CI.

PR #464, opened 2023-10-27
Add additional zig releases to zig toolchain

jazzdan:
More recent version of zig are required to compile code on modern versions of macOS (at least 13.x, and probably 14.x as well), otherwise zig seems to be unaware of the macOS version and errors out.

I tested this on Linux and it still works. It gets _further_ on macOS, but runs in to more problems with the linker even after #462. I'll file a separate issue for that once this is accepted so that the error can be reproduced on the main branch.


ndmitchell:
Thanks!

aherrmann (commented):
Thank you! Looks good!

ndmitchell:
Thanks, merged!

Issue #222, opened 2023-05-01
Support symlinks output for Remote Execution


sluongng:
Currently, buck2 is wrapping RE's ActionResult with `TActionResult2` https://github.com/facebook/buck2/blob/e1408804ad7ec4a602a5b172fbf338483f48030f/remote_execution/oss/re_grpc/src/response.rs#L72-L83

This struct is missing `output_symlinks` and the 2 deprecated fields `output_file_symlinks` and `output_directory_symlinks` that allow the remote executor to return symlink outputs to the client.

In some special cases, `output_symlinks` could be used to create dangling symlinks (symlinks with target could not be resolved on the build environments) to support some packaging rules. I.e. an RPM may contain symlinks with targets that only exist in the production environment.

thoughtpolice:
This is one blocker (among others) for `buck2-nix` using RE currently, because some of my build rules output symlinks to `/nix/store`, which is another use case for this. That will definitely have to come sometime after #193 lands, of course.

krallin:
I think it shouldn't be very hard to include, and the callsites probably need little changing to account for it. We don't actually have a backend that supports those internally, however so that'd need a little testing somewhere else.

alexlian:
Reviewing and this should probably be closed with this being addressed with:

https://github.com/facebook/buck2/commit/fb974e4dbd76279603e855ed1fd1a786d761df98

PR #477, opened 2023-11-05
Implement missing parts of OSS RE caching

cormacrelf:
This enables action caching on OSS builds. There were two missing pieces:

- write_action_result, for writing to the action cache. This was the major piece; without it, all you have is a bunch of uploaded CAS blobs for various things, but no way to find them again from the action digest. Now that the action results are uploaded as well, buck can find them when it's querying for cached actions based on the calculated digest alone.
- upload_blob was a less-exercised piece of code, it was only used for uploading stdout and stderr, and even then, only if they were more than 50KiB. I have added a test for it, but not sure if `internal` is the right one for it to be in, or when these tests get run.

I've successfully used this implementation with BuildBuddy, both from Linux (using remote_enabled = True), and from macOS, with remote_enabled = False but caching still enabled. I have only tested it with RBE v2.3+ output_paths mode.

As an implementation note, I found it very difficult to even figure out that these parts weren't implemented. Buck2 was just saying "0% cache hits mate" and reporting that every `re_action_cache` (= attempts to fetch a cached action, not attempts to cache an executed action!) had failed. None of these failures were logged anywhere. I don't know what the appropriate mechanism to log this stuff is, but I whacked a tracing::warn for failed action result cache writes in there.

### DIFF
--- a/app/buck2_execute/src/re/client.rs
+++ b/app/buck2_execute/src/re/client.rs
@@ -320,7 +320,7 @@ impl RemoteExecutionClient {
 
     pub async fn upload_blob(
         &self,
-        blob: Vec<u8>,
+        blob: InlinedBlobWithDigest,
         use_case: RemoteExecutorUseCase,
     ) -> anyhow::Result<TDigest> {
         self.data
@@ -1051,10 +1051,12 @@ impl RemoteExecutionClientImpl {
 
     pub async fn upload_blob(
         &self,
-        blob: Vec<u8>,
+        blob: InlinedBlobWithDigest,
         use_case: RemoteExecutorUseCase,
     ) -> anyhow::Result<TDigest> {
-        self.client().upload_blob(blob, use_case.metadata()).await
+        let digest = blob.digest.clone();
+        self.client().upload_blob(blob, use_case.metadata()).await?;
+        Ok(digest)
     }
 
     async fn materialize_files(
--- a/app/buck2_execute/src/re/client.rs
+++ b/app/buck2_execute/src/re/client.rs
@@ -1143,6 +1143,9 @@ impl RemoteExecutionClientImpl {
                     ..Default::default()
                 },
             )
+            .inspect_err(|err| {
+                tracing::warn!("write_action_result failed: {err}");
+            })
             .await
     }
 }
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -41,11 +41,15 @@ use re_grpc_proto::build::bazel::remote::execution::v2::Digest;
 use re_grpc_proto::build::bazel::remote::execution::v2::ExecuteOperationMetadata;
 use re_grpc_proto::build::bazel::remote::execution::v2::ExecuteRequest as GExecuteRequest;
 use re_grpc_proto::build::bazel::remote::execution::v2::ExecuteResponse as GExecuteResponse;
+use re_grpc_proto::build::bazel::remote::execution::v2::ExecutedActionMetadata;
 use re_grpc_proto::build::bazel::remote::execution::v2::FindMissingBlobsRequest;
 use re_grpc_proto::build::bazel::remote::execution::v2::FindMissingBlobsResponse;
 use re_grpc_proto::build::bazel::remote::execution::v2::GetActionResultRequest;
 use re_grpc_proto::build::bazel::remote::execution::v2::GetCapabilitiesRequest;
+use re_grpc_proto::build::bazel::remote::execution::v2::OutputDirectory;
+use re_grpc_proto::build::bazel::remote::execution::v2::OutputFile;
 use re_grpc_proto::build::bazel::remote::execution::v2::ResultsCachePolicy;
+use re_grpc_proto::build::bazel::remote::execution::v2::UpdateActionResultRequest;
 use re_grpc_proto::google::bytestream::byte_stream_client::ByteStreamClient;
 use re_grpc_proto::google::bytestream::ReadRequest;
 use re_grpc_proto::google::bytestream::ReadResponse;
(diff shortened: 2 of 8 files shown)


cormacrelf:
The TODOs around symlinking are related to https://github.com/facebook/buck2/issues/222

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -771,7 +815,118 @@ impl REClient {
...
+        let tree_digest = tdigest_to(output_directory.tree_digest);
+        anyhow::Ok(OutputDirectory {
+            path: output_directory.path,
+            tree_digest: Some(tree_digest.clone()),
+            // TODO(cormacrelf): check whether buck2_execute::directory::directory_to_re_tree
+            // conforms with the requirements of passing `true` here (see .proto file)
+            is_topologically_sorted: false,
+        })
+    })?;
+    anyhow::Ok(ActionResult {
```
ndmitchell, line 902: Any reason to not just `Ok`?

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -771,7 +815,118 @@ impl REClient {
...
+        stderr_digest: stderr_digest.map(tdigest_to),
+    })
+}
+
 fn convert_action_result(action_result: ActionResult) -> anyhow::Result<TActionResult2> {
+    if !action_result.output_symlinks.is_empty()
+        || !action_result.output_file_symlinks.is_empty()
+        || !action_result.output_directory_symlinks.is_empty()
+    {
+        anyhow::bail!(
```
ndmitchell, line 925: We tend to avoid bail because it can be hard to reason about - easier to see the explicit return

remote_execution/oss/re_grpc/src/response.rs
```diff
@@ -91,6 +91,9 @@ pub struct TSubsysPerfCount {
 pub struct TActionResult2 {
     pub output_files: Vec<TFile>,
     pub output_directories: Vec<TDirectory2>,
+    // TODO: output_symlinks (use in preference when output_paths mode is used the execution side)
+    // TODO: output_file_symlinks (deprecated)
```
ndmitchell, line 95: Why mark this as a todo? Is it something we should add? Or is this just for when servers return deprecated stuff?

app/buck2_execute/src/re/client.rs
```diff
@@ -1141,6 +1143,9 @@ impl RemoteExecutionClientImpl {
                     ..Default::default()
                 },
             )
+            .inspect_err(|err| {
+                tracing::warn!("write_action_result failed: {err}");
```
ndmitchell, line 1147: Does this not appear in https://github.com/facebook/buck2/blob/main/app/buck2_execute_impl/src/executors/caching.rs#L108C25-L108C25 as a warning? It certainly should

ndmitchell (commented):
Thanks so much for this, and the great testing you have done3. Two things I'm still figuring out:

1. Why didn't you get a warn - I've noted where I expected that to pop out. Is that line not being hit for you?
2. Still figuring out if there are internal implications for upload blob taking a digest, or if it should have done so all along.

ndmitchell:
Having upload blob take a digest shouldn't be a big deal since I think the amount of data flowing through this API isn't big enough to matter. I can see why this is more convenient and matches the upload API better.

However, once we are going through that path, maybe we should just delete `upload_blob` entirely and reuse `upload`? That already takes a `InlinedBlobWithDigest` so then we could have fewer functions?

remote_execution/oss/re_grpc/src/client.rs
cormacrelf: It's just splatter from copy pasting the function below and inverting every operation, instead of coding it from scratch. I'll tidy this up, a few of these don't even need to be Result.

app/buck2_execute/src/re/client.rs
cormacrelf: That line may have run, but I definitely did not see it in the terminal. Thanks for the pointer, I will have a look at why.

cormacrelf:
> However, once we are going through that path, maybe we should just delete `upload_blob` entirely and reuse `upload`? That already takes a `InlinedBlobWithDigest` so then we could have fewer functions?

There is a lot of "code overhead" for each piece of GRPC client API you want to expose. There are about 5 layers of nesting that, for `fn upload_blob`, add exactly nothing at each layer. Each layer basically does "return this.field.upload_blob(...)". So it would certainly be economical to delete one of those function stacks. However I imagine there are opportunities to collapse the nesting and make it less costly to maintain in the first place. I might report back on that.

ndmitchell:
Some of these layers of forwarding are so we can inject the internal closed source RE client (closed source because it uses a lot of systems that aren't open source, rather than because it has magical properties). They might need to stay (or might not).

cormacrelf:
Hi @benbrittain, I haven't had any time to work on this recently. I've given you push access to my fork cormacrelf/buck2, so you can keep working on the rbe-writes branch.

miaoyipu:
> Is there anything I can do here to help push this to being landed? I'm happy to finish the PR if you'd like @cormacrelf

@cormacrelf  @benbrittain any update on this PR? This might help our RE execution efficiency, and I can help if no one is working on it.

benbrittain:
Hey @miaoyipu! It's on my near term TODO but I'm certainly not against you picking it up. If you do so though, let me know :)

avdv:
Hi @cormacrelf

> I've successfully used this implementation with BuildBuddy, both from Linux (using remote_enabled = True), and from macOS, with remote_enabled = False but caching still enabled. I have only tested it with RBE v2.3+ output_paths mode.

I am also trying to use this with Buildbuddy and disabled `remote_enabled` on Linux but I cannot get it to work. Could you share your config so I can compare that to mine? Thank you!

thoughtpolice:
@cormacrelf As of #912 landing, I think this PR is finally obsolete and all its functionality has been implemented!

cormacrelf:
Great work man! Thanks for plugging on.

PR #483, opened 2023-11-07
Killing processes with job objects on Windows

Overhatted:


app/buck2_forkserver/src/run/mod.rs
```diff
@@ -337,11 +395,19 @@ impl KillProcess for DefaultKillProcess {
...
         }
-        // `start_kill` is just `std::process::Child::kill` on Windows.
-        // Ignore the error because `kill` fails on Windows if the process has been terminated
-        // even if we did not wait for it.
-        let _ignore = child.start_kill();
-        Ok(())
+
+        #[cfg(windows)]
+        {
+            if true {
```
ndmitchell, line 401: Why `if true` ?

app/buck2_forkserver/src/run/mod.rs
```diff
@@ -337,11 +395,19 @@ impl KillProcess for DefaultKillProcess {
...
+                // On Windows we use Jobs to manage the processes so we don't
+                // use the default impl.
+                // We use `if true` here to do less conditional compilation
```
ndmitchell, line 404: I guess you mean to avoid some warnings?

app/buck2_forkserver/src/run/mod.rs
```diff
@@ -370,6 +436,20 @@ async fn kill_process_impl(
     }
 }
 
+#[cfg(windows)]
+fn kill_job(job_handle: &WindowsJobHandle) -> anyhow::Result<()> {
+    unsafe {
+        let terminate_job_result = winapi::um::jobapi2::TerminateJobObject(job_handle.get_handle(), 1);
+
+        if terminate_job_result == winapi::shared::minwindef::FALSE {
+            Err(anyhow::Error::msg("Failed to terminate job to kill"))
```
ndmitchell, line 445: We would usually use `anyhow::anyhow!("Failed to terminate job to kill")`

app/buck2_forkserver/src/run/mod.rs
```diff
@@ -444,6 +530,104 @@ where
...
+
+#[cfg(windows)]
+fn resume_process(child: &tokio::process::Child) -> anyhow::Result<()> {
+    let process_id = child.id().ok_or_else(|| anyhow::Error::msg("Failed to get the process id"))?;
+    let main_thread_id = get_main_thread(process_id)?;
+    resume_thread(main_thread_id)
+}
+
+#[cfg(windows)]
+fn get_main_thread(process_id: u32) -> anyhow::Result<winapi::shared::minwindef::DWORD> {
```
ndmitchell, line 562: I do wonder if these windows specific bits should be in `src/win`?

ndmitchell (commented):
Thanks so much!

Looks plausible at a first glance, but one of our windows experts should go through it in more detail. CC @lmvasquezg @KapJI

app/buck2_forkserver/src/run/mod.rs
Overhatted: I just copied the above code for Unix. I have no idea if I should add an `if true`  or not. The unix code is here: https://github.com/facebook/buck2/blob/22b925eec6e4da59a2d9c6ca202aa2c126c258fa/app/buck2_forkserver/src/run/mod.rs#L331

app/buck2_forkserver/src/run/mod.rs
Overhatted: I replaced all my usages of `anyhow::Error::msg` with `anyhow::anyhow!`.

ndmitchell:
Thanks so much!

PR #499, opened 2023-11-16
Fixed parameters declaration in targets command

Overhatted:
This was triggering the following assert:

```
thread 'main' panicked at 'Command targets: Argument or group 'package_values_regex' specified in 'conflicts_with*' for 'package-values' does not exist', C:\Users\<username>\.cargo\registry\src\index.crates.io-6f17d22bba15001f\clap-3.2.25\src\builder\debug_asserts.rs:237:13
```


scottcao (approved):

ndmitchell:
Thanks so much. I guess we rarely run any tests or anything with debug assertions on. I wonder if we should have at least a smoke test in debug mode to catch these kind of violations.

Issue #480, opened 2023-11-06
How to consume http_archive C++ headers?


jhdub23:
I'm experimenting with `http_archive`, I but can't figure out how to consume it in a simple C++ program.  I'm using the [buck2 getting started](https://buck2.build/docs/getting_started/) guide and incorporating [easylogging](https://github.com/abumq/easyloggingpp) as a simple test (header only version).

I modifed "main.cpp" to `#include <easylogging++.h>`, but I can't figure out the correct BUCK build file to get this to compile.  In the BUCK file, I added/modified:

```
http_archive(
  name = 'archive-easylogging',
  urls = [
    'https://github.com/abumq/easyloggingpp/releases/download/9.89/easyloggingpp_v9.89.tar.gz'
  ],
  sha256 = 'f91f2d3d560513c8a595f5ab0b10075e8fa7ee5eb92147a0925bc9a4183e0c48',
  visibility = ["PUBLIC"],
  sub_targets = ['easylogging++.h']
)

cxx_binary(
    name = "main",
...
    deps = [
        ":print",
        ":archive-easylogging",
    ],
    preprocessor_flags = ["-I$(location :archive-easylogging)"],
)
```

The above fails to compile as the `easylogging++.h` header file cannot be found.  I'm sure that I'm doing something wrong, but I'm new to buck/buck2/bazel.  The correct preprocessor_flags are not getting passed to the compiler.  I've searched the docs and issues, but I'm stumped.  Can someone show me the correct way to consume http_archives for C++?

zjturner:
What *is* being passed for `-I`?  Open your argsfile and see what its actually putting there.  `buck2 log what-failed` and then `cat` the argsfile that it references in the command line.

Your example looks correct to me (except that you dont' need the `sub_targets` attribute on the `http_archive` rule), and the same as what I'm doing in a situation where it works.

jhdub23:
Commenting out the preprocessor_flags line has no effect on the args file.  Additional trace:

```bash
>buck2 log what-failed
Showing commands from: buck2 build //:main
[...]

PR #501, opened 2023-11-18
Fixed the example hello_world library generation.

jhdub23:
The "print" library incorrectly included the main.cpp file in its list of sources. The main.cpp file should be used only as a source for the final executable.

This resolves issue #480.


ndmitchell:
Thanks for the fix!

PR #502, opened 2023-11-19
Add build info across all requests

vaskomitanov:


cormacrelf:
There is already a type defined by RBE for this, which https://github.com/facebook/buck2/pull/310 sends. You could pick up that PR and add the config flag that was requested.

thoughtpolice:
Yes, sorry for not getting around to it lately. I think #310 is the right way forward, but note that PR is very incomplete. Most of the plumbing for the RBEv2 metadata isn't there, so I think a lot of refactorings need to go in, first.

vaskomitanov:
PR 310 looks exactly what I needed, thanks for starting this effort @thoughtpolice. I will rebase my changes of yours and use the `correlated_invocations_id` for this purpose.

ndmitchell:
@vaskomitanov - are all your rebasings done now?

vaskomitanov:
> @vaskomitanov - are all your rebasings done now?

Yes. There are few fields not yet populated but even in this state we have all of the needed information for efficient scheduling for action execution and caching.
I can add more information in a subsequent PR.

thoughtpolice:
Thanks for taking this over the finish line! Please note that @krallin did I have one important adjustment he wanted me to make in https://github.com/facebook/buck2/pull/310#issuecomment-1612723781

> Could we preferably expose this via a config option rather than a build flag? We do use our internal RE in order to test the OSS build, and that doesn't build with fbcode_build, so this would break that testing.

So, we need to refactor this part of the diff in particular to just use a `buckconfig` value:

https://github.com/facebook/buck2/pull/502/commits/2eaa19cdeab2c782e258cb455c05ce1db70a122f#diff-0bb786b03f4f35aef10055270239e2afbf4f4b0ba0fa214b38ffa858c4bc9e95R1257-R1323

Maybe we should call it:

```
[buck2_re_client]
use_fbcode_metadata = true
```

And then default it to `false` so OSS users don't have to even know about it.

vaskomitanov:
> ```
> [buck2_re_client]
> use_fbcode_metadata = true
> ```
> 
> And then default it to `false` so OSS users don't have to even know about it.
OK, sorry for the wait (thanksgiving), I've made requested changes.
As a side note, to keep the current behavior, I've made `use_fbcode_metadata`= true. Can make it false by default if everybody thinks that should be the way to go.

Yashshiva75:
THIS IS A SAMPLE COMMENT

thoughtpolice:
> As a side note, to keep the current behavior, I've made `use_fbcode_metadata`= true. Can make it false by default if everybody thinks that should be the way to go.

IMO it should be turned off by default; after all, the code path in question is located inside the RPC client for the OSS Bazel API, so it's the exact path all OSS users rely on. This codepath should work by default in OSS builds.

We could `#[cfg]` that choice, but I'm sure Meta can override it with a simple `-c buck2_re_client.use_fbcode_metadata=true` flag to `buck2` in their test harness.

vaskomitanov:
> use_fbcode_metadata

done

vaskomitanov:
> > As a side note, to keep the current behavior, I've made `use_fbcode_metadata`= true. Can make it false by default if everybody thinks that should be the way to go.
> 
> IMO it should be turned off by default; after all, the code path in question is located inside the RPC client for the OSS Bazel API, so it's the exact path all OSS users rely on. This codepath should work as expected in OSS builds, i.e. with OSS metadata.
> 
> We could `#[cfg]` that choice, but I'm sure Meta can override it with a simple `-c buck2_re_client.use_fbcode_metadata=true` flag to `buck2` in their test harness.

OK, done, default is "false"

vaskomitanov:
@ndmitchell any additional steps I need to do to get this PR merged?

ndmitchell:
Sorry for the delay - was travelling then got a bit ill. Just running through the code now.

app/buck2_execute_impl/src/executors/action_cache.rs
```diff
@@ -179,9 +176,10 @@ async fn query_action_cache_and_download_result(
         re_use_case,
         digest_config,
         manager,
+        &identity,
         buck2_data::CacheHit {
             action_digest: digest.to_string(),
-            action_key,
+            action_key: Some(identity.action_key.clone()),
```
JakobDegen, line 182:
```suggestion
            action_key: if log_action_keys { Some(identity.action_key.clone()) } else { None },
```

To preserve existing behavior?

app/buck2_execute/src/re/uploader.rs
```diff
@@ -350,9 +357,13 @@ impl Uploader {
 
         // Upload
         let upload_res = if !upload_files.is_empty() || !upload_blobs.is_empty() {
+            let mut mtd = use_case.metadata();
+            if let Some(identity) = identity {
+                apply_identity(identity, &mut mtd);
+            }
```
JakobDegen, line 363: Hm, so I'm not the expert on this code, but I'm not super clear on the difference between the cases where this is applied vs the cases where it isn't? Is there a reason that some of these are more important than others?

I would've expected the `Option<&ReActionIdentity>` to instead become a mandatory parameter to `use_case.metadata()`, so that call sites have to explicit ask for the `None` behavior instead of letting this go forgotten

JakobDegen (commented):
Sorry that this took so long to get to. Overall this looks right, just a couple of thoughts

app/buck2_execute_impl/src/executors/action_cache.rs
vaskomitanov: Done, please see the updated PR

app/buck2_execute/src/re/uploader.rs
vaskomitanov: Great question! 
This PR covers the RE GRPC request sequence: `find missing blobs` -> `upload missing blobs` -> `execute action` -> `download produced artifacts`. 
This way on the server side we can be smart when replicating CAS artifacts and dispatching actions (we will be able to correlate`action execution` with preceeding `artifact uploads`). 
Without this change we are simply receiving 100s of parallel `find missing blobs` requests, 100s of parallel `upload missing blobs` and `execute action` without any correlation between them (even though they are correlated).
Other cases are not covered (can be covered in the future) because a) are not relevant regarding replication or request dispatching, b) will unnecessarily increase the grpc requests payload (in the header section).

JakobDegen: > Other cases are not covered (can be covered in the future) because a) are not relevant regarding replication or request dispatching

Leaving things not completely finished in this PR is fine, but just for clarity: What are the other cases? I guess I see action cache fetches and maybe `cas_artifact` actions. Anything I'm missing?

> will unnecessarily increase the grpc requests payload (in the header section).

I think even (and maybe especially) if this is intentionally not being filled out in some cases we should make that explicit in the code by requiring the parameter on `use_case.metadata()` and then writing a comment to explain why we pass `None` in some cases. Otherwise this will be forgotten about in the future

vaskomitanov: That makes sense. I’ve updated the code to address this concern.

JakobDegen:
The label was technically right that this was reverted, but I relanded it a day later so all good :)

Issue #452, opened 2023-10-21
How to install bxl outputs to specific folder


Overhatted:
I have the bxl function in the bottom of this post which I'm using to merge all the compile_commands.json in a repository, but the end result is in buck-out. Is there a way for me to move it to a specific location in the repository? For example something like:
`buck2 bxl create_compile_commands.bxl:create_compile_commands --out-dir .`

I could, of course, just copy the file with the shell but the fact that the output folder has a hash in its name makes it kind of complicated. Also, [here](https://buck2.build/docs/rfcs/bxl/#cpp-lsp), it seems to be implied that in the future we could set a bxl function to auto-update its output when its inputs have changed. Not sure if that's true (it would be very useful) but the shell solution would also not work in that situation.

More specifically, clangd searches for a compilation database in any of the parent directories so I would like to have buck2 copy the merged compilation database into the root of the repository. I can also point it to a specific sub-directory of the repository but buck2 has the hashes in the folder path making this difficult to maintain.

If copying of bxl outputs to a specific directory is not a current or planned feature, how is the LSP server given the compilation database inside META?

```
load("@prelude//cxx:cxx_toolchain_types.bzl", "CxxToolchainInfo")

def _create_compile_commands(ctx):
    bxl_actions = ctx.bxl_actions(
        toolchains = "toolchains//:cxx",
        target_platform = "prelude//platforms:default",
    )

    all_targets = ctx.cquery().kind("cxx", "...")

    content = cmd_args()
    for target, value in ctx.build(map(lambda x: x.label.with_sub_target("compilation-database"), all_targets)).items():
        content.add(value.artifacts()[0])

    argfile = bxl_actions.actions.declare_output("comp_db.argsfile")
    bxl_actions.actions.write(argfile.as_output(), content)

    db = bxl_actions.actions.declare_output("compile_commands.json")
    toolchain_cxx = ctx.unconfigured_sub_targets("toolchains//:cxx")
    cmd = cmd_args(bxl_actions.toolchains[toolchain_cxx][CxxToolchainInfo].mk_comp_db[RunInfo])
    cmd.add("merge")
    cmd.add("--flatten")
    cmd.add(cmd_args(db.as_output(), format = "--output={}"))
    cmd.add(cmd_args(argfile, format = "@{}"))
    #cmd.hidden(entries.values())
    bxl_actions.actions.run(cmd, category = "cxx_compilation_database_merge")

    ctx.output.ensure(db)

create_compile_commands = bxl_main(
[...]

Issue #307, opened 2023-06-26
Generated `compile_commands.json` uses relative paths for `directory` field


cbarrete:
Hi,

I'm having issues using the generated the output from the `compilation-database` subtarget, which seems to boil down to clangd not liking relative `directory` fields.

Here's how to reproduce the issue using `examples/with_prelude`:

```
> buck2 init --git
...
> buck2 build cpp/hello_world:main[compilation-database]
...
> find -name compile_commands.json
./buck-out/v2/gen/root/524f8da68ea2a374/cpp/hello_world/__main__/compile_commands.json
> ln -s ./buck-out/v2/gen/root/524f8da68ea2a374/cpp/hello_world/__main__/compile_commands.json .
> nvim cpp/hello_world/main.cpp
```

At this point clangd starts, but it generates a bunch of erroneous diagnostics and doesn't let me go to the definition of `print_hello`. If I edit `compile_commands.json` to make `directory` an absolute path and open `main.cpp` again, everything works.

Arguably, clangd could support relative paths, but looking around on the internet indicates that this is unlikely (also clang-tidy depends on absolute `directory` paths).

Could the `directory` path that Buck2 generates be made absolute?

Thanks!

ndmitchell:
The problem is that if you write out absolute paths, its no longer possible to cache the compilation database between users. It's fairly annoying.

knopp:
Is sharing compilation database between users a common thing to do? It's an artefact of build tool, why would you share it instead of generate per user? 

Another question is whether the utility of being able to share the compilation database outweigh not being able to use clangd, which in our case for example would be a complete dealbreaker.

ndmitchell:
Sharing between users is super common, since it speeds everything up.

That said, sharing a useless pile of bytes that doesn't work has no value. Maybe the toolchain should have a boolean as to whether it generates absolute paths (with no caching) or relative paths (which are cached). I'm trying to find out how we deal with this internally, since we use `clangd` somehow.

ndmitchell:
It seems like internally we use Bxl as a wrapper that takes the `compilation-database` target, turns the paths to absolute, and then uses that for `clangd`. A bit fiddly, but does ensure that everyone can share most of the work.

cbarrete:
> Sharing between users is super common, since it speeds everything up.

I would expect generating the compilation database to be very fast, is it actually worth speeding up via caching?
[...]

Issue #194, opened 2023-04-23
Create a compile_commands database for a cxx binary and its dependencies


lvaroqui:
Hello!

I set up a little C++ project in order to test Buck2 (I don't have any experience with Buck or Bazel).

One of the first thing I tried to do is setting up the generation of a compile_commands database for my IDE. I found the sub_target `compile_commands` but it seems it only allows generating the db for the binary and not its dependencies. Is it possible to do so? Also, I have to ask explicitly for the sub_target, but I would prefer to always update the db when the binary is updated, is this possible?

Thanks for your help!

ndmitchell:
There is a `compilation-database` subtarget on cxx_library and cxx_binary. Internally at Meta we have an LSP mode that when a file is opened, it uses the `owners` query (through Bxl) to find the right target, grabs the `[compilation-database]` for that target, and then uses that to drive the LSP. We haven't plumbed any of that into open source pieces. How do existing C++ LSP servers get the compilation database? If that is configurable, then it probably isn't too hard to hook it up to what Buck2 provides. But I don't think anyone has tried yet.

yeswalrus:
Chiming in since I'd love to see this (it's a huge pain in bazel). Generally speaking, clangd looks it up in a single monolithic compile_commands.json. Handling headers can be a challenge in that context. 
For good clangd support, you'd want to aggregate all the compilation database files together output to the source root, and include entries for both source & header files. See https://github.com/hedronvision/bazel-compile-commands-extractor/blob/main/ImplementationReadme.md for a detailed explanation of what's required (and some technical details about how it happens in bazel)

cpsauer:
^ Author of that Bazel tool here. Can confirm it was a pain :) 

(Arriving not because backlinks are my batsignal or something; I just happened to be checking out the status of buck2.)

@ndmitchell, thanks for all you guys are doing here to share this with the world. Could I add to the header-only question: How do you guys handle the case where a parent target configures its children? For example, where an iOS rule configures cxx_libraries to cross compile?

photex:
I use clangd a lot as my C++ lsp, and previously only ever configured it for a single compilation database. But after looking into their issue tracker  a bit, I saw that it was pointed out that you can alternatively configure clangd by paths in a repo:
https://clangd.llvm.org/config.html#if

[...]

PR #510, opened 2023-11-28
Add a bxl for generating a unified compile_commands.

zjturner:
This adds a simple bxl for generating a global compilation database for all targets in a build.  It is very fast, able to generate a database for around 4,000 files in under 4 seconds.

This opens the door for external users being able to integrate with traditional LSPs like clangd.

This addresses several outstanding issues for external users, including #194, #307, and #452.

### DIFF
--- a/prelude/cxx/tools/compile_commands.bxl
+++ b/prelude/cxx/tools/compile_commands.bxl
@@ -1,0 +1,97 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is licensed under both the MIT license found in the
+# LICENSE-MIT file in the root directory of this source tree and the Apache
+# License, Version 2.0 found in the LICENSE-APACHE file in the root directory
+# of this source tree.
+
+load("@prelude//:paths.bzl", "paths")
+load("@prelude//utils:set.bzl", "set")
+
+# Create a unified compilation database for all targets.  This is useful when trying to integrate
+# buck2 with traditional LSPs like clangd, which expect not a compilation database per target, but
+# rather a single global compilation database.  The user can optionally specify a target pattern
+# with --targets=<pattern> to generate a commpilation database for a subset of targets.
+#
+# This works by using aquery to query the action graph on each target in the selected target set,
+# and then filtering out actions with a category other than cxx_compile.
+#
+# It is possible for multiple targets to contain the same file.  In this case, we choose to select
+# only the first file and throw away any subsequent compilation database entries referring to the
+# same file.  LSP servers do not handle this case in general anyway.
+#
+# Presently the bxl writes the stdout, and the user is expected to redirect this to a file.  A
+# possible future improvement could perhaps provide an option to write it as an artifact to the
+# output folder and only print the path to the output file to stdout, so that it could benefit from
+# caching.
+#
+# The bxl is quite fast.  It has no dependencies on actually running any actions (unlike building
+# the [compilation-database] subtarget for a given target), and on a moderately sized project with
+# around 4000 source files, generation times on the order of 5 seconds or less can be expected.
+#
+# Sample Usage:
+#     buck2 bxl root//prelude/tools/bxl/compile_commands.bxl:gen_compilation_database > compile_commands.json
+#
+
+def create_compilation_database_entry(ctx, target, buildfile_folder, action):
+
+    # the "cmd" attribute looks like a list.  For example, it might be something like:
+    #   cmd = [cl.exe, -c, foo.cpp]
+    # but in reality this is quite literally the string "[cl.exe, -c, foo.cpp]".  We need
+    # it as an actual list, so we have to employ this unfortunate hack of stripping the brackets,
+    # splitting on comma operator, and then stripping spaces from each entry to make an actual
+    # list.  Hopefully this doesn't fail in any weird edge cases, but it seems to be the best we
+    # can do until buck2 upstream actually stores a Starlark list as the value of this attribute.
+    cmd = action.attrs.cmd.value()
+    cmd = cmd.strip("[]").split(',')
+    cmd = list(map(lambda x: x.strip(), cmd))
+
+    # Get the path of the source file relative to the project root by appending the identifier
+    # to the directory of the BUCK file.
+    project_rel_path = paths.join(buildfile_folder, action.attrs.identifier.value())
+
+    # Now build the compilation database entry
+    entry = {
+        "file": project_rel_path,
+        "directory": ctx.fs.abs_path_unsafe("root//"),
+        "arguments": cmd
+    }
+    return entry
+
+def gen_compilation_database_impl(ctx):
+    # Generate compilation database for all targets unless user requests a more narrow set
+    target_filter = ctx.cli_args.targets or '...'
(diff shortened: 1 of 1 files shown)


zjturner:
@Overhatted

zjturner:
> A fix for the CI issue landed on master, you should pick it up if you rebase.
> 
> I have to say I don't quite understand the need for aquery here. The obvious thing to me would have been to run analysis on the whole world to gather all the compilation DBs, and then just merge them. How does that approach go wrong?

When I tried that (maybe I did it wrong) it was extremely slow.  It was an early attempt, so my memory is rusty, but I recall that generation of all the compilation dubs was very slow, because the sub target to generate the compilation db in the first place had dependencies on other targets (didn’t investigate what) and it would build things before it was able to generate the compilation db.  I don’t know what actions that “build things” step ran (didn’t investigate), I just remember that it was slow.  This method generates a universe compilation db very fast.

JakobDegen:
> because the sub target to generate the compilation db in the first place had dependencies on other targets

The only action that's run by compilation DB generation does argsfile expansion. I could maybe imagine this forcing targets in the `srcs` attribute or something of that sort to get built, but we should fix that if so. But if you don't need argsfile expansion, which seems to be the case here, then I think just having adding a `compilation-database-unexpanded` subtarget, and combining that across the entire repo, would be the right thing to do

ndmitchell:
So the context here is that when you generate a compilation database, you basically have two choices:

1. Generate the database with zero dependencies. That basically ends up looking like the aquery solution.
2. Generate the database with lots of dependencies - this ensures that things like generated code are produced properly. It also means that you can literally run the command lines produced and they work.

I believe our `[compilation-database]` target goes for lots of dependencies. I originally worked on a version with zero dependencies, but then our IDE teams wanted them added to support their use cases.

@JakobDegen is suggesting we have `[compilation-database-unexpanded]`. We probably should, but I don't think this solution is harmful, so suggest we merge it anyway?

prelude/cxx/tools/compile_commands.bxl
```diff
@@ -0,0 +1,97 @@
...
+    # the "cmd" attribute looks like a list.  For example, it might be something like:
+    #   cmd = [cl.exe, -c, foo.cpp]
+    # but in reality this is quite literally the string "[cl.exe, -c, foo.cpp]".  We need
+    # it as an actual list, so we have to employ this unfortunate hack of stripping the brackets,
+    # splitting on comma operator, and then stripping spaces from each entry to make an actual
+    # list.  Hopefully this doesn't fail in any weird edge cases, but it seems to be the best we
+    # can do until buck2 upstream actually stores a Starlark list as the value of this attribute.
+    cmd = action.attrs.cmd.value()
+    cmd = cmd.strip("[]").split(',')
+    cmd = list(map(lambda x: x.strip(), cmd))
```
ndmitchell, line 47: Can we get this chunk of code pulled out into a separate function? We probably will want to do something in future to clean this up, and having it separate makes it easier to see what to do.

CC @wendy728 for how we might want to unpack cmd_args values in future.

prelude/cxx/tools/compile_commands.bxl
```diff
@@ -0,0 +1,97 @@
...
+        buildfile_path = ctx.fs.project_rel_path(target.buildfile_path)
+        buildfile_folder = paths.dirname(buildfile_path)
+
+        # Now walk each compilation action and generate a compilation database entry for it.
+        for action in compile_actions:
+            entry = _create_compilation_database_entry(ctx, buildfile_folder, action)
+            filename = entry["file"]
+            if not files.contains(filename):
+                entries.append(entry)
+            files.add(filename)
```
ndmitchell, line 88: Indent this line one level?

cbarrete:
@zjturner Did you plan on finishing this PR and getting it merged? If not, would you be ok with someone picking it up? It seems that @ndmitchell only had minor comments about this version and was otherwise ready to merge it.

zjturner:
Sorry for the lack of response, I've been gone for some time.  Please feel free to take it over.

dav1d-wright:
I have a question about this bxl script, and would be happy to help out if possible. Bear in mind I'm only just getting started with buck2 and have no buck1 experience yet. 

The issues I am currently facing is with how the `targets` are passed to this script, namely:
1. I somehow can't manage to pass multiple targets to the script
  a) e.g. I'd like for the compdb to be built for `//:main` and `//:test`, but I can't figure out the command-line syntax for it
2. If I try to build the compdb for all targets, `ctx.configured_targets(target_filter)` fails for my demo project that uses thirdparty dependencies that have BUCK files. Details on this below.

I put together a very simple demo project, with which I am trying to get acquainted with the build system. I have integrated `googletest` and `range-v3` at the moment, as a way to understand how to integrate thirdparty dependencies. I am considering git submodules or vendoring, where I will add BUCK files for projects that don't have them, or modify them for projects that have them but some issues arise (e.g. not supporting buck2). 

The problem that I have stumbled across is with `range-v3`. This library does have `BUCK` files, and provides them also for examples contained in the repository. If I run the bxl script, I get the following error:

```
buck2 bxl //compile_commands.bxl:gen_compilation_database --

Traceback (most recent call last):
  File <builtin>, in <module>
  * compile_commands.bxl:68, in _gen_compilation_database_impl
      targets = ctx.configured_targets(target_filter)
error: Error in configured node dependency, dependency chain follows (-> indicates depends on, ^ indicates same configuration as previous):
       root//thirdparty/range-v3/example:accumulate_ints (prelude//platforms:default#904931f735703749)
    -> root//:range-v3 (^)

  --> compile_commands.bxl:68:15
   |
68 |     targets = ctx.configured_targets(target_filter)
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |


Caused by:
    0: looking up unconfigured target node `root//:range-v3`
    1: Unknown target `range-v3` from package `root//`.
       Did you mean one of the 2 targets in root//:BUCK?
```

What I found is that these examples refer to parts of the library, expecting them to be in the root folder, which is not the case for my setup (and presumably not for most). My project is not building these example project, but it seems like they are still picked up by  `configured_targets` for some reason, and failing. 


Any help on these two issues would be greatly appreciated. I'd be happy to contribute the solution back, if something can be :slightly_smiling_face:

cjhopman:
If we merge https://github.com/facebook/buck2/pull/810, do we still want this version?

cbarrete:
@cjhopman I'm obviously biased, but I don't think that the version in this PR is very useful, because it is fast at the cost of correctness: dependencies are not materialized, code generation is not run, etc. This completely breaks tooling like clang-tidy in all but the most trivial cases, and more permissive tools like clangd are useable, but not with flawless integration.

My version is slower but more correct. When running on a mid-sized monorepo (a few millions of lines of C++, but not dozens), what actually takes time is building our code generator from source. The actual generation of the compilation database is very quick. Note that my version is not perfect because it builds the compilation database for a single platform, and BXL doesn't seem to "follow" transitions, but the result remains very usable, even in a heavily (split) transitioned codebase.

alexlian:
@zjturner Following up if #810 covers your needs, or is there more to follow up on here?

PR #517, opened 2023-12-05
Added example using Buck with Visual Studio Code on Windows

Overhatted:
1. I made the default platform in the prelude public so it can be inherited from. I don't know why it wasn't already like that. Maybe I'm misunderstanding how to use it
2. Not sure if I'm using the config_setting and constraint_setting properly since it seems duplicated.
3. The script to generate the compile_commands.json was more or less copied from #510.

There are still a lot of improvements to be made but I think this is good enough to be merged:
1. With MSVC the compilation should probably be done with /Z7 by default and then, if the "stripped" sub-target is asked for, that flag is removed.
2. The cxx toolchain needs the compilation options looked at, not sure if Linux's defaults work properly.
3. The Install.ps1 and Copy.ps1 need to be ported to Linux.

### DIFF
--- a/prelude/toolchains/cxx.bzl
+++ b/prelude/toolchains/cxx.bzl
@@ -67,7 +67,6 @@ def _system_cxx_toolchain_impl(ctx: AnalysisContext):
         shared_library_name_default_prefix = ""
         shared_library_name_format = "{}.dll"
         shared_library_versioned_name_format = "{}.dll"
-        additional_linker_flags = ["msvcrt.lib"]
         pic_behavior = PicBehavior("not_supported")
     elif ctx.attrs.linker == "g++" or ctx.attrs.cxx_compiler == "g++":
         pass
--- a/examples/visual_studio/.buckconfig
+++ b/examples/visual_studio/.buckconfig
@@ -1,0 +1,17 @@
+[repositories]
+root = .
+prelude = prelude
+toolchains = toolchains
+none = none
+
+[repository_aliases]
+config = prelude
+fbcode = none
+fbsource = none
+buck = none
+
+[parser]
+target_platform_detector_spec = target:root//...->toolchains//:default
+
+[build]
+execution_platforms = prelude//platforms:default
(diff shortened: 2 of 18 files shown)


ndmitchell:
Thanks! I think our bar for samples is relatively low (if this is what you want to do, its super helpful to get started) so would be happy to get this included.

Overhatted:
Now I think it's ready for merging. Let me know why the "Meta Import Checks" is failing and what I can do to fix it.

Overhatted:
Hello @ndmitchell, can you let me know why the "Meta Import Checks" fails? Is it the .vscode files?

JakobDegen:
@Overhatted yes, that is why. I'll get that fixed, in the meantime I'm trying to see if I can get someone on the windows side internally to review this, otherwise Neil is right that the bar for examples is fairly low and we can just get this merged

prelude/toolchains/cxx.bzl
```diff
@@ -67,7 +67,6 @@ def _system_cxx_toolchain_impl(ctx: AnalysisContext):
         shared_library_name_default_prefix = ""
         shared_library_name_format = "{}.dll"
         shared_library_versioned_name_format = "{}.dll"
-        additional_linker_flags = ["msvcrt.lib"]
```
KapJI, line 70: This one was added to make hello_world example to work on Windows. Do you know how does it build now?

KapJI (approved):
Looks good to me, thank you!

prelude/toolchains/cxx.bzl
Overhatted: I can't think of any good reason. I tried to reproduce a failure like that with reasonable flags but couldn't get that error. Like it says in the documentation https://learn.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170 the compiler places a linking directive in the object file specifying the runtime library to use (by default MT_StaticRelease on my machine).

If you pass in this flag https://learn.microsoft.com/en-us/cpp/build/reference/zl-omit-default-library-name?view=msvc-170 you do get that error but I did a CTRL+F through the repository and didn't find that flag anywhere.

Issue #520, opened 2023-12-13
genrule with output folder consumed by cxx_library


zjturner:
I want to have a genrule that outputs multiple cpp files to a directory, and then have those files consumed and all compiled by a cxx_library.  I tried this:

```
genrule(
    name="foo-gen",
    srcs=["foo.json", "generate.py"],
    out = "foo-folder",   # This is the folder where all the output files go
    cmd = "python generate.py foo.json -o ${out}/foo"
)

cxx_library(
    name = "foo",
    srcs = glob(["$(output :foo-gen)/foo/*.cpp)"])
```

This doesn't work, I guess because you can't expand string parameter macros inside of a glob.  The problem appears to be with the cxx rule, and not with the genrule, as the genrule can certainly create the directory and output multiple files there.

Is there anything that can be done to handle this use case (generate multiple files from one input, consume them from a dependent rule)?

thoughtpolice:
This seems to be basically the equivalent of "how do I do `glob` on an `http_archive`", which I somewhat addressed here: https://github.com/facebook/buck2/discussions/387#discussioncomment-7136698

I think you would need to write a custom rule that called `cxx_library_impl` and used dynamic dependencies to discover the list of files from the source directory. I'd been meaning to explore this kind of idea but didn't get too far before...

zjturner:
If that’s the case, couldn’t `cxx_library` be taught to do it natively?

cormacrelf:
Any attrs.source() can accept either a file or something with a DefaultInfo. DefaultInfo has a default_outputs field, and that's a plural; it's a list. So I wonder if you can just provide any old dependency with multiple default_outputs. Including the output of a dynamic dependency-using glob rule.

cormacrelf:
Alas, apparently not. This gives "Expected a single artifact from root//scratch:id (platforms//:host#9f5b0576f0e10a13), but it returned 2 artifacts"

```starlark
# BUCK
load(":defs.bzl", "verbatim")

verbatim(
    name = "id",
    srcs = [ "one.cpp", "two.cpp" ],
)

cxx_library(
    name = "lib",
    srcs = [":id"]
)

# defs.bzl
verbatim = rule(
    impl = lambda ctx: [ DefaultInfo(default_outputs = ctx.attrs.srcs) ],
    attrs = { "srcs": attrs.list(attrs.source()) },
)
```

zjturner:
[...]

PR #522, opened 2023-12-14
Fix attrs.list(attrs.one_of(attrs.source(), xxx)))

cormacrelf:
In #520, I was wondering why you couldn't pass a label with multiple `default_outputs` to `cxx_library`. This is why:

https://github.com/facebook/buck2/blob/c7637e656536d9e5a4cdeb2f6c801e71c40c0df6/prelude/decls/cxx_common.bzl#L17

The srcs attr was exercising this code here, and hitting the error "Expected a single artifact".

Note that this will not help with `attrs.tuple(attrs.source(), xxx)`. That still has weird behaviour. For the [example](https://github.com/facebook/buck2/issues/520#issuecomment-1855376386) I posted, if you use `srcs = [(":id", ["-Wall"])]` instead, you get this:

> Value `<source scratch/two.cpp>` of type `artifact` does not match the type annotation `list[resolved_macro]` for argument `flags`

You would have to add some third variant to the one_of with `attrs.tuple(attrs.list(attrs.source), ...)` to handle that case.

### DIFF
--- a/app/buck2_analysis/src/attrs/resolve/configured_attr.rs
+++ b/app/buck2_analysis/src/attrs/resolve/configured_attr.rs
@@ -110,7 +110,7 @@ impl ConfiguredAttrExt for ConfiguredAttr {
             ConfiguredAttr::Tuple(list) => {
                 let mut values = Vec::with_capacity(list.len());
                 for v in list.iter() {
-                    values.append(&mut v.resolve(pkg.dupe(), ctx)?);
+                    values.push(v.resolve_single(pkg.dupe(), ctx)?);
                 }
                 Ok(ctx.heap().alloc(AllocTuple(values)))
             }
--- a/app/buck2_analysis/src/attrs/resolve/configured_attr.rs
+++ b/app/buck2_analysis/src/attrs/resolve/configured_attr.rs
@@ -81,6 +81,8 @@ impl ConfiguredAttrExt for ConfiguredAttr {
         match self {
             // SourceLabel is special since it is the only type that can be expand to many
             ConfiguredAttr::SourceLabel(src) => SourceAttrType::resolve_label(ctx, src),
+            // OneOf could contain a SourceLabel
+            ConfiguredAttr::OneOf(box l, _) => l.resolve(pkg, ctx),
             _ => Ok(vec![self.resolve_single(pkg, ctx)?]),
         }
     }


cormacrelf:
Whoops, too much in this PR.

cormacrelf:
I guess I can fix the tuple weirdness too. Tuples should run resolve_single on their child attrs, otherwise you end up with the wrong number of things in the tuple and a misalignment of types.

cormacrelf:
I have a second diff that adds context to the errors, if that's appropriate. I found it a huge improvement, but it might make normal errors a bit verbose? Let me know if you want this.

```
For:
    somerule(
        gumbo = {
            "files": ":ls",
        }
    )
Before:
    Expected a single artifact from root//scratch:ls (platforms//:host#9f5b0576f0e10a13), but it returned 2 artifacts
After:
    0: Resolving attribute `gumbo`
    1: In attrs.dict() value for key "files"
    2: Resolving attrs.source() with label root//scratch:ls (platforms//:host#9f5b0576f0e10a13)
    3: Expected a single artifact from root//scratch:ls (platforms//:host#9f5b0576f0e10a13), but it returned 2 artifacts

For:
    cxx_library(
        name = "lib",
        srcs = [ ("one.cpp", ["$(hello)"]) ]
    )
Before:
    0: Error resolving `$(hello)`.
    1: There was no mapping for hello.
After:
    0: Resolving attribute `srcs`
    1: In attrs.list() [0]
    2: In attrs.tuple() [1]
    3: In attrs.list() [0]
    4: In attrs.arg()
    5: Error resolving `$(hello)`.
    6: There was no mapping for hello.
```

cormacrelf:
These errors use `anyhow::Context`, which is blessed by being very simple, but I think we could count up recursion depth as the stack unwinds through the ?s, and add context conditionally. Certainly possible.

Does buck2_error have anything useful for that? I see it in commits all the time but I haven't used it. It seems to have a Context trait like anyhow.

cjhopman:
I think it'd be good for us to figure out with buck2_error how to propagate up information necessary for producing good messages without needing that information itself to be a line in the error message.

Like, we should be able to make that message be:

```
After:
    0: Failed to coerce attribute `srcs`.
           value `$(hello)` cannot be coerced to attr.arg() (in `srcs[0][1][0]`)
           because there is no mapping for template `hello` (available user-defined mappings are `foo`, `bar`) 
           See https://buck2.build/docs/api/build/globals/#templateplaceholderinfo for information about string templates
```

the main difference there just that 1-5 are all collapsed into the second line

PR #536, opened 2024-01-16
Fix Haskell linker error on MacOS

### LINKED ISSUE #535: haskell_library fails on MacOS
Trying to build a `haskell_library` target with Buck2 on MacOS fails with a linker error of the form
```
ld: unknown option: -soname
```

The reason seems to be that platform specific properties are currently hard-coded to Linux values in the Haskell rules, e.g. the [`-soname` flag](https://github.com/facebook/buck2/blob/20c366bd08eeb0c2143bc2942bfbfece13dd591b/prelude/haskell/haskell.bzl#L527). For reference, the cxx toolchain [distinguishes such flags by linker type](https://github.com/facebook/buck2/blob/main/prelude/cxx/linker.bzl).

aherrmann:
Closes #535

- Haskell toolchain: automatic host arch name
- Configure platform specific linker flags

### DIFF
--- a/prelude/haskell/haskell.bzl
+++ b/prelude/haskell/haskell.bzl
@@ -21,7 +21,6 @@ load(
 load(
     "@prelude//cxx:linker.bzl",
     "LINKERS",
-    "get_shared_library_name_linker_flags",
     "get_shared_library_flags",
 )
 load(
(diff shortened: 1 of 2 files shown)


ndmitchell:
Thanks @aherrmann! Looks great. Might see if we can turn on our GitHub Mac/Haskell tests once this lands :)

ndmitchell:
Added to CI in f8c45694a9e1acc463fd0c988133e1201e923d2f and it works great. Thanks once again.

PR #565, opened 2024-02-08
Integrate with dotslash

bigfootjon:
As discussed internally, we want to use dotslash as a distribution mechanism for bucks. This PR stands up the initial integration.

In order to test I had to fix some things in the GitHub Actions workflows, so I included those here.

Ref: https://dotslash-cli.com/
Ref: https://github.com/facebook/dotslash-publish-release

Tasks: T175738402

Test plan:

See this action on my fork: https://github.com/bigfootjon/buck2/actions/runs/7836463888/job/21385144587

Which created this release: https://github.com/bigfootjon/buck2/releases/tag/2024-02-08

It contains a correct dotslash file: https://gist.github.com/bigfootjon/39c63a3a6772f66fc13c84c9116a56d7

(this version predates a fix I made to use "buck2" as the filename instead of hermes, oops)

When run using the opensource version of dotslash it correctly downloads and runs the right binary!

### DIFF
--- a/.github/dotslash-config.json
+++ b/.github/dotslash-config.json
@@ -1,0 +1,29 @@
+{
+    "exclude-github-release-provider": true,
+    "outputs": {
+      "hermes": {
+        "platforms": {
+          "macos-aarch64": {
+            "regex": "^buck2-aarch64-apple-darwin.zst$",
+            "path": "buck2-aarch64-apple-darwin"
+          },
+          "linux-aarch64": {
+            "regex": "^buck2-aarch64-unknown-linux-gnu.zst$",
+            "path": "buck2-aarch64-unknown-linux-gnu"
+          },
+          "macos-x86_64": {
+            "regex": "^buck2-x86_64-apple-darwin.zst$",
+            "path": "buck2-x86_64-apple-darwin"
+          },
+          "windows-x86_64": {
+            "regex": "^buck2-x86_64-pc-windows-msvc.exe.zst$",
+            "path": "buck2-x86_64-pc-windows-msvc.exe"
+          },
+          "linux-x86_64": {
+            "regex": "^buck2-x86_64-unknown-linux-gnu.zst$",
+            "path": "buck2-x86_64-unknown-linux-gnu"
+          }
+        }
+      }
+    }
+  }
--- a/.github/dotslash-config.json
+++ b/.github/dotslash-config.json
@@ -1,7 +1,7 @@
 {
     "exclude-github-release-provider": true,
     "outputs": {
-      "hermes": {
+      "buck2": {
         "platforms": {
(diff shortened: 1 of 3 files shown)


bolinfest:
I'm very excited to see this!

Since you may be the first "real" user of this action, I'm certainly interested in feedback on the API.

One thing I didn't have a chance to build in is the ability to add metadata to the generated DotSlash file along the lines of the code-compose-lsp example in the blog post:

https://engineering.fb.com/2024/02/06/developer-tools/dotslash-simplified-executable-deployment/

This buck2 use case might be a good case to explore this space!

.github/dotslash-config.json
```diff
@@ -0,0 +1,29 @@
+{
+    "exclude-github-release-provider": true,
+    "outputs": {
+      "hermes": {
```
bolinfest, line 4: eh, hermes?

bigfootjon: Bad copypasta, fixed!

bigfootjon:
> I'm very excited to see this!
> 
> Since you may be the first "real" user of this action, I'm certainly interested in feedback on the API.

So far it looks great! It was super easy to setup, I did have some slight confusion over what to set as the "path" for a single file archive, but the dotslash-cli.com docs covered this case so I wasn't blocked.

> One thing I didn't have a chance to build in is the ability to add metadata to the generated DotSlash file along the lines of the code-compose-lsp example in the blog post:
> 
> https://engineering.fb.com/2024/02/06/developer-tools/dotslash-simplified-executable-deployment/
> 
> This buck2 use case might be a good case to explore this space!

Yeah metadata like that could be useful, especially provenance info. For a V1 this seems to work perfectly fine. Are you going to open a GitHub Issue in the action repo about this feature?

bolinfest:
> Yeah metadata like that could be useful, especially provenance info. For a V1 this seems to work perfectly fine. Are you going to open a GitHub Issue in the action repo about this feature?

If you were willing to open one with a concrete use case (rather than a theoretical one that I make up), I think that would be a more compelling starting point, if that's OK.

bolinfest:
Also, the `name` in the DotSlash file is `buck2`, right? I wonder if we should embed the version in the name or perhaps formally make that a separate field in the JSON.

An advantage of a separate field is that it might be helpful with DotSlash garbage collection later if we could identify "you have N different versions of the same DotSlash file"...

/cc @zertosh @dtolnay

bigfootjon:
I think adding a standard header to the generated file would serve this purpose, especially if it included the tag/release it came from (what I proposed in the above issue)

bolinfest:
Yes, but I also want to have “best practices” for people who generate their
own DotSlash files through other means so we can exploit this consistency
for things like garbage collection.

PR #570, opened 2024-02-19
dotslash: use `-linux-musl` builds instead of `linux-gnu`

thoughtpolice:
Prebuilt `linux-gnu` buck2 binaries will not work on systems with non- traditional FHS layouts and `ld.so` locations, like NixOS, Guix, or your favorite 10 year old distro release with an old glibc. Use `linux-musl` builds instead to fix this.

### DIFF
--- a/.github/dotslash-config.json
+++ b/.github/dotslash-config.json
@@ -8,8 +8,8 @@
             "path": "buck2-aarch64-apple-darwin"
           },
           "linux-aarch64": {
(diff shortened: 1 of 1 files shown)


thoughtpolice:
Frankly, this should be fixed in Dotslash as well (it should just provide statically linked linux binaries; currently it just seems to ship Ubuntu-built binaries), but I think that's a bit easier to solve on the users end since Dotslash works with stable rust and is therefore more easily supported in contexts like NixOS/Guix.

Issue #532, opened 2024-01-05
Remote execution can't find executable


Overhatted:
The remote execution with Buildbarn example doesn't seem to work. I'm running Buildbarn in Window's Docker Desktop and Buck2 in Ubuntu inside WSL2.
I get:
`Error was returned on the stream by RE: Failed to run command: Cannot find executable "sh" in search paths ""`
This comes from here: https://github.com/buildbarn/bb-remote-execution/blob/4b3a11b508cf06a3dfa17f141b3ea505bc039434/pkg/runner/local_runner_unix.go#L78
It seems like the "search paths" should be the PATH that Buck2 sends to the remote execution server but I can't find any place in Buck2 where I can set that. Where can I set it?

For the moment I'm using absolute paths in system_cxx_toolchain. Maybe I'm missing something since no one else is having this problem.

yaoddao:
Hi Overhatted, I met the same question https://github.com/facebook/buck2/issues/549. I think the reason of the problem may be the cooperation between Buck2 and Buildbarn has a problem.

yaoddao:
I think：
- When using relative paths, eg. 'sh', Buck2 sends 'sh' and path '' to Buildbarn, Buildbarn can not search 'sh' in path ''.
- When using absolute paths, eg. '/usr/bin/sh', Buck2 sends '/usr/bin/sh' and path '' to Buildbarn, Buildbarn directly use '/usr/bin/sh' to  run.
- But Bazel does not have this problem, so I suspect Bazel not only sent 'sh' but also additional path information to Buildbarn, when using relative paths 'sh'. This additional information can let Buildbarn to find 'sh' in PATH environment.

yaoddao:
I think the reason of the problem must be a cooperation problem between Buck2 and Buildbarn.
The cooperation between Buck2 and Buildbuddy does not have the same question.
When buck2 sends Linux command, eg. g++,sh, and so on, Buildbarn only searches the command in path "", rather than in PATH.
But when buck2 sends Linux command, Buildbuddy always searches the command in PATH.

Also, the cooperation between Bazel and Buildbarn also does not have the same question. I think that Bazel not only sends Linux the command but also additional information to Buildbarn. So Buildbarn can search for the command correctly

stagnation:
To confirm, does buck2 send an empty $PATH variable?
https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto#L591

[...]

PR #584, opened 2024-02-28
Document how to set the Buildbarn PATH

stagnation:
Following the discussion in #532 .

### DIFF
--- a/examples/remote_execution/buildbarn/README.md
+++ b/examples/remote_execution/buildbarn/README.md
@@ -1,14 +1,46 @@
-## Remote execution integration with Buildbarn
+# Remote execution integration with Buildbarn
 
 This project provides a small example of what a project that utilizes
 [Buildbarn](https://github.com/buildbarn).
 
 In this document, we will go over the key configs used in this setup. Using a
 local docker-compose deployment from the
-[example deployment repo](https://github.com/buildbarn/bb-deployments). If you
-already have a Buildbarn deployment you can skip that.
+[example deployment repo](https://github.com/buildbarn/bb-deployments).
+If you already have a `Buildbarn` deployment you can use that instead.
 
-### Deploy a local Buildbarn
+## Deploy a local Buildbarn
+
+### Set up a basic PATH
+
+`Buildbarn` runs all actions in a completely bare environment by default, and
+relies on the build client to specify the `$PATH` and other environment variables
+in the REAPI action message. As `buck2` does not send a default `$PATH` build
+will fail with errors about missing executables.
+
+Add the environment variables to the worker config you want - normally either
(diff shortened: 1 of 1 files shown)


examples/remote_execution/buildbarn/README.md
```diff
@@ -1,14 +1,46 @@
...
+
+`Buildbarn` runs all actions in a completely bare environment by default, and
+relies on the build client to specify the `$PATH` and other environment variables
+in the REAPI action message. As `buck2` does not send a default `$PATH` build
+will fail with errors about missing executables.
+
+Add the environment variables to the worker config you want
+`docker-compose/config/worker-fuse-ubuntu22-04.jsonnet` or
+`docker-compose/config/worker-hardlinking-ubuntu22-04.jsonnet`
+`.buildDirectories[].runners[]`, after the `workerId:` dictionary.
```
JakobDegen, line 23: I'm having some trouble parsing this sentence, did a part of it get cut out?

stagnation: Good point, thanks! I will revise it.

What I try to say (concisely) is: update whichever of the two paths you use, as you can choose either hardlinking or FUSE. Which for the sake of the example is problably bad. Hardlinking is the simplest but a little slower, we could clarify the example by dropping FUSE. Next, to update the proto file, which is a simple insertion but in a big structure, so here is a `jq`-style path for where to add it. Unfortunately this cannot be done with `yq` or `jq`.
And then, there is the big diff listing because it is still annoying to describe what to do :S.

I hope this is better:
```
Add the environment variables to the worker configs: `worker-fuse-ubuntu22-04.jsonnet` and `worker-hardlinking-ubuntu22-04.jsonnet`. The proto configuration specifies a `buildDirectories` list, with a `runner` list inside. There is only one runner in each config file. That runner should be modified to have the environment variables values:

      environment_variables: {
         PATH: '/bin:/usr/bin:/usr/local/bin',
      },

They can be put anywhere in the `runner`. The example diff below adds them at the end, after `workerId`.

It is also possible to make a crude insertion with `sed`:

     sed -i '/workerId/i \      environment_variables: { PATH: \'/bin:/usr/bin:/usr/local/bin\'},' docker-compose/config/worker-*-ubuntu22-04.jsonnet

```

JakobDegen: I think your original explanation was fine in principle, I just didn't quite understand the grammatical structure. Something like this would clear it up for me if it's accurate

> Add the environment variables to the worker config you want - normally either `path1` or `path2` - after the `workerId` field like this:

But the more detailed one is just as good, completely up to you

JakobDegen:
Thanks!

PR #585, opened 2024-02-29
Fix the issues link

NoClip:
The old link reference to relative URL (https://github.com/facebook/buck2/blob/main/issues) which results to a broken URL (404 error).


JakobDegen:
Thanks!

PR #581, opened 2024-02-26
Adding rule to build Windows .rc files

Overhatted:
Adds the ability to compile resource files: https://learn.microsoft.com/en-us/windows/win32/menurc/about-resource-files

More details are in the commit messages.

PR #591, opened 2024-03-05
[vswhere.py] Add support for EWDK to the prelude / msvc toolchain

jrmolin:
Microsoft releases an Enterprise WDK (EWDK) bundle for building drivers and applications without needing a Visual Studio installation. This is very useful for CI builds, where unpacking a zip or mounting an iso file is all that is needed to spin up a build VM, instead of needing to download, install, validate, install WDK, install SDK, repair WDK, etc. Also, there is no license requirement to use the EWDK.

This updates the `vswhere.py` file to extract the necessary binaries, include dirs, and library dirs from the environment produced by calling the `SetupBuildEnv.bat` utility in the EWDK. Once this is done, the rest of a buck2 build operates normally.

To use this, download and "install" the EWDK to the filesystem. 

```
> set EWDKDIR=C:\ewdk11
> buck2 build //path/to/target
```


prelude/toolchains/msvc/vswhere.py
```diff
@@ -202,6 +202,76 @@ def write_tool_json(out, tool):
...
+    LIB = [lib_path]
+    INCLUDE = [include_path]
+
+    ucrt = Path(env.get("UCRTContentRoot"))
+    ucrt_version = env.get("Version_Number")
+
+    if ucrt and ucrt_version:
+        PATH.append(ucrt / "bin" / ucrt_version / "x64")
+        LIB.append(ucrt / "lib" / ucrt_version / "ucrt" / "x64")
+        INCLUDE.append(ucrt / "include" / ucrt_version / "ucrt")
```
KapJI, line 257: `ucrt` is always set at this place.
```suggestion
    ucrt = Path(env["UCRTContentRoot"])
    ucrt_version = env.get("Version_Number")

    if ucrt_version:
        PATH.append(ucrt / "bin" / ucrt_version / "x64")
        LIB.append(ucrt / "lib" / ucrt_version / "ucrt" / "x64")
        INCLUDE.append(ucrt / "include" / ucrt_version / "ucrt")
```

prelude/toolchains/msvc/vswhere.py
```diff
@@ -202,6 +202,76 @@ def write_tool_json(out, tool):
...
+
+    sdk = Path(env.get("WindowsSdkDir"))
+    sdk_version = ucrt_version
+    if sdk and sdk_version:
```
KapJI, line 261:
```suggestion
    sdk = Path(env["WindowsSdkDir"])
    sdk_version = ucrt_version
    if sdk_version:
```

prelude/toolchains/msvc/vswhere.py
```diff
@@ -202,6 +202,76 @@ def write_tool_json(out, tool):
...
+        PATH.append(sdk / "bin" / "x64")
+        LIB.append(sdk / "lib" / sdk_version / "um" / "x64")
+        INCLUDE.append(sdk / "include" / sdk_version / "um")
+        INCLUDE.append(sdk / "include" / sdk_version / "cppwinrt")
+        INCLUDE.append(sdk / "include" / sdk_version / "winrt")
+        INCLUDE.append(sdk / "include" / sdk_version / "shared")
+
+    exe_names = "cl.exe", "lib.exe", "ml64.exe", "link.exe"
```
KapJI, line 269: Let's move this to global const to avoid duplicating.

prelude/toolchains/msvc/vswhere.py
```diff
@@ -216,6 +286,8 @@ def main():
         lib_exe = find_in_path("lib.exe")
         ml64_exe = find_in_path("ml64.exe")
         link_exe = find_in_path("link.exe")
```
KapJI, line 288: With const list this can be simplified.
```suggestion
        cl_exe, lib_exe, ml64_exe, link_exe = (find_in_path(exe) for exe in EXE_NAMES)
```

prelude/toolchains/msvc/vswhere.py
```diff
@@ -202,6 +202,76 @@ def write_tool_json(out, tool):
     out.write(j)
 
 
+# for use with the ewdk to grab the environment strings
+def get_env(ewdkdir: str):
```
KapJI, line 206:
```suggestion
def get_ewdk_env(ewdk_dir: Path):
```

prelude/toolchains/msvc/vswhere.py
```diff
@@ -202,6 +202,76 @@ def write_tool_json(out, tool):
...
+
+    # We need to write the script that will make the important variables available
+    # This gets run from the temporary build directory, so this does not pollute the sources
+    vcVarsExtract = os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                "VcVarsExtract.bat")
+    if not os.path.exists(vcVarsExtract):
+        with open(vcVarsExtract, "w") as f:
+            f.write("@echo off\r\ncall %%1 %%2 %%3 %%4 %%5 %%6 %%7 %%8 %%9 > NUL\r\nset")
+
+    batch_file = os.path.join(ewdkdir, "BuildEnv", "SetupBuildEnv.cmd")
```
KapJI, line 221:
```suggestion
    env_script = ewdk_dir / "BuildEnv" / "SetupBuildEnv.cmd"
```

prelude/toolchains/msvc/vswhere.py
```diff
@@ -202,6 +202,76 @@ def write_tool_json(out, tool):
...
+    http://stackoverflow.com/questions/3503719/#comment28061110_3505826 (ahal)
+    """
+
+    # We need to write the script that will make the important variables available
+    # This gets run from the temporary build directory, so this does not pollute the sources
+    vcVarsExtract = os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                "VcVarsExtract.bat")
+    if not os.path.exists(vcVarsExtract):
+        with open(vcVarsExtract, "w") as f:
+            f.write("@echo off\r\ncall %%1 %%2 %%3 %%4 %%5 %%6 %%7 %%8 %%9 > NUL\r\nset")
```
KapJI, line 219: You shouldn't create files next to the target inputs. Buck should set temporary directory for you.
```suggestion
    with tempfile.NamedTemporaryFile(prefix="VcVarsExtract", suffix=".bat") as tmp:
        print("@echo off", file=tmp)
        print("call %* > NUL", file=tmp)
        print("env", file=tmp)
```
```

File: prelude/toolchains/msvc/vswhere.py [hunk 8]
```diff
@@ -202,6 +202,76 @@ def write_tool_json(out, tool):
...
+
+    batch_file = os.path.join(ewdkdir, "BuildEnv", "SetupBuildEnv.cmd")
+
+    # This does not work as a list of strings, unfortunately
+    cmd = vcVarsExtract + ' "' + batch_file + '" ' + "amd64"
+
+    output = subprocess.check_output(cmd)
+    output = output.decode("utf-8")
```
KapJI, line 227: I don't get why list of arguments doesn't work. I think this should work.
```suggestion
    output = subprocess.check_output([tmp.name, env_script, "amd64"]).decode("utf-8")
```

KapJI (commented):
I like the idea, this should be useful.

But parsing of env vars and error handling needs to be improved.

prelude/toolchains/msvc/vswhere.py
jrmolin: Most of this is fine, but the final `set` can't be changed to `env`, because the `\` are imperfectly escaped. I also had to `delete=False`

prelude/toolchains/msvc/vswhere.py
jrmolin: For this, I mostly copied the structure of the code that used `vswhere` to find components. I can change it to what you want.

prelude/toolchains/msvc/vswhere.py
KapJI: Sure, let's keep `set` then.

prelude/toolchains/msvc/vswhere.py
```diff
@@ -202,6 +202,76 @@ def write_tool_json(out, tool):
...
+            first,second = line.split("=", 1)
+            env[first] = second
+
+    return env
+
+
+def find_with_ewdk():
+    env = get_env(os.environ.get("EWDKDIR"))
+
+    installation_path = Path(env.get("VSINSTALLDIR"))
```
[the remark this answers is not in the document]
jrmolin: Do you want these in a try/except block?

prelude/toolchains/msvc/vswhere.py
```diff
@@ -202,6 +203,74 @@ def write_tool_json(out, tool):
...
+    http://pythonwise.blogspot.fr/2010/04/sourcing-shell-script.html (Miki Tebeka)
+    http://stackoverflow.com/questions/3503719/#comment28061110_3505826 (ahal)
+    """
+
+    output = None
+
+    # We need to write the script that will make the important variables available
+    with tempfile.NamedTemporaryFile(prefix="VcVarsExtract", suffix=".bat", mode="w", delete=False) as tmp:
+        print(f"temp file at {tmp.name}")
+        output = tmp.name
```
KapJI, line 219: It's either unused or need to be replaced in `cmd`. In latter case rename it to `wrapper_script` to avoid clash with `output`  from `check_output`.
```suggestion
```

prelude/toolchains/msvc/vswhere.py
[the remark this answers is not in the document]
KapJI: If it crashes with reasonable error, it should be fine. In this case that will be `KeyError: "VSINSTALLDIR"`.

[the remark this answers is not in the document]
jrmolin: Ah, I see your point. Thanks for the suggestions!

prelude/toolchains/msvc/vswhere.py
jrmolin: It was for debugging when it was getting cleaned up, and I missed it. Deleting now.

jrmolin:
if https://github.com/facebook/buck2/pull/581 gets merged, I'll need to make some edits.

PR #597, opened 2024-03-08
Fix repositories link in key concepts doc

John-Colvin:
I'm assuming the backticks were accidental

manually verified it works in the markdown, I assume that will work for the rendered site as well.


rdrey:
Thanks for the fix!

PR #607, opened 2024-03-17
Added support for headers in the windows_resource rule

Overhatted:
Like other cxx rules it supports specifying "headers" or "include_directories" and "raw_headers".

As a test, I added a header dependency to the test I provided in the previous PR: #581.

The final result being the following files:
BUCK:
```
# A rule that includes a single .rc file and compiles it into an object file.
windows_resource(
    name = "resources",
    srcs = [
        "resources.rc",
    ],
    header_namespace = "",
    headers = ["message.h"],
)

# A rule that links against the above windows_resource rule.
# A rule that includes a single .rc file and compiles it into an object file.
windows_resource(
    name = "resources",
    srcs = [
        "resources.rc",
    ],
    header_namespace = "",
    headers = ["message.h"],
)

# A rule that links against the above windows_resource rule.
cxx_binary(
    name = "app",
    srcs = [
        "main.cpp",
    ],
    deps = [
        ":resources"
    ],
    linker_flags = [
        "User32.lib",
    ],
)
```
main.cpp:
```
#include <iostream>
#include <Windows.h>

int main(int argc, const char* argv[])
{
	std::string message;
	message.resize(50);
	int returnValue = LoadStringA(NULL, 4, message.data(), message.size());
	if (returnValue == 0)
	{
		std::cout << "Failed to read resource";
		return 1;
	}

	message.resize(returnValue);
	std::cout << message;
	return 0;
}
```
message.h:
```
#define MESSAGE_TO_PRINT "Hello from header"
```
resources.rc:
```
#include <winresrc.h>

#include "message.h"

STRINGTABLE
BEGIN
4 MESSAGE_TO_PRINT
END
```

### DIFF
--- a/prelude/cxx/windows_resource.bzl
+++ b/prelude/cxx/windows_resource.bzl
@@ -5,13 +5,34 @@
 # License, Version 2.0 found in the LICENSE-APACHE file in the root directory
 # of this source tree.
 
+load("@prelude//apple:xcode.bzl", "get_project_root_file")
 load("@prelude//cxx:cxx_context.bzl", "get_cxx_toolchain_info")
+load("@prelude//cxx:headers.bzl", "cxx_get_regular_cxx_headers_layout")
+load("@prelude//cxx:preprocessor.bzl", "cxx_merge_cpreprocessors", "cxx_private_preprocessor_info")
 load("@prelude//linking:link_groups.bzl", "LinkGroupLibInfo")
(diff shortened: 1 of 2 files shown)


Overhatted:
Hello @KapJI, you reviewed my last Windows PR #581, can you take a look at this one?

KapJI:
Looks good to me, thanks!

PR #617, opened 2024-04-11
Use transitive set for HaskellLibraryInfo

avdv:
This PR changes the fields of HaskellLibraryProvider to contain a transitive set of HaskellLibraryInfo instead of using a list.

- remove `merge_haskell_link_infos` since it is now unused
- apply change to ide script

### DIFF
--- a/prelude/haskell/haskell.bzl
+++ b/prelude/haskell/haskell.bzl
@@ -62,6 +62,7 @@ load(
     "@prelude//haskell:library_info.bzl",
     "HaskellLibraryInfo",
     "HaskellLibraryProvider",
+    "HaskellLibraryInfoTSet",
 )
 load(
     "@prelude//haskell:link_info.bzl",
@@ -69,7 +70,6 @@ load(
     "HaskellProfLinkInfo",
     "attr_link_style",
     "cxx_toolchain_link_style",
-    "merge_haskell_link_infos",
 )
 load(
     "@prelude//haskell:toolchain.bzl",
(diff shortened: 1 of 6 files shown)


ndmitchell:
Thanks for the changes! This looks great. I'm heading off on holiday until Tuesday, so have sent it off for internal testing, but will probably land it once I'm back.

prelude/haskell/haskell.bzl
```diff
@@ -536,7 +538,8 @@ def _build_haskell_lib(
     lib_short_path = paths.join("lib-{}".format(artifact_suffix), libfile)
 
     linfos = [x.prof_info if enable_profiling else x.info for x in hlis]
-    uniq_infos = dedupe(flatten([x[link_style] for x in linfos]))
+    # only gather direct dependencies
+    uniq_infos = [x[link_style].reduce("root") for x in linfos]
```
avdv, line 542: We just noticed that there is a `value` field available on a tset. This would render the `root` reduction useless. Should I change this and push again, or is there any reason not to use `value`?

ndmitchell: That sounds like a great plan. I've just landed it, so a new diff would be most welcome.

avdv: PR is up: https://github.com/facebook/buck2/pull/621

ndmitchell:
The issue is when you have a Haskell library with `deps = [":a", ":a"]` - so the same dependency twice. That then generates an entry for the Haskell package database which is malformed and GHC complains. I've now added such an instance to our internal test suite (which we're working on open sourcing slowly...) so just trying to reapply.

ndmitchell:
Bad news - there was a significant performance regression from this code. It generates command lines that upset GHC. I don't yet know why, but investigating. Reverted once more.

ndmitchell:
There are two places this code causes a significant performance regression. In both cases it comes from the fact that multiple packages might exist within the same package_db. Before, since we deduplicated everything by value, you might have 100 packages and 20 package_dbs. If you give GHC 100 package dbs (even if there are only 20 unique ones) the performance crashes dramatically. That's especially true if you have a package_db representing all of stackage, which both occurs a lot and is slow to load. I've spun up #624 as a version with your two patches and my deduplication code in. Aiming to land that next week, but being cautious given this has gone wrong twice now 😄

avdv:
> There are two places this code causes a significant performance regression. In both cases it comes from the fact that multiple packages might exist within the same package_db. Before, since we deduplicated everything by value, you might have 100 packages and 20 package_dbs. If you give GHC 100 package dbs (even if there are only 20 unique ones) the performance crashes dramatically.

Oh, that's interesting. Thank you for pushing this forward!

ndmitchell:
To update, there ended up being a third place, which I fixed too. All sorted now and seems to have had no regressions for a while. Thanks for the contribution!

PR #621, opened 2024-04-17
Use `value` attribute of tset for HaskellLibraryInfo

avdv:
This is a follow-up to PR #617 which simplifies accessing direct dependencies for haskell libraries by using the `value` attribute of tsets and removes the `root` reduction for `HaskellLibraryInfoTSet` which was used for this purpose before.


ndmitchell:
Thanks, working on reapplying #617 then this should apply fine.

Issue #612, opened 2024-04-02
How to use multiple execution platforms


Overhatted:
I have a project that can be built for either Linux or Windows, without any cross-compilation so Linux and Windows are also the two execution platforms. Since there is remote execution, I wanted to be able to build for Linux from my Windows machine using RE.

To keep this short I created a sample project: https://github.com/facebook/buck2/compare/main...Overhatted:buck2-strip-prefix:multiple_target_platforms
It takes inspiration from the following issue: #487
The project just runs some Linux commands which obviously doesn't make sense to do on Windows but it's just a sample for me to understand the execution platform resolution.

When running "buck2 build --target-platforms root//platforms:linux tests:tests" on Windows, I wanted it to select the Linux execution platform automatically (since it's the only one with the "config//os/constraints:linux" constraint) but it tries to run it on the Windows platform, presumably because it is the first one declared, so I don't think I'm using the constraints correctly.

To reproduce the project above:
1. Initialize the repository by running: `buck2 init --git`, `cd prelude` and then `git reset --hard 3b5706b425a88a8641aad8d8ffcc6cdf441b9db0`.
2. Fill the correct remote execution information in .buckconfig
3. Change the "container-image" value to point to a Windows and a Linux docker image available on the RE, depending on the value of platform_name.

Running the command, "buck2 build --target-platforms root//platforms:linux tests:tests", I wanted the genrule commands to be ran on Linux but instead they are ran on Windows, demonstrated by the error message I receive: 'Failed to start process: exec: "sh": executable file not found in %PATH%'

How can I add constraints to the execution platform so that Buck2 automatically selects the correct execution platform for the selected target platform?

JakobDegen:
> I wanted it to select the Linux execution platform automatically (since it's the only one with the "config//os/constraints:linux" constraint) but it tries to run it on the Windows platform, presumably because it is the first one declared

Correct.

> How can I add constraints to the execution platform so that Buck2 automatically selects the correct execution platform for the selected target platform?

There is no general way to do that, but I'll explain how to solve your problem before I explain why.

You have a couple options and the best one depends on the particular situation. For this one, you have a `:tests` target that uses a bunch of `cmd_args` that assume things are running on Linux. In that case, it's appropriate to just set `exec_compatible_with = ["config//os:linux"]` on the target. This tells buck that that target can only be built on a Linux execution platform, and so it will not pick an execution platform that doesn't have the Linux constraint. This sort of case is somewhat common.

[...]

PR #629, opened 2024-04-28
Support multiple execution platforms with system_cxx_toolchain

Overhatted:
As mentioned in #612, I have a C++ project that can be built for either Linux or Windows, without any cross-compilation so Linux and Windows are also the two execution platforms. Since there is remote execution, I wanted to be able to build for Linux from my Windows machine using RE, and vice versa.

From the response I got in the above issue, I need to add an `exec_dep` to the toolchain, pointing a compiler target, that then has a `target_compatible_with` with the execution platform the compiler can run on.

I tried to add the `target_compatible_with` directly to the toolchain but it didn't work, probably intentional.

So I moved some parts of the current system_cxx_toolchain to a separate `NativeCompiler` provider that is then returned by some rules defining a compiler. This compiler target then has the `target_compatible_with` attribute.

Can this be merged or should I just have this toolchain in my own repository? It seems useful for everyone, the only disadvantage I can see being that it is not fully backwards compatible. Hopefully I didn't completely misunderstand how execution platforms are supposed to be used.

I improved the Visual Studio example to use this new updated toolchain to easily support remote execution.

## Manual testing

I used the Visual Studio example and setup my RE workers with Buildbarn. Unfortunately Buildbarn's Windows workers don't currently work with the linking step, I haven't investigated why. The build targeting Windows made from a Windows host works because the linking is done locally but the one done from a Linux host fails.

#### Windows
```
buck2 build --target-platforms //buck2_utils/platforms:windows_debug :main
```
Passes
#### Linux
```
buck2 build --target-platforms //buck2_utils/platforms:linux_debug :main
```
Passes

##### Target Windows
```
buck2 build --no-remote-cache -c buck2_re_client.enabled=true --target-platforms //buck2_utils/platforms:windows_debug :main
```
Passes
##### Target Linux
```
buck2 build --no-remote-cache -c buck2_re_client.enabled=true --target-platforms //buck2_utils/platforms:linux_debug :main
```
Passes

##### Target Windows
```
buck2 build --no-remote-cache -c buck2_re_client.enabled=true --target-platforms //buck2_utils/platforms:windows_debug :main
```
Linking fails (The system cannot open the device or file specified.)
##### Target Linux
```
buck2 build --no-remote-cache -c buck2_re_client.enabled=true --target-platforms //buck2_utils/platforms:linux_debug :main
```
Passes

### DIFF
--- a/examples/visual_studio/.buckconfig
+++ b/examples/visual_studio/.buckconfig
@@ -11,7 +11,14 @@ fbsource = none
 buck = none
 
 [parser]
-target_platform_detector_spec = target:root//...->toolchains//:default
+target_platform_detector_spec = target:root//...->root//buck2_utils/platforms:windows_debug
 
 [build]
-execution_platforms = prelude//platforms:default
+execution_platforms = root//buck2_utils/platforms:default
+
+[buck2_re_client]
+action_cache_address = grpc://localhost:8980
+engine_address = grpc://localhost:8980
+cas_address = grpc://localhost:8980
+tls = false
+instance_name = fuse
--- a/examples/visual_studio/buck2_utils/install.ps1
+++ b/examples/visual_studio/buck2_utils/install.ps1
@@ -29,7 +29,8 @@ begin
 			[parameter(Mandatory=$true)] [String] $SourcePath,
 			[parameter(Mandatory=$true)] [String] $TargetPath
 		)
-		Copy-Item $SourcePath -Destination $TargetPath
+		New-Item -ItemType File -Path $TargetPath -Force | Out-Null
+		Copy-Item $SourcePath -Destination $TargetPath | Out-Null
 	}
 }
 process
--- a/examples/visual_studio/buck2_utils/platforms/BUCK
+++ b/examples/visual_studio/buck2_utils/platforms/BUCK
@@ -1,0 +1,38 @@
+load(":defs.bzl", "execution_platform")
+
+execution_platform(
+    name = "default",
+    visibility = ["PUBLIC"],
+)
+
+platform(
+    name = "windows_debug",
+    constraint_values = [
+        "config//os/constraints:windows",
+        "root//buck2_utils/configuration:debug",
+    ],
+)
+
+platform(
+    name = "windows_release",
+    constraint_values = [
+        "config//os/constraints:windows",
+        "root//buck2_utils/configuration:release",
+    ],
+)
+
+platform(
+    name = "linux_debug",
+    constraint_values = [
+        "config//os/constraints:linux",
+        "root//buck2_utils/configuration:debug",
+    ],
+)
+
+platform(
+    name = "linux_release",
+    constraint_values = [
+        "config//os/constraints:linux",
+        "root//buck2_utils/configuration:release",
+    ],
+)
--- a/examples/visual_studio/buck2_utils/platforms/BUCK
+++ b/examples/visual_studio/buck2_utils/platforms/BUCK
@@ -1,6 +1,6 @@
-load(":defs.bzl", "execution_platform")
+load(":defs.bzl", "execution_platforms")
 
-execution_platform(
+execution_platforms(
     name = "default",
     visibility = ["PUBLIC"],
 )
--- a/prelude/toolchains/cxx.bzl
+++ b/prelude/toolchains/cxx.bzl
@@ -20,8 +20,10 @@ load(
 )
 load("@prelude//cxx:headers.bzl", "HeaderMode")
 load("@prelude//cxx:linker.bzl", "is_pdb_generated")
+load("@prelude//decls/common.bzl", "buck")
 load("@prelude//linking:link_info.bzl", "LinkOrdering", "LinkStyle")
 load("@prelude//linking:lto.bzl", "LtoMode")
+load("@prelude//os_lookup:defs.bzl", "OsLookup")
 
 CxxToolsInfo = provider(
     fields = {
@@ -36,7 +38,6 @@ CxxToolsInfo = provider(
         "archiver_type": provider_field(typing.Any, default = None),
         "linker": provider_field(typing.Any, default = None),
         "linker_type": provider_field(typing.Any, default = None),
-        "os": provider_field(typing.Any, default = None),
     },
 )
 
@@ -53,7 +54,6 @@ def _legacy_equivalent_cxx_tools_info_windows(ctx: AnalysisContext, default_tool
         archiver_type = default_toolchain.archiver_type,
         linker = default_toolchain.linker if ctx.attrs.linker == None or ctx.attrs.linker == "link.exe" else ctx.attrs.linker,
         linker_type = default_toolchain.linker_type,
-        os = default_toolchain.os,
     )
 
 def _legacy_equivalent_cxx_tools_info_non_windows(ctx: AnalysisContext, default_toolchain: CxxToolsInfo) -> CxxToolsInfo:
@@ -69,7 +69,6 @@ def _legacy_equivalent_cxx_tools_info_non_windows(ctx: AnalysisContext, default_
         archiver_type = default_toolchain.archiver_type,
         linker = default_toolchain.linker if ctx.attrs.linker == None else ctx.attrs.linker,
         linker_type = default_toolchain.linker_type,
-        os = default_toolchain.os,
     )
 
 def _system_cxx_toolchain_impl(ctx: AnalysisContext):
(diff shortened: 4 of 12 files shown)


prelude/toolchains/cxx/clang/tools.bzl
```diff
@@ -0,0 +1,43 @@
...
+            linker = "clang++",
+            linker_type = "gnu",
+            os = ctx.attrs.os
+        ),
+    ]
+
+path_clang_tools = rule(
+    impl = _path_clang_tools_impl,
+    attrs = {
+        "os": attrs.string(default = _get_current_os()),
```
KapJI, line 41: This means when you want to target Linux from Windows, it's still be "windows" here. You probably want to have something like `_exec_os_type` in prelude.

examples/visual_studio/.buckconfig
```diff
@@ -11,7 +11,14 @@ fbsource = none
 buck = none
 
 [parser]
-target_platform_detector_spec = target:root//...->toolchains//:default
+target_platform_detector_spec = target:root//...->root//buck2_utils/platforms:windows_debug
```
KapJI, line 14: Why change this?

examples/visual_studio/buck2_utils/platforms/BUCK
```diff
@@ -0,0 +1,38 @@
+load(":defs.bzl", "execution_platform")
+
+execution_platform(
```
KapJI, line 3: To make it less confusing.

```suggestion
execution_platforms(
```

prelude/toolchains/cxx.bzl
```diff
@@ -22,52 +22,35 @@ load("@prelude//cxx:headers.bzl", "HeaderMode")
 load("@prelude//cxx:linker.bzl", "is_pdb_generated")
 load("@prelude//linking:link_info.bzl", "LinkOrdering", "LinkStyle")
 load("@prelude//linking:lto.bzl", "LtoMode")
-load("@prelude//toolchains/msvc:tools.bzl", "VisualStudio")
-load("@prelude//utils:cmd_script.bzl", "ScriptOs", "cmd_script")
+
+NativeCompiler = provider(
```
KapJI, line 26: "Compiler" usually means single tool, maybe something like "SystemCxxToolchainInfo"?

prelude/toolchains/cxx.bzl
```diff
@@ -170,49 +163,29 @@ def _system_cxx_toolchain_impl(ctx: AnalysisContext):
...
-        cmd = cmd_args(
-            ctx.attrs.linker_wrapper[RunInfo],
-            linker,
-        ),
-        os = ScriptOs("windows"),
-    )
+def _run_info(args):
+    return None if args == None else RunInfo(args = [args])
+
+def _get_default_compiler() -> str:
```
KapJI, line 169: Should this be selectified?

prelude/toolchains/cxx.bzl
```diff
@@ -170,49 +163,29 @@ def _system_cxx_toolchain_impl(ctx: AnalysisContext):
...
+        return "prelude//toolchains/msvc:msvc_tools"
+    else:
+        return "prelude//toolchains/cxx/clang:path_clang_tools"
 
 system_cxx_toolchain = rule(
     impl = _system_cxx_toolchain_impl,
     attrs = {
+        "compiler": attrs.exec_dep(providers = [NativeCompiler], default = _get_default_compiler()),
         "c_flags": attrs.list(attrs.string(), default = []),
-        "compiler": attrs.string(default = "cl.exe" if host_info().os.is_windows else "clang"),
```
KapJI, line 200: This is a breaking change for anyone who uses these attributes in `system_cxx_toolchain`. I don't think we should remove them.

KapJI (changes requested):

examples/visual_studio/buck2_utils/install.ps1
```diff
@@ -29,7 +29,8 @@ begin
 			[parameter(Mandatory=$true)] [String] $SourcePath,
 			[parameter(Mandatory=$true)] [String] $TargetPath
 		)
-		Copy-Item $SourcePath -Destination $TargetPath
+		New-Item -ItemType File -Path $TargetPath -Force | Out-Null
+		Copy-Item $SourcePath -Destination $TargetPath | Out-Null
```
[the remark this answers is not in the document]
Overhatted: Because Copy-Item doesn't create the target directory if it doesn't exist: https://stackoverflow.com/questions/7523497/should-copy-item-create-the-destination-directory-structure
For example the first time the install script is ran, the target "$InstallDirectory" does not exist which makes Copy-Item fail when it tries to copy "main.exe" to "install/main.exe".

prelude/toolchains/cxx/clang/tools.bzl
Overhatted: Thanks for the tip. I had't realised that the default could also be a select. But here, since the system_cxx_toolchain doesn't do cross-compilation, we want the "os" to be the target OS so I changed this to buck.target_os_type_arg.

examples/visual_studio/.buckconfig
Overhatted: Previously the visual_studio example was using a single target platform (toolchains//:default) that would change one of the constraint_values (debug/release) depending on a read_config call. In this PR I changed the example by creating 4 different target platforms (windows/linux and debug/release) in root//buck2_utils/platforms/BUCK. I only changed this because it seemed like the intended way to use Buck2. Now the target platform is specified by using the "--target-platforms" command line option.

examples/visual_studio/buck2_utils/platforms/BUCK
Overhatted: I did this change manually because I also needed to make the change in buck2_utils/platforms/defs.bzl.

prelude/toolchains/cxx.bzl
Overhatted: Done

prelude/toolchains/cxx.bzl
Overhatted: I selectified this.

prelude/toolchains/cxx.bzl
Overhatted: I reintroduced the attributes and I think the end result is fully backwards compatible. I did some tests on Linux and Windows using those parameters.

Overhatted:
After the review, I made all the changes to this PR in the last 7 commits where the first 6 were done to fix issues pointed out in the review and the last one was done as another change.

Overhatted:
@KapJI Not sure if I'm supposed to ping.

JakobDegen (changes requested):

prelude/toolchains/cxx.bzl
```diff
@@ -22,52 +22,68 @@ load("@prelude//cxx:headers.bzl", "HeaderMode")
 load("@prelude//cxx:linker.bzl", "is_pdb_generated")
 load("@prelude//linking:link_info.bzl", "LinkOrdering", "LinkStyle")
 load("@prelude//linking:lto.bzl", "LtoMode")
-load("@prelude//toolchains/msvc:tools.bzl", "VisualStudio")
-load("@prelude//utils:cmd_script.bzl", "ScriptOs", "cmd_script")
+
+SystemCxxToolchainInfo = provider(
```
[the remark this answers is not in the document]
Overhatted: No problem. I renamed SystemCxxToolchainInfo to CxxToolsInfo.

prelude/toolchains/cxx.bzl
```diff
@@ -170,50 +197,34 @@ def _system_cxx_toolchain_impl(ctx: AnalysisContext):
...
-        ),
-        os = ScriptOs("windows"),
-    )
+def _run_info(args):
+    return None if args == None else RunInfo(args = [args])
 
 system_cxx_toolchain = rule(
     impl = _system_cxx_toolchain_impl,
     attrs = {
+        "toolchain_info": attrs.exec_dep(providers = [SystemCxxToolchainInfo], default = select({
```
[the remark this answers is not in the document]
Overhatted: Thanks for the tip, this API was getting weird indeed. I created a separate rule called `cxx_tools_info_toolchain` (feel free to suggest a better name). I reused almost all code from the previous system_cxx_toolchain. Inclusively, for system_cxx_toolchain, I'm using a attrs.default_only of cxx_tools_info to provide the defaults rather than the previous way of using `default = ... if host_info().is_windows else ...` inside the rule declaration. This avoids repeating the defaults in two different places but let me know if you'd rather I separate these rules further.

prelude/toolchains/cxx.bzl
```diff
@@ -22,52 +22,68 @@ load("@prelude//cxx:headers.bzl", "HeaderMode")
...
-            linker = msvc_tools.link_exe
-        linker = _windows_linker_wrapper(ctx, linker)
+
+    toolchain_info = ctx.attrs.toolchain_info[SystemCxxToolchainInfo]
+    toolchain_info = _legacy_equivalent_toolchain_info_windows(ctx, toolchain_info) if toolchain_info.os == "windows" else _legacy_equivalent_toolchain_info_non_windows(ctx, toolchain_info)
+
+    archiver_supports_argfiles = toolchain_info.os != "macos"
+    additional_linker_flags = ["-fuse-ld=lld"] if toolchain_info.os == "linux" and toolchain_info.linker != "g++" and toolchain_info.cxx_compiler != "g++" else []
+
+    if toolchain_info.os == "windows":
```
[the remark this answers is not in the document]
Overhatted: I was already more-or-less getting this "os" attribute from `_target_os_type`, indirectly though the `SystemCxxToolchainInfo` (now `CxxToolsInfo`). But I did the change you requested, removing the os from `CxxToolsInfo` and adding it directly to the cxx toolchain rules.
I think the change makes sense because the "os" only determines the flags and command line arguments that the `CxxToolsInfo` will run with, it is not a characteristic of the compiler/linker/CxxTool itself. Not sure if this the same reasoning as you but if the end result is the same, it doesn't matter that much.

prelude/toolchains/cxx.bzl
```diff
@@ -170,49 +203,54 @@ def _system_cxx_toolchain_impl(ctx: AnalysisContext):
...
-        ),
-        os = ScriptOs("windows"),
-    )
+def _run_info(args):
+    return None if args == None else RunInfo(args = [args])
 
 system_cxx_toolchain = rule(
     impl = _system_cxx_toolchain_impl,
     attrs = {
+        "cxx_tools_info": attrs.default_only(attrs.exec_dep(providers = [CxxToolsInfo], default = select({
```
JakobDegen, line 212:
```suggestion
        "_cxx_tools_info": attrs.default_only(attrs.exec_dep(providers = [CxxToolsInfo], default = select({
```

Just to make it a little more obvious to users that this is an implementation detail of the rule (ideally we'd support `hide = True` for the docs but that's a problem for another day)

prelude/toolchains/cxx.bzl
```diff
@@ -170,49 +203,54 @@ def _system_cxx_toolchain_impl(ctx: AnalysisContext):
...
+            "DEFAULT": "prelude//toolchains/cxx/clang:path_clang_tools",
+            "config//os:windows": "prelude//toolchains/msvc:msvc_tools",
+        }))),
```
JakobDegen, line 215: So, while using a select to swap between these would normally definitely be the right thing to do, the `system_*_toolchain` rules are a little bit special in that they're supposed to be low-effort defaults that work in somewhat degenerate environments. Relevantly to you, I think for now the expectation probably is that these toolchains should work even while using a platform that does not specify a `config//os:` constraint (as I said, degenerate environments). Now, while I could certainly be convinced that we should change that, I think for the purposes of this PR it's probably better to retain behavior by using a `if host_info` on this rule specifically, instead of a select

JakobDegen (commented):
Just one minor item, otherwise looks great

JakobDegen (approved):

PR #641, opened 2024-05-09
Add find_missing_blobs call on uploads to check if a blob has already been uploaded.

TheGrizzlyDev:
This way the client won't need to reupload existing blobs over and over again which should reduce build duration and network throughput significantly.

### DIFF
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -432,21 +432,21 @@ impl Drop for REClient {
 /// Used to defer reading of NamedDigest contents till
 /// actual execution of upload and prevent opening too many
 /// files at the same time.
-enum BatchUploadRequest {
-    Blob(InlinedBlobWithDigest),
-    File(NamedDigest),
+enum BatchUploadRequest<'a> {
+    Blob(&'a InlinedBlobWithDigest),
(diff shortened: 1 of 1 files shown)


JakobDegen:
Hm, I think we're already doing this, just on the buck2 side:

https://github.com/facebook/buck2/blob/a279e83d6f4b966d1f72b32414ec4c8a88c4c6fe/app/buck2_execute/src/re/uploader.rs#L180

There was some discussion about this not really having all the behaviors that people wanted on discord, but the ask there was specifically for deduplicating racing uploads. I don't know that trying to do this again at a second place in the stack is going to have much benefit

TheGrizzlyDev:
> Hm, I think we're already doing this, just on the buck2 side:
> 
> https://github.com/facebook/buck2/blob/a279e83d6f4b966d1f72b32414ec4c8a88c4c6fe/app/buck2_execute/src/re/uploader.rs#L180
> 
> There was some discussion about this not really having all the behaviors that people wanted on discord, but the ask there was specifically for deduplicating racing uploads. I don't know that trying to do this again at a second place in the stack is going to have much benefit

yeah, my bad, I realised that a bit later and instead of closing this PR I closed https://github.com/facebook/buck2/pull/642 which on the other hand is needed.

PR #642, opened 2024-05-09
Remove arbitrary concurrency limiter for uplods and replace with a configurable limit that is disabled by default.

TheGrizzlyDev:
The existing limit doesn't really help to achieve the goals it states as it cannot help in conditions where multiple clients are using RE at the same time, which are the common exercise conditions. Since the flag can be used to limit quickly uploads in case of some issues with RBE I opten in for keeping it in some form, by making it a setting, though it likely isn't very useful, so it's disabled by default. Instead we should rely on other control mechanisms like number of connections and TCP or HTTP/2 flow control or, like in this case, explicit errors coming from the RE protocol itself (like `RESOURCE_EXHAUSTED` or others).

### DIFF
--- a/app/buck2_re_configuration/src/lib.rs
+++ b/app/buck2_re_configuration/src/lib.rs
@@ -264,6 +264,8 @@ pub struct Buck2OssReConfiguration {
     pub instance_name: Option<String>,
     /// Use the Meta version of the request metadata
     pub use_fbcode_metadata: bool,
+    /// Maximum number of concurrent upload requests.
+    pub max_concurrent_uploads: Option<usize>,
 }
 
 #[derive(Clone, Debug, Default, Allocative)]
@@ -352,6 +354,10 @@ impl Buck2OssReConfiguration {
                     property: "use_fbcode_metadata",
                 })?
                 .unwrap_or(true),
+            max_concurrent_uploads: legacy_config.parse(BuckconfigKeyRef {
+                section: BUCK2_RE_CLIENT_CFG_SECTION,
+                property: "max_concurrent_uploads",
+            })?,
         })
     }
 }
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -76,11 +76,6 @@ use crate::metadata::*;
 use crate::request::*;
 use crate::response::*;
 
-// RBE Services (e.g. Buildbarn) may not be robust against having too many files open at
-// once. Limit to an arbitrary reasonable number since this information is not expressed
-// in the Capabilities message query.
-const CONCURRENT_UPLOAD_LIMIT: usize = 64;
-
 const DEFAULT_MAX_MSG_SIZE: usize = 4 * 1000 * 1000;
 
 fn tdigest_to(tdigest: TDigest) -> Digest {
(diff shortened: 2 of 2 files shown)


JakobDegen:
cc @benbrittain who originally added this

> The existing limit doesn't really help to achieve the goals it states as it cannot help in conditions where multiple clients are using RE at the same time, which are the common exercise conditions

Yeah, but even if they're the common exercise conditions, it's still pretty easy to imagine that they might not be the failure conditions. Ie you could imagine some server that has one thread per client or something like that and falls over if a client shows up trying to write too many big files at once.

I think there's a nice in-between option to go for though, which both preserves this property and sounds closer to what you actually want: Can we only apply this concurrency to the non-batched uploads, and let batched uploads proceed at unlimited concurrency?

TheGrizzlyDev:
> Yeah, but even if they're the common exercise conditions, it's still pretty easy to imagine that they might not be the failure conditions.

Which is why there is an optional configuration to limit concurrency. Any default value would be making likely incorrect assumptions about the implementation and exercise conditions of a RE server and this is just wrong as there are multiple implementations and all of them are wildly different and can run on extremely large distributed clusters with a wide array of different hardware and software specifications. This invariably results in lower performances by default with no real impact. On the other hand if such a limit is necessary for a RE server they can state this setting in their docs. I think this is already a middle ground and brings the potential benefit of limiing concurrency whilst keeping a performant setup by default, wdyt?

TheGrizzlyDev:
friendly ping @JakobDegen :)

TheGrizzlyDev:
Ping :)

valadez-abel:
I tested this on our RE setup, and without it the client seemed to be stuck for 40+ minutes on re_upload stages, but with this PR, it's no longer blocked on re_upload.

zjturner:
Agree that it would be great to see some forward progress on this.  @JakobDegen is there anything you're waiting for?

TheGrizzlyDev:
ping @JakobDegen

JakobDegen:
@TheGrizzlyDev yeah, sorry, I realized that in my previous message I managed to actually not right any of the things that I was actually most worried about.

The reason that I think this kind of a concurrency limiter makes sense isn't so much the server side of things, but rather the client side. When it comes to batched uploads, or other uploads that are coming from memory and not from disk, I think I'm willing to trust the system network stack to make sure that resources are effectively utilized. But for non-batched uploads, those are generally reading from disk as a part of the request, and limiting concurrency the concurrency of disk accesses in buck2 does seem like something we should do, if for no other reason than that we don't want to run out of file descriptors. Thoughts?

TheGrizzlyDev:
> those are generally reading from disk as a part of the request, and limiting concurrency the concurrency of disk accesses in buck2 does seem like something we should do, if for no other reason than that we don't want to run out of file descriptors. Thoughts?

if that's the case then this limiter still wouldn't help and you'd need a completely different type of limiter that controls open file descriptors within a whole build. Also, I find it highly unlikely this could happen, whereas the current limiter is certain to slow down uploads by a huge factor.

zjturner:
I’ll also add that for large projects, you’ll already hit your fd limit just by the fs monitor, so in practice everyone with a large enough project is already going to uncap.

Even beyond that, the concurrency in inherently limited by the max number of concurrent actions, as well as by natural serialization points in the build, which is all to say that the present behavior is far worse than any potential impact caused by increasing the number of file descriptors

TheGrizzlyDev:
@JakobDegen ping :)

cjhopman:
There may be better longer term solutions here, but this seems like a fine improvement over the current state. I think we should just change the key to something like `max_concurrent_uploads_per_action` (it's per upload_request, but that should basically be 1 per action and per_action i think is more understandable). I'd lean towards prefixing it with `oss_` since it's only going to affect the oss client (or maybe with `grpc_`), but either way is fine.

PR #644, opened 2024-05-10
[buck2] Add find_missing cache to oss re client

miaoyipu:
We are trying to avoid calling too many ``find_missing`` for duplicated digests when uploading. This feature originally should enabled on cas daemon but seems missing on oss version. We are seeing 2-4x improvement on the longest re_upload time and aggregated uploading time.


remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -754,19 +761,39 @@ impl REClient {
                         ttl: 60,
                     },
                 );
+                let mut find_missing_cache = self.find_missing_cache.lock().unwrap();
```
JakobDegen, line 764: Might be better to avoid locking this on every iteration?

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -488,6 +493,7 @@ impl REClient {
             grpc_clients,
             capabilities,
             instance_name,
+            find_missing_cache: Mutex::new(LruCache::new(NonZeroUsize::new(50 << 20).unwrap())), // 50Mb
```
JakobDegen, line 496: Since `TDigest`s are small, this is actually really big. That means that it's going to have potential problems with artifact expiration. Can we add some kind of a timestamp to this and then invalidate this cache after, say, 12 hours so that we know we won't have TTL problems?

JakobDegen (approved):
I'll note that this is also sort of halfway to an implementation of upload deduplication - if you were to just replace the `bool` with a

```
enum CacheState {
    AlreadyUploaded,
    Uploading(Future),
}
```

miaoyipu:
@JakobDegen the new revision addressed your comments. I don't have the merge permission. If you feels the PR is good to go, could you help me to merge the PR? Thanks

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -414,11 +419,42 @@ pub struct GRPCClients {
...
+}
+
+struct FindMissingCache {
+    cache: LruCache<TDigest, DigestRemoteState>,
+    ttl: Duration,
+    last_check: Instant,
+}
+
+impl FindMissingCache {
+    fn chear_if_ttl_expires(&mut self) {
```
JakobDegen, line 434:
```suggestion
    fn clear_if_ttl_expires(&mut self) {
```

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -414,11 +419,42 @@ pub struct GRPCClients {
...
 
+enum DigestRemoteState {
+    ExistsOnRemote,
+    Missing
+}
+
+struct FindMissingCache {
+    cache: LruCache<TDigest, DigestRemoteState>,
+    ttl: Duration,
+    last_check: Instant,
```
JakobDegen, line 430:
```suggestion
    /// To avoid a situation where we cache that an artifact is available remotely, but the artifact then expires
    /// we clear our local cache once every `ttl`.
    ttl: Duration,
    last_clear: Instant,
```

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -488,6 +524,11 @@ impl REClient {
             grpc_clients,
             capabilities,
             instance_name,
+            find_missing_cache: Mutex::new(FindMissingCache {
+                cache: LruCache::new(NonZeroUsize::new(50 << 20).unwrap()), // 50Mb
+                ttl: Duration::from_secs(12 * 60 * 60), // 12 hours
```
JakobDegen, line 529:
```suggestion
                ttl: Duration::from_secs(12 * 60 * 60), // 12 hours. TODO: Tune this parameter
```

JakobDegen (commented):
Just a couple more nits, will merge after

PR #652, opened 2024-05-15
Make prelude support nasm

zjturner:
We use a certain open source third party library which has some assembly files which can only be built with [nasm](https://nasm.us/).  nasm has a couple of command line differences that are relevant for the purposes of getting minimal nasm support:

1. response files don't use the syntax `@foo.argsfile`, rather they use the syntax `-@foo.argsfile`.
2. Content of response files cannot be quoted.
3. src file input is not prefixed with `-c`.

Since our toolchain now conditionally uses nasm or the default toolchain assembler, we are taking advantage of `cxx_toolchain_override()`.  This was missing the ability to override some of the assembler arguments (e.g. we need to set `compiler_type="nasm"`), and also we discovered a few bugs where some fields were not null checked.  So all of this is fixed in this PR.

One thing I don't entirely understand is depfiles, or if I need to do anything.  nasm accepts a couple of relevant command line options, which I don't fully understand how to use

```
    -M            generate Makefile dependencies on stdout
    -MG           d:o, missing files assumed generated
    -MF file      set Makefile dependency file
    -MD file      assemble and generate dependencies
    -MT file      dependency target name
    -MQ file      dependency target name (quoted)
    -MP           emit phony target
```

I can get a `.d` file to generate by passing `-MD /path/to/d/file -MT /path/to/obj/file` but I don't know what's going to happen if I omit this.  Everything seems to work for me.  but maybe it's because none of my asm files depend on any other external files.  Willing to investigate this further with some guidance, but the PR already works for me as is, so if people are ok with the current missing depfile support, so am I.


JakobDegen (approved):

JakobDegen:
Gh says there are conflicts, could you rebase? Will import afterwards

zjturner:
Done

PR #661, opened 2024-05-22
Fix str being returned instead of Path() in get_sdk10_dir

samarunraj:
When both WindowsSdkDir and WindowsSDKVersion environment variables are set, **get_sdk10_dir()**
returns the path set in WindowsSdkDir as a str instead of Path which will then cause a type error
(TypeError: unsupported operand type(s) for /: 'str' and 'str')  in **find_with_vswhere_exe()** when
the returned path is used.


JakobDegen:
Thanks!

PR #671, opened 2024-06-04
Only absolutize exe if path is a file and executable

### LINKED ISSUE #670: Failed to spawn a process -- buck2 tries to execute directory
In our project I declared an action:
```starlark
    nix_build = cmd_args([
        "nix",
        "build",
        "--no-update-lock-file",
        "--no-use-registries",
        cmd_args("--out-link", out_link.as_output()),
        cmd_args(drv, format = "{}^*"),
    ], hidden = deps)
    ctx.actions.run(nix_build, category = "nix_build", identifier = package, local_only = True)
```
When running that action I got a mysterious error:
```
Action failed: toolchains//:ghc_package_db (nix_build word8)
Local command returned non-zero exit code <no exit code>
Reproduce locally: `env -- 'BUCK_SCRATCH_PATH=buck-out/v2/tmp/toolchains/904931f735703749/__ghc_package_db__/nix_build/w ...<omitted>... ge_db__/__action__1__/word8/out.link /nix/store/wxkgbc4hz6y85zvg2xs81mjg8ha6frff-word8-0.1.3.drv'^*' (run `buck2 log what-failed` to get the full command)`
stdout:
stderr:
Spawning executable `nix` failed: Failed to spawn a process
```
When trying to reproduce it locally, using the command from the log, it works:
```console
$ env -- 'TMPDIR=/home/claudio/source/repo/buck-out/v2/tmp/toolchains/904931f735703749/_ghc_package_db__/nix_build/word8' 'BUCK_SCRATCH_PATH=buck-out/v2/tmp/toolchains/904931f735703749/_ghc_package_db__/nix_build/word8' BUCK2_DAEMON_UUID=e24cf324-c871-4855-99a9-a724821100c3' 'BUCK_BUILD_ID=b414deac-6ea2-4aa1-ac30-d3413d518fa1' nix build --no-update-lock-file --no-use-registries --out-link buck-out/v2/gen/toolchains/904931f735703749/__ghc_package_db__/__action__1__/word8/out.link /nix/store/wxkgbc4hz6y85zvg2xs81mjg8ha6frff-word8-0.1.3.drv'^*'
$ echo $status
0
```
Running buck2 under strace, revealed:
```
execve("/home/claudio/source/repo/nix", ["/home/claudio/source/repo/nix", "build", "--no-update-lock-file", "--no-use-registries", "--out-link", "buck-out/v2/gen/toolchains/904931f735703749/__ghc_package_db__/__action__1__/base-compat/out.link", "/nix/store/y3zshdaaxfzq64ikjd1w5h4hh65rxi9y-base-compat-0.12.3.drv^*"], 0x7f33e9c4d400 /* 173 vars */) = -1 EACCES (Permission denied)
```
So, there is indeed a directory called `nix` in the project root directory which it is trying to execute.

I think this is because the code introduced in https://github.com/facebook/buck2/commit/37f3d258887d0c4565a23eafb2520057d923c6ac does only check if the file exists, not that it is an executable non-directory:
https://github.com/facebook/buck2/blob/85f33c6f9645df3ca4cf13d1f1c0883653cec234/app/buck2_forkserver/src/run.rs#L348-L360

Also, it would have been nice to have a better error message that mentions the real cause of the error and the file it tried to execute.

My current workaround is to wrap command with `env -- `:
```starlark
    nix_build = cmd_args([
        "env", "--",
        "nix",
        "build",
        "--no-update-lock-file",
        "--no-use-registries",
        cmd_args("--out-link", out_link.as_output()),
        cmd_args(drv, format = "{}^*"),
    ], hidden = deps)
```
That should work as long as nobody creates an `env` directory I guess... :smile:

avdv:
- Fix typo
- Only absolutize exe if abs path refers to a file and is executable

Fixes #670


app/buck2_forkserver/src/run.rs
```diff
@@ -353,7 +354,20 @@ pub fn maybe_absolutize_exe<'a>(
 
     let abs = spawned_process_cwd.join(exe);
     if fs_util::try_exists(&abs).context("Error absolute-izing executable")? {
-        return Ok(abs.into_path_buf().into());
+        let metadata = fs::metadata(&abs).context("Error getting metadata for path")?;
```
JakobDegen, line 357: This looks like a bit of a race condition to me. Can you add `fs_util::metadata_if_exists` instead and then use that?

JakobDegen (changes requested):
Sorry this took me so long to get to. Approach generally looks correct

app/buck2_forkserver/src/run.rs
avdv: Good catch! Done.

JakobDegen:
> I think the fix is incorrect. I mean, it fixes the problem reported by a user, but instead this should should just avoid "absolutizing" if given path does not contain slashes (so it is searched in $PATH).

> Current fix has the same problem as before: if there's nix executable file in the current directory, it will be executed instead of nix in $PATH.

-- @stepancheg 

This seems like a reasonable alternative to me, thoughts?

avdv:
> This seems like a reasonable alternative to me, thoughts?

Yes, maybe it's a good idea to take a step back. The problem resolved by https://github.com/facebook/buck2/commit/37f3d258887d0c4565a23eafb2520057d923c6ac was described like this:
> If you do `Command::new("a").current_dir("b")` with your cwd being `c`, on UNIX you run `c/b/a`, on Windows you run `c/a`.

As noted by @stepancheg, that is not really true, since on Unix, you run "a" found in the PATH , the binary relative to `c/b` is not picked up (except if `.` is in the PATH).

On Windows, however, a binary is first picked up from the current directory, before doing a PATH lookup, AFAIK.

Since this is solving a problem specific to Windows, the fix should actually only be applied on Windows. How about *never* absolutize on unix, but check if the executable exists in the spawned processes' cwd and return the absolute path for that on Windows.

I am just wondering though, you would have to include the file extension in the command currently, otherwise this fails right? I mean, Windows looks for various file extensions it denotes as executable (like .exe, .com, .cmd, .bat et cetera) and this would fall short if you mean to run "c/b/nix.exe" but only specify `nix` as the command...

JakobDegen:
> How about never absolutize on unix

This part seems right to me

> but check if the executable exists in the spawned processes' cwd and return the absolute path for that on Windows

This depends - does window generally behave as if `.` is a part of the path? If not, we should consider restricting this behavior to the case where the path has at least one `/`

> I am just wondering though, you would have to include the file extension in the command currently, otherwise this fails right? I mean, Windows looks for various file extensions it denotes as executable (like .exe, .com, .cmd, .bat et cetera) and this would fall short if you mean to run "c/b/nix.exe" but only specify nix as the command

This seems correct. It's possible that this means we have to re-implement a significant part of Window's command resolution.

I guess it would probably be a good start if someone could properly document what it even is that Windows does here, and then we can go from there

PR #680, opened 2024-06-14
re_grpc_proto: add serde support

sluongng:
Add support for serde::{Serialize, Deserialize} to remote execution
protobuf.
Short term, this enables quick debugging use cases in OSS setup as user
could print out the requests and responses using serde_json.

### DIFF
--- a/app/buck2_data/src/lib.rs
+++ b/app/buck2_data/src/lib.rs
@@ -14,7 +14,7 @@ use std::fmt;
 
 pub mod action_key_owner;
 
-mod serialize_timestamp {
+pub mod serialize_timestamp {
     use serde::Deserialize;
     use serde::Deserializer;
     use serde::Serialize;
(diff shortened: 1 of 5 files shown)


sluongng:
This works with Cargo build... but I have no clue how to turn the build.rs file into something buck2 compatible. Pointers would be much appreciated.

JakobDegen:
Sorry for taking a while to get to this.

👍 on the overall initiative, I guess my one comment is that this is a different approach than what we do in buck2_data:

https://github.com/facebook/buck2/blob/757832ba876eed5f768916180b722a33a6173165/app/buck2_data/build.rs#L133

While I would be ok to take this if your approach is clearly necessary for some reason, I'd generally be in favor of a bit of consistency otherwise

sluongng:
@JakobDegen Yeah I saw the current approach during my initial investigation, however, I find it a bit tedious to maintain custom serialization code, especially when it's a solved problem in the open source ecosystem.

However, it's not a hill I am willing to fight for so I pushed a new version that follows the existing style. PTAL! 🙏

Issue #226, opened 2023-05-02
Open API design for BuckEvent data


sluongng:
Today, the build telemetry is done pretty much through Scribe (Meta's internal message queue system) into Scuba (Meta's telemetry platform).

This resulted in Buck2 implementing the EventSink trait for Scribe only:

```rust
/// A sink for events, easily plumbable to the guts of systems that intend to produce events consumeable by
/// higher-level clients. Sending an event is synchronous.
pub trait EventSink: Send + Sync {
    /// Sends an event into this sink, to be consumed elsewhere. Explicitly does not return a Result type; if sending
    /// an event does fail, implementations will handle the failure by panicking or performing some other graceful
    /// recovery; callers of EventSink are not expected to handle failures.
    fn send(&self, event: BuckEvent);

    /// Sends a control event into this sink, to be consumed elsewhere. Control events are not sent to gRPC clients.
    fn send_control(&self, control_event: ControlEvent);

    /// Collects stats on this sink (e.g. messages accepted, rejected).
    fn stats(&self) -> Option<EventSinkStats>;
}
```

In here:

- `BuckEvent` is defined in `buck_data/data.proto` which is great for backward compatibility

- `ControlEvent` is an enum
   ```rust
   /// An event that can be produced by Buck2 that is not intended to be presented to the user, but rather is used to
   /// communicate with other parts of Buck2.
   #[derive(Clone, From)]
   pub enum ControlEvent {
       /// A command result, produced upon completion of a command.
       CommandResult(Box<CommandResult>),
       /// A progress event from this command. Different commands have different types.
       PartialResult(PartialResult),
   }
   ```
   where `CommandResult` and `PartialResult` are both defined in `buck2_cli_proto/daemon.proto`. Not sure yet how backward 
   compatible guarentee this is just yet.

- `EventSinkStats` is a struct defined in `buck2_events` crate, not guarantee to be backward compatible.

---

It would be nice if we could define the telemetry data in a unified interface/data format similar to [Bazel's BEP](https://bazel.build/remote/bep) so that third parties could implement the server side for it. At the very least, having the data structures defined in a unified protobuf would help assuring that the future changes are backward compatible.

krallin:
Is this what you're looking for?

https://github.com/facebook/buck2/blob/main/app/buck2_data/data.proto

cjhopman:
[...]

PR #685, opened 2024-06-17
BuckEvent: provide a Buck Event Publisher proto

sluongng:
Provide a BuckEvent Publisher service which emits BuckEvents to an
external server implementation.

Closes https://github.com/facebook/buck2/issues/226

### DIFF
--- a/app/buck2_event_publisher_proto/event_publisher.proto
+++ b/app/buck2_event_publisher_proto/event_publisher.proto
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under both the MIT license found in the
+ * LICENSE-MIT file in the root directory of this source tree and the Apache
+ * License, Version 2.0 found in the LICENSE-APACHE file in the root directory
+ * of this source tree.
+ */
+
+syntax = "proto3";
+
+import "data.proto";
+
+package event_publisher;
+
+message BuckEventRequest {
+  // A trace-unique 64-bit identifying the stream.
+  uint64 stream_id = 1;
+
+  buck.data.BuckEvent event = 2;
+};
+
+message BuckEventResponse {
+  // A trace-unique 64-bit identifying the stream.
+  uint64 stream_id = 1;
+
+  // The trace ID of the event that has been committed.
+  uint64 trace_id = 2;
+};
+
+service BuckEventPublisher {
+  rpc StreamBuckEvent(stream BuckEventRequest) returns (stream BuckEventResponse);
+};
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ members = [
     "app/buck2_event_observer",
     "app/buck2_events",
     "app/buck2_event_log",
+    "app/buck2_event_publisher_proto",
     "app/buck2_execute",
     "app/buck2_execute_impl",
     "app/buck2_external_cells",
--- a/app/buck2_event_publisher_proto/BUCK
+++ b/app/buck2_event_publisher_proto/BUCK
@@ -1,0 +1,18 @@
+load("@fbcode//buck2:proto_defs.bzl", "rust_protobuf_library")
+load("@fbsource//tools/build_defs:glob_defs.bzl", "glob")
+
+oncall("build_infra")
+
+rust_protobuf_library(
+    name = "buck2_event_publisher_proto",
+    srcs = glob(["src/**/*.rs"]),
+    build_script = "build.rs",
+    build_env = {
+        "BUCK_HACK_DATA_PROTOC_INCLUDE": "$(location //buck2/app/buck2_data:data_proto)",
+    },
+    protos = ["event_publisher.proto"],
+    deps = [
+        "fbsource//third-party/rust:tonic",
+        "//buck2/app/buck2_data:buck2_data",
+    ],
+)
--- a/app/buck2_event_publisher_proto/Cargo.toml
+++ b/app/buck2_event_publisher_proto/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "buck2_event_publisher_proto"
+
+edition = "2021"
+license = { workspace = true }
+repository = { workspace = true }
+version = "0.1.0"
+
+[dependencies]
+prost = { workspace = true }
+tonic = { workspace = true }
+
+buck2_data = { workspace = true }
+
+[build-dependencies]
+buck2_protoc_dev = { workspace = true }
--- a/app/buck2_event_publisher_proto/build.rs
+++ b/app/buck2_event_publisher_proto/build.rs
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under both the MIT license found in the
+ * LICENSE-MIT file in the root directory of this source tree and the Apache
+ * License, Version 2.0 found in the LICENSE-APACHE file in the root directory
+ * of this source tree.
+ */
+
+use std::env;
+use std::io;
+use std::path::PathBuf;
+
+fn main() -> io::Result<()> {
+    let proto_files = &["event_publisher.proto"];
+
+    let data_include = if let Ok(value) = env::var("BUCK_HACK_DATA_PROTOC_INCLUDE") {
+        let path = PathBuf::from(value);
+        path.parent().unwrap().to_str().unwrap().to_owned()
+    } else {
+        "../buck2_data".to_owned()
+    };
+
+    buck2_protoc_dev::configure()
+        .setup_protoc()
+        .extern_path(".buck.data", "::buck2_data")
+        .compile(proto_files, &[".", &data_include])
+}
--- a/app/buck2_event_publisher_proto/src/lib.rs
+++ b/app/buck2_event_publisher_proto/src/lib.rs
@@ -1,0 +1,12 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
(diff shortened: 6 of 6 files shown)


app/buck2_event_publisher_proto/event_publisher.proto
```diff
@@ -0,0 +1,33 @@
...
+message BuckEventRequest {
+  // A trace-unique 64-bit identifying the stream.
+  uint64 stream_id = 1;
+
+  buck.data.BuckEvent event = 2;
+};
+
+message BuckEventResponse {
+  // A trace-unique 64-bit identifying the stream.
+  uint64 stream_id = 1;
```
JakobDegen, line 25: > Instead of having stream_id we can just create a new grpc call for each stream, that would be more natural.

- @stepancheg

JakobDegen (commented):
Copying some comments left on phabricator:

> I think we need the rest of the stack to accept this.
> Is it not clear how this will be used.
> (Also I'm not 100% sure grpc is the right choice, maybe simple piping to stdin would be sufficient).

I think I agree that I'd like to see a little more code before landing this just yet

app/buck2_event_publisher_proto/event_publisher.proto
JakobDegen: Actually, to partially respond to Stiopa on this: That seems hard to practically implement on the buck2 side, since we may create a number of different instances of the client for each command. And even if we didn't do that, it seems unwise to prevent ourselves from doing that in the future

JakobDegen:
Actually, I have another thing to consider: Internally, we've found that the size of the event logs can be extremely substantial and that keeping the entire mechanism reliable hasn't been super straightforward. Now, while I don't if anyone else's repo is big enough to be affected by that, and I'm sure that some of the unreliability is a result of misbehaviors specific to our infra, I can also imagine that some of the lessons learned from that are more generally useful.

To go into detail a bit, the way our scribe client works is roughly that it allocates the "real" client as a global, and gets a handle to it for each instance of the type you see. The global then keeps a queue of messages that is used to 1) batch writes, and 2) rate limit itself. While the rate limiting is probably not necessary in OSS, the write batching might be of interest. That might be something to consider for this API. Hopefully that information also explains what `send_now` does. I'm going to try and get our client open sourced for you to at least look at and evaluate what parts of it you want to reuse or not

sluongng:
There seem to be 2 concerns here:

1. Client initialization and `stream_id`:

I plan to mimic Bazel's approach here. In Bazel, each "invocation" gets a unique stream ID. As events get sent to the server, the server will reply with "ack" (identified via trace_id) for events that have persisted. Network disruption could happen and in such cases, the client might want to retry sending events that were not "ack" by the server. Upon resend, the events could be sent to a different server instance, thus the `stream_id` is needed to identify the retry stream is the same as the previous disrupted stream.

2.

sluongng:
> I think we need the rest of the stack to accept this.
Is it not clear how this will be used.
(Also I'm not 100% sure grpc is the right choice, maybe simple piping to stdin would be sufficient).

From my perspective, it could be potentially a huge time investment to code the rest of the implementation without agreeing on the spec first. At least some agreements on the grpc service would be a positive signal for me to put more time into this.

The current design is modeled after `PublishBuildToolEventStream` https://github.com/bazelbuild/bazel/blob/0b280ac442aa51c7cca1c311658dbd1a6346dff5/third_party/googleapis/google/devtools/build/v1/publish_build_event.proto#L72-L84. Which is a battle-tested RPC with multiple server implementations available among the REAPI ecosystem. A sample client implementation is available in [Bazel (java)](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/buildeventservice/client/BuildEventServiceClient.java;drc=74e51890398d513694e48b79ee30bb8b99c314fc) as well as plenty of proxy implementations available on Github.

With that said, I do understand that this is going to be a Buck2-specific API. So if there are any alternative proposals, I would love to hear them.

JakobDegen:
> Could you please elaborate a bit more on why batch sending needs to be implemented on the client side?

So the thing to keep in mind is that our event sending is done synchronously in a bunch of latency sensitive places. While a benchmark would certainly be much better than guessing, my expectation is that doing a network op for each of these events might be prohibitively slow, at least in some of our builds.

That being said, I don't know that we have to figure this out now. Happy to let you experiment and see if it matters or not.

> From my perspective, it could be potentially a huge time investment to code the rest of the implementation without agreeing on the spec first. At least some agreements on the grpc service would be a positive signal for me to put more time into this.

Yeah, sorry, I was a bit too brief before in just asking for more code. I think you're right to ask for alignment on use of a GRPC API in general, and I'll bring that up at our team meeting tomorrow to make sure we're all good to move forward in general.

The main thing I would ask for from your side is that the service definition itself is clearly marked as being unstable until we've gathered some experience with using it. The way we've reported events to our own service has changed quite a lot in the past, and I imagine that much of that configuration may at some point need room in the API (the first thing that comes to mind is that we only send a subset of events right now, and I imagine not all consumers are going to agree on which subset is the right one...).

Other than that, I don't think it makes much of a difference whether we merge this PR now or wait for some additional code on the client before doing so, but hopefully the things I mentioned above are enough for you to be able to continue work on your side

sluongng:
> Yeah, sorry, I was a bit too brief before in just asking for more code. I think you're right to ask for alignment on use of a GRPC API in general, and I'll bring that up at our team meeting tomorrow to make sure we're all good to move forward in general.

Thank you so much. Looking forward to this.

> So the thing to keep in mind is that our event sending is done synchronously in a bunch of latency sensitive places. While a benchmark would certainly be much better than guessing, my expectation is that doing a network op for each of these events might be prohibitively slow, at least in some of our builds.

This is a good call out.

In Bazel, there is a flag that could switch sending build events asynchronously vs synchronously. The reason for synchronous is that in many OSS setups, the CI worker is an ephemeral CI container/VM that will get shut down right after the build is finished. In those scenarios, it might benefit folks to turn on synchronous event sending so that their builds would wait for all events to be sent before exiting and initializing container shutdown.

I think sending events in async/sync mode is tangent to sending events via stream/batch request. Let's not conflate the 2. There are ways to implement async event sink while using grpc bidi stream as well as batch request while supporting queue by-passing for send_now. For example, we could maintain an internal priority queue in Buck Daemon and send_now events will always get the highest priority. I think we could hash this out in future PRs when we dive into implementation details.

> the service definition itself is clearly marked as being unstable until we've gathered some experience with using it.

I have no problem with this. If there is any existing convention in Buck2 to mark unstable service, I would be happy to follow. If there isn't, I will name the configuration key to activate this `unstable_buck_event_service = grpcs://my.service.com/buckevent` or `experimental_buck_event_service = grpcs://my.service.com/buckevent`. We could also rename the `.proto` file or the Service name itself, although I think it's a bit overkill.

Let me know what you prefer here.

> Other than that, I don't think it makes much of a difference whether we merge this PR now or wait for some additional code on the client before doing so, but hopefully the things I mentioned above are enough for you to be able to continue work on your side

I think the main diff will be who has to pay the rebase cost.

If the review time for future PRs is short, I have no problem leaving this unmerged. However, if the lead time is long, I will have to rebase my PR stack and account for existing refactoring(codemod) in Buck2 repo, which could be a pain to deal with. So I would prefer to have this merged earlier rather than later.

sluongng:
@JakobDegen friendly ping. I am looking for a confirmation on the gRPC API direction before investing more time into this.

JakobDegen:
Alright, sorry for the delay. We had some back and forth on this internally and our ourselves not quite sure what we think the right answer is.

> The reason for synchronous is that in many OSS setups, the CI worker is an ephemeral CI container/VM that will get shut down right after the build is finished. In those scenarios, it might benefit folks to turn on synchronous event sending so that their builds would wait for all events to be sent before exiting and initializing container shutdown.

So we actually have the same problem, which is why even though our event sending is asynchronous in the sense that it doesn't do a network op on each `send` call, we will still make sure to flush all our buffers at the end of the command before returning to the user. I don't think this is the reason you should be electing to go full sync.

> I think sending events in async/sync mode is tangent to sending events via stream/batch request. Let's not conflate the 2. There are ways to implement async event sink while using grpc bidi stream as well as batch request while supporting queue by-passing for send_now. For example, we could maintain an internal priority queue in Buck Daemon and send_now events will always get the highest priority. I think we could hash this out in future PRs when we dive into implementation details.

👍 

> I have no problem with this. If there is any existing convention in Buck2 to mark unstable service

I think a comment at the top of the proto file would be more than enough

> If the review time for future PRs is short, I have no problem leaving this unmerged. However, if the lead time is long, I will have to rebase my PR stack and account for existing refactoring(codemod) in Buck2 repo, which could be a pain to deal with. So I would prefer to have this merged earlier rather than later.

Yeah that's fair. Let's merge this then and we can iterate

I'll add one additional point that came up internally:

Given how big logs can sometimes be, and that users are often working from home on not-great internet (or god forbid, from a hotel or plane or something), the default state was that we had complete data for not even 90% of builds. Fixing that to get 99% complete data was a significant investment for us. The part where we currently have machine local queues in a couple different places that allow us to bridge any periods of reduced internet connectivity were a very big contributing part of that. It's not blocking, but I wanted to at least raise awareness of that so you get a chance to consider it.

If you're otherwise comfortable with this, let me know and I'll go ahead and merge it

aherrmann:
Related: https://github.com/facebook/buck2/pull/811 provides a draft implementation of BES support for Buck2.

iguridi:
@sluongng friendly ping. I want to confirm this PR is still the desired way to go forward on this

sluongng:
Im closing this in favor of the approach laid out in https://github.com/googleapis/googleapis/pull/988#issuecomment-2718603564

Would love to get some feedbacks from Meta side about that approach 🙏

PR #687, opened 2024-06-18
Rename the `root` cell

cbarrete:
This commit renames the `root` cell to facilitate embedding it as a dependency to Rust projects. While the Buck2 repository does not have an official public API, depending on it is the most reasonable way of writing custom test runners, which is important for OSS users until TestInfo V2 lands.

`root` is a common name for root cells, it is generic and it is the default when running `buck2 init`, making it a bad name that is not descriptive and likely to conflict with downstream cells.

As described in the added docs, this commit is not an ideal solution, since it requires configuration hacks:
- The root `.buckconfig` must point to Buck2's shim directory.
- Buck2's `.buckconfig` must be edited to remove cells that are already defined in the root.

Relates to #684.

### DIFF
--- a/.buckconfig
+++ b/.buckconfig
@@ -1,6 +1,6 @@
 [repositories]
-root = .
 prelude = prelude
+buck2 = .
 shim = shim
 
 [repository_aliases]
@@ -14,4 +14,4 @@ buck = shim
 bazel_skylib = shim
 
 [parser]
-target_platform_detector_spec = target:root//...->prelude//platforms:default target:shim//...->prelude//platforms:default
+target_platform_detector_spec = target:buck2//...->prelude//platforms:default target:shim//...->prelude//platforms:default
--- a/docs/users/advanced/buck2_library.md
+++ b/docs/users/advanced/buck2_library.md
@@ -1,0 +1,32 @@
+---
+id: buck2_library
+title: Consuming code from the `buck2` repository
+---
+
+It is possible to depend on targets from the Buck2 repository, which you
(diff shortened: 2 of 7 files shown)


cbarrete:
Hello @JakobDegen, do you know who I should tag to have a look at this? I'm very open to feedback, especially on the caveats that I outlined. I'd like to remove them, but I don't know what would be a good way.

zjturner:
What about places like this that use an implicit root?

https://github.com/facebook/buck2/blob/main/app/buck2_test_runner/BUCK#L18-L21

cbarrete:
I also thought they would have to be made explicit, but as I understand it, a bare `//` means "the current cell" rather than "the root cell". If I'm not mistaken, there are no further changes required, which is nice.
Since the build/CI pass just fine, and I don't have root cell conflicts anymore with my experiment, I'm fairly confident that this is correct.

cbarrete:
I'm not sure who else to tag, could you point me to the right person @ndmitchell?
I'm also happy to update `CONTRIBUTING.md` with how it is appropriate to ask for review of pull requests if you let me know what is preferred.

JakobDegen:
Sorry this took so long. This looks fine to me in principle, I'll see what I need to do to get it through CI internally too

JakobDegen:
Oh, oof, could you rebase this? Otherwise should be easy to get merged

cbarrete:
@JakobDegen Is something blocking this PR?

cbarrete:
No problem, thanks for doing this!

Issue #571, opened 2024-02-20
C++ mixed linking


cbarrete:
I'm trying to build a `cxx_binary` with mixed linking: third-party code is consumed as shared objects and internal libraries are consumed as static libraries.

Is this use case supported by the prelude, and if so, how?

It seems that setting `link_style` in a `cxx_binary` target affects all dependencies: if I set it to `static`, the third-party `prebuilt_cxx_library` shared objects are missing in the `linker.argsfile`, if I set it to `dynamic`, my own libraries are built as shared objects. Setting the `link_style` per `cxx_library` seems to have no effect.

cbarrete:
Here's a minimal repro on top of a bare `buck2 init --git` with the 2024-02-15 release:
```diff
diff --git a/BUCK b/BUCK
index 1cb6b38..5d96647 100644
--- a/BUCK
+++ b/BUCK
@@ -1,7 +1,18 @@
-# A list of available rules and their signatures can be found here: https://buck2.build/docs/api/rules/
+cxx_binary(
+    name = "test",
+    srcs = ["main.cpp"],
+    deps = [":shared_lib", ":static_lib"],
+    link_style = None,
+)
+
+cxx_library(
+    name = "shared_lib",
+    srcs = ["shared_lib.cpp"],
+    link_style = "shared",
+)
 
-genrule(
-    name = "hello_world",
-    out = "out.txt",
-    cmd = "echo BUILT BY BUCK2> $OUT",
+cxx_library(
+    name = "static_lib",
+    srcs = ["static_lib.cpp"],
+    link_style = "static",
 )
diff --git a/main.cpp b/main.cpp
new file mode 100644
index 0000000..d0b8ca7
--- /dev/null
+++ b/main.cpp
@@ -0,0 +1,8 @@
+void dynamic_lib_fn();
+void static_lib_fn();
+
+int main()
+{
+    dynamic_lib_fn();
+    static_lib_fn();
+}
diff --git a/shared_lib.cpp b/shared_lib.cpp
new file mode 100644
index 0000000..6451836
--- /dev/null
+++ b/shared_lib.cpp
@@ -0,0 +1,3 @@
+void dynamic_lib_fn()
+{
+}
diff --git a/static_lib.cpp b/static_lib.cpp
new file mode 100644
index 0000000..28ce184
--- /dev/null
+++ b/static_lib.cpp
@@ -0,0 +1,3 @@
+void static_lib_fn()
+{
+}
```
[...]

PR #692, opened 2024-06-20
Document `link_style` and `preferred_linkage` for C++ rules

cbarrete:
Based on https://github.com/facebook/buck2/issues/571#issuecomment-1960032466.

Also use enums instead of strings where relevant.

<hr>

This commit does not add docs for _all_ the attributes with those names (e.g. in the Apple or Python ecosystems) because I am not familiar with those.


cbarrete:
@cjhopman is this commit stuck somewhere internally? I haven't seen it synced out to the OSS version.

PR #693, opened 2024-06-20
github: build aarch64-linux binaries with 16k-page jemalloc

thoughtpolice:
jemalloc builds in the system page size as part of its (internal) ABI at compile time, which means that if you build with a smaller page size than what your system allows, it doesn't work.

In particular, this impacts Linux on Apple Silicon, which doesn't natively support 4k pages which are the default. Set it to 16k so that the prebuilt binaries work OOTB. **EDIT**: Apparently the Raspberry Pi 5 ships with a 16k page kernel by default, as well.

### DIFF
--- a/.github/workflows/upload_buck2.yml
+++ b/.github/workflows/upload_buck2.yml
@@ -93,6 +93,12 @@ jobs:
         env:
           RUSTFLAGS: "-C strip=debuginfo -C codegen-units=1"
         run: |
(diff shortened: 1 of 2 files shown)


thoughtpolice:
Tested with the following binaries, which seemed to work: https://github.com/thoughtpolice/buck2/actions/runs/9605026647

JakobDegen (approved):

PR #701, opened 2024-06-21
haskell: test example on macos

shayne-fletcher:
Summary: re-enable the haskell example on darwin/x86_64.

Differential Revision: D58876986


joel8485:
Hola

PR #702, opened 2024-06-21
Add a `--build` flag to `buck2` test

cbarrete:


cbarrete:
@ndmitchell @JakobDegen Hello, would you mind having a look when you have a moment?
Thanks a lot!

cbarrete:
Thanks, I look forward to hearing from this!

I don't know about the history either, but I can see the current behavior making sense as well, as a target pattern which matches expensive targets (e.g. binaries that pull in a lot of targets and/or are heavily templated; packages/release tarballs/OCI images/etc.) would cause a "simple" `buck2 test` to be surprisingly expensive, but I personally wouldn't mind making this the default (potentially with a flag to opt out/only build what is necessary for tests).

JakobDegen:
One additional thought: If we do go the direction of a flag, we should probably use the same `--build-{default,run}-info` flags that `buck2 build` uses, instead of coming up with something new. But let's agree on whether we should have a flag at all before getting into that.

Also, @cbarrete , can you say a little bit about your use case for this?

cbarrete:
I just used `--build` because that's what Neil suggested first, likely without thinking too much about it. I'm happy to rename it to whatever fits best, change the details of how things get wired, etc. I'll leave the bikeshedding to people who actually have a clear vision for the CLI!

> Also, @cbarrete , can you say a little bit about your use case for this?

Ultimately, I'm trying to simplify/improve our CI setup. Right now we have a script that looks vaguely like:
```sh
targets=$(supertd stuff)
buck2 build --skip-incompatible-targets $targets
buck2 build --skip-incompatible-targets $targets --target-platforms //non/default:platform
buck2 build --skip-incompatible-targets $targets --target-platforms //some/other:platform

buck2 test --skip-incompatible-targets $targets
buck2 test --skip-incompatible-targets $targets --target-platforms //non/default:platform
buck2 test --skip-incompatible-targets $targets --target-platforms //some/other:platform
```

Which I'd ideally like to turn into:
```sh
targets=$(supertd stuff)
buck2 test $targets --configurations //default:platform //non/default:platform //some/other:platform
```

I'm omitting test runner business, but I want something that lets me send individual test logs to a backend for a nicer UX than everything in one file. I'll also gloss over the configuration stuff, that's a separate issue (in short, it would be nice to not run them sequentially, but also some targets in different language that aren't affected by the target incompatibility end up being tested multiple times, which is suboptimal).

So what's relevant here is merging the build and test steps together, which is both nicer and more efficient (fewer DAGs = more goodness overall).

The end goal is to make CI so vanilla that what devs run locally is as close to what runs remotely as possible by default. The differences would be the supertd business, which is ok to me, and something people can opt into; and the custom test runner, which is not relevant locally. I have a personal opinion that it would be a GoodThing for CI systems to be very thin and leverage good build systems as much as possible, so that plays into that as well.

JakobDegen:
Alright, since people internally seem to have surprisingly few opinions on this, I say let's land it without changing the default for now. That being said, I would ask you to switch away from a `--build` flag and instead offer both `--build-run-info` and `--build-default-info`, matching what `buck2 build` has

cbarrete:
Hey, I haven't forgotten about that, but just haven't had the time to finish this up. I hope to get to it before the end of the week though.

app/buck2_client/src/commands/test.rs
```diff
@@ -225,6 +255,8 @@ impl StreamingCommand for TestCommand {
                         .transpose()
                         .context("Invalid `timeout`")?,
                     ignore_tests_attribute: self.ignore_tests_attribute,
+                    build_default_info: !self.skip_default_info,
+                    build_run_info: !self.skip_run_info,
```
JakobDegen, line 259: Isn't this missing a `|| self.build_default_info`?

cbarrete: Not really because the two flags are exclusive by being in the same group, so only checking the opt-in is enough. This is how it is done for the `build` command, so I kept things consistent, but I don't mind being explicit instead if you prefer.

JakobDegen: Right, but the opt-in is `--build-default-info`, no? At least that's what the comment claims

cbarrete: Of course, yes :facepalm:
Sorry for having you double check my logic, I will address this tomorrow.

JakobDegen (approved):

cbarrete:
Hello @JakobDegen, is something blocking this diff internally? Is there something I can do to help move this forward?

cbarrete:
@Nero5023 This PR has been stuck for a few months, do you know if someone/who could look at it?

I have a few more PRs that I'm having trouble getting reviewed: https://github.com/facebook/buck2/pulls/cbarrete

app/buck2_client/src/commands/test.rs
```diff
@@ -157,6 +157,36 @@ If include patterns are present, regardless of whether exclude patterns are pres
...
+    )]
+    build_run_info: bool,
+
+    #[allow(unused)]
+    #[clap(
+        long,
+        group = "run-info",
+        help = "Do not build runtime dependencies (this is the default)"
+    )]
+    skip_run_info: bool,
```
ahornby, line 188: do we need this? seems unused,  not passing build_run_info should keep the default?

app/buck2_client/src/commands/test.rs
```diff
@@ -157,6 +157,36 @@ If include patterns are present, regardless of whether exclude patterns are pres
...
+    )]
+    build_default_info: bool,
+
+    #[allow(unused)]
+    #[clap(
+        long,
+        group = "default-info",
+        help = "Do not build default info (this is the default)"
+    )]
+    skip_default_info: bool,
```
ahornby, line 173: do we need this, seems unused?   not passing build_default_info would already do the default?

app/buck2_client/src/commands/test.rs
cbarrete: This is just keeping symmetry as @JakobDegen asked earlier in this PR. My understanding is that the point is to be able to explicitly opt-in/out in case defaults change.

I don't actually have a strong opinion on UX here btw. In fact, I would personally prefer not needing those flags and having `buck2 test` also build the targets that match the patterns (at least by default), but apparently some people at Meta pushed back against this :shrug:

app/buck2_client/src/commands/test.rs
cbarrete: See other thread.

alexlian:
Apologies this lingered so long. Could I ask for a rebase, and given prior discussion here I'll aim to see this through (unless internally someone ended up forming an opinion)

cbarrete:
Hi @alexlian, the base has completely changed since opening this PR and I'm not interested in looking into this again at the moment.

I"ll close the PR for now to get it off of your list of open PRs, but I know that other people in the community are interested in this, so anyone please feel free to pick this up if you want to.

cbarrete:
@alexlian I've found some free time to redo this after all, but GitHub won't let me reopen this PR, so see https://github.com/facebook/buck2/pull/1044.

PR #728, opened 2024-07-29
Force UTF8 output regardless of the system code page

christianjann:
vswhere will print localized text to the console in localized Windows which cannot be encoded by Python in utf8 and the script fails, we can work around this by passing -utf8 to vswhere 2.5 or newer (https://github.com/microsoft/vswhere/wiki/Encoding)


dtolnay (approved):
Thank you!

PR #739, opened 2024-08-09
bugfix: tweak prelude path for os_lookup

pollend:
I have prelude nested in another directory. this seems to be the only line where this loads from root and not the alias @prelude.


dtolnay (approved):
Thanks!

PR #743, opened 2024-08-19
Add subtarget information to a test target name

cbarrete:
This lets users know which subtarget failed, which is crucial when testing multiple test subtargets (e.g. when there is one per configuration, generated via transitions).

Before:
```
2 TESTS FAILED
  ✗ root//:dummy_test
  ✗ root//:dummy_test
```

After:
```
2 TESTS FAILED
  ✗ root//:dummy_test[gcc13-cxx20-debug]
  ✗ root//:dummy_test[clang18-cxx20-debug]
```

When a default provider is built (no subtargets), the output is unchanged, because `ProviderName`'s `Display` implementation returns an empty string for default providers.

### DIFF
--- a/app/buck2_test/src/translations.rs
+++ b/app/buck2_test/src/translations.rs
@@ -32,7 +32,7 @@ pub(crate) fn build_configured_target_handle(
     let label = target.target().unconfigured();
     let cell = label.pkg().cell_name().to_string();
     let package = label.pkg().cell_relative_path().to_string();
-    let target_name = label.name().to_string();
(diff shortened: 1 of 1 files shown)


cbarrete:
@JakobDegen Those Windows failures are unrelated, so I think that the patch is ready.

cbarrete:
@ndmitchell Could you have a look at this, or alternatively who I should tag as a reviewer?

cbarrete:
@stepancheg maybe? Sorry for insisting, this has just come up again internally.

PR #745, opened 2024-08-21
rust-project: add `--sysroot-mode` to `develop-json`

thoughtpolice:
End-users might want to specify the path to their sysroot in various ways, but by default they probably just want to use whatever is implied by `rustup`, which can be found by `rustc --print=sysroot`

This implements a new `develop-json --sysroot=mode=<MODE>` command, where `<MODE>` might be:

- `rustup` (default): run `rustc --print=sysroot` to figure it out
- `buckconfig`: look in `rust.sysroot_src_path` (or whatever it's called)
- `path:some/path/here`: some random file path to a sysroot
- `cmd:some command here`: a command that is passed to the OS

This should basically allow any needed customization including cases where buck is using a vendored rustc, which may itself refer to a target name and use a command.

### DIFF
--- a/integrations/rust-project/src/cli/develop.rs
+++ b/integrations/rust-project/src/cli/develop.rs
@@ -106,13 +106,26 @@ impl Develop {
         }
 
         if let crate::Command::DevelopJson {
+            sysroot_mode,
             args,
             log_scuba_to_stdout: _,
         } = command
         {
             let out = Output::Stdout;
-            let sysroot = SysrootConfig::BuckConfig;
             let mode = select_mode(None);
+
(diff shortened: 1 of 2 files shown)


davidbarsky:
Thanks! I'll import this as a diff shortly.

thoughtpolice:
I went ahead and renamed the `rustup` option to `rustc` (since there's nothing really rustup specific about it!)

davidbarsky:
Pretty confident that the CI failures are not your fault. I'm going to ignore them when landing this; the main branch has been failing for a minute.

PR #750, opened 2024-08-23
Only upload large files once.

hugwijst:
Currently, upload requests are handled in parallel without knowledge of other ongoing requests. If multiple actions depend on the same set of large locally available artifacts, this results in many upload requests of the same blobs.

To solve this behavior, we store new large blob upload requests in a dashmap. If a request is already ongoing, wait a notification of that upload finishing instead of starting a new upload.

### DIFF
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -19,6 +19,7 @@ use std::time::Instant;
 use anyhow::Context;
 use buck2_re_configuration::Buck2OssReConfiguration;
 use buck2_re_configuration::HttpHeader;
+use dashmap::DashMap;
 use dupe::Dupe;
 use futures::future::BoxFuture;
 use futures::future::Future;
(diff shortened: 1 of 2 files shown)


remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -478,13 +478,21 @@ impl FindMissingCache {
...
+}
+
 pub struct REClient {
     runtime_opts: RERuntimeOpts,
     grpc_clients: GRPCClients,
     capabilities: RECapabilities,
     instance_name: InstanceName,
     // buck2 calls find_missing for same blobs
     find_missing_cache: Mutex<FindMissingCache>,
+    prev_uploads: dashmap::DashMap<TDigest, OngoingUploadStatus>,
```
ndmitchell, line 495: Usual convention would be `use dashmap::DashMap` and then not qualify it onwards

alexlian:
Superceded by 945 which also addresses a crate import. So closing this older one.

PR #751, opened 2024-08-24
Fix recent bugs with shim/third-party

adam-singer:
Observed builds at [TraceMachina/buck2](https://github.com/TraceMachina/buck2/actions/workflows/build-and-test.yml) recently broke due to inconsistencies with shim/third-party directory. 

`reindeer buckify` will not delete existing `Cargo.lock` and `BUCK` if they already exist. `reindeer update` might assume that `reindeer buckify` was done first. A simple and consistent way is to ensure those files are deleted before calling `reindeer` instead of assuming how `reindeer` will behave. 

[serde-rs](https://github.com/serde-rs/json/blob/5ebf65cc480f90714c94f82099ca9161d80cbb10/build.rs#L11) recently introduced a rustc configuration flag that has to be set.`fast_arithmetic` chooses the architecture based on `target_arch`, repeat the same in the fixup scripts.

`zstd` is missing header files, most likely not caught due to stale `Cargo.lock` and `BUCK` files.


shim/third-party/rust/fixups/serde_json/fixups.toml
```diff
@@ -1 +1,9 @@
 buildscript = []
+
+[platform_fixup.'cfg(any(target_arch = "aarch64", target_arch = "loongarch64", target_arch = "mips64", target_arch = "powerpc64", target_arch = "wasm32", target_arch = "x86_64"))']
+buildscript = []
+rustc_flags = ['--cfg=fast_arithmetic="64"']
+
+[platform_fixup.'cfg(not(any(target_arch = "aarch64", target_arch = "loongarch64", target_arch = "mips64", target_arch = "powerpc64", target_arch = "wasm32", target_arch = "x86_64")))']
+buildscript = []
+rustc_flags = ['--cfg=fast_arithmetic="32"']
```
dtolnay, line 9: Please run the upstream build script, instead of transcribing the logic like this.

Example:

https://github.com/facebook/buck2/blob/9beff407c2a67de0b0f28dc900dbf50841d2f14a/shim/third-party/rust/fixups/num-traits/fixups.toml#L1-L2

dtolnay (changes requested):

shim/third-party/rust/fixups/serde_json/fixups.toml
adam-singer: sgtm, I didn't realize thats what the syntax does

adam-singer:
@dtolnay  ptal

dtolnay (approved):
Thanks!

PR #752, opened 2024-08-24
rust-project: name the binary `buck2-rust-project`

thoughtpolice:
The name `rust-project` is glamorous but far too generic to be meaningful when talking about it, and it's not immediately clear what it's even referring to (imagine searching for it on Google). I think `buck2-rust-project` is *much* better and also gives a template for follow up tools like e.g. `buck2-cxx- project`


thoughtpolice:
Going to close this one too, for now. It can be revisited later probably; practically speaking I can just rename the dotslash files in my own repo, so it's not that bad right now...

PR #755, opened 2024-08-27
Fix isdigit / upgrade derive_more 1.0.0 in Cargo.toml

adam-singer:
Seems derive_more upgrade might be a WIP, if not the case here is a PR that can be applied or merged to fix it from the open source / github side. Please feel free to close if this being resolved internally.

### DIFF
--- a/app/buck2_action_impl/BUCK
+++ b/app/buck2_action_impl/BUCK
@@ -11,7 +11,7 @@ rust_library(
         "fbsource//third-party/rust:async-trait",
         "fbsource//third-party/rust:chrono",
         "fbsource//third-party/rust:dashmap",
-        "fbsource//third-party/rust:derive_more-1",
+        "fbsource//third-party/rust:derive_more",
         "fbsource//third-party/rust:either",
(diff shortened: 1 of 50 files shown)


ndmitchell:
Thanks! I screwed a bunch of stuff up here. Will try and unscrew it up!

ndmitchell:
00c906c6f83317629f78d7da3457f45fd356087c and daddd975d12446fba25a2875d37662030eff8db2 were taken from your patch. The idea of the version stripping was to remove the `-1` while we internally upgrade (it should be gone by the end of the week), but also make us immune to future library upgrades. That said, given how many bugs were in that code, I don't think anyone could have guessed its intentions. My only excuse is I thought this was tested in CI, but was hilariously wrong. Thanks for catching all my bugs!

adam-singer:
@ndmitchell sgtm. A plug for https://www.nativelink.com, we are using remote execution for buck2 source on github, thats how we noticed some of these small issues, https://github.com/TraceMachina/buck2/actions/workflows/build-and-test.yml

PR #765, opened 2024-09-03
buck2_execute: implement OSS `upload_blob` for `local_only` cache uploads

thoughtpolice:
Forward-port of patch 4 in <https://github.com/facebook/buck2/pull/477>, providing a clear piece of missing functionality: in the event that stdout or stderr were more than 50KiB of output when caching `local_only` actions, then this dead path was taken, and so stdout/stderr would not be uploaded successfully in the cache.

### DIFF
--- a/app/buck2_execute/src/re/client.rs
+++ b/app/buck2_execute/src/re/client.rs
@@ -303,7 +303,7 @@ impl RemoteExecutionClient {
 
     pub async fn upload_blob(
         &self,
-        blob: Vec<u8>,
+        blob: InlinedBlobWithDigest,
         use_case: RemoteExecutorUseCase,
     ) -> buck2_error::Result<TDigest> {
         self.data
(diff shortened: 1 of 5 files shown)


remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -762,11 +762,21 @@ impl REClient {
 
     pub async fn upload_blob(
```
ahornby, line 763: make it match internal RE signature:
```suggestion
     pub async fn upload_blob_with_digest(
/
        &self,
        blob: Vec<u8>,
        digest: TDigest,
        metadata: RemoteExecutionMetadata,
    ) -> anyhow::Result<TDigest> {
        let blob = InlinedBlobWithDigest{digest:digest.clone(), blob, ..Default::default()};
        self.upload(
            metadata,
            UploadRequest {
                inlined_blobs_with_digest: Some(vec![blob]),
                files_with_digest: None,
                directories: None,
                upload_only_missing: false,
                ..Default::default()
            },
        )
        .await?;
        Ok(digest)
    }
```

app/buck2_execute/src/re/client.rs
```diff
@@ -1230,17 +1230,19 @@ impl RemoteExecutionClientImpl {
 
     pub async fn upload_blob(
```
ahornby, line 1231: updated to match signature used for internal RE call

```suggestion
    pub async fn upload_blob(
        &self,
        blob: InlinedBlobWithDigest,
        use_case: RemoteExecutorUseCase,
    ) -> buck2_error::Result<TDigest> {
        with_error_handler(
            "upload_blob",
            self.get_session_id(),
            self.client()
                .upload_blob_with_digest(blob.blob, blob.digest, use_case.metadata(None))
                .await,
        )
        .await
    }
```

ahornby (commented):
@thoughtpolice, commented with some small changes to make the PR apply cleanly vs the internal RE implementation, please apply, rebase and re-push the PR.

IanChilds:
Apologies for long delay. Looks good, I've commandeered internally to fix some lints and also keep the return value from the client's `upload_blob_with_digest` call, should be able to land soon. Thanks for this!

PR #769, opened 2024-09-04
buck2_validation: tweak and slightly improve error message

thoughtpolice:
Previously, when using `ValidationInfo()` with a failure, it ends up turning out an awkward message that turns

    { "status": "failure", "message": "you failed the test!" }

into

    "you failed the test!".

After this patch, it comes out as:

    you failed the test!

Which seems much more reasonable and easy to parse. It also gives the validation message more control over the punctuation.

### DIFF
--- a/app/buck2_validation/src/cached_validation_result.rs
+++ b/app/buck2_validation/src/cached_validation_result.rs
@@ -31,7 +31,7 @@ pub(crate) enum CachedValidationResultData {
 #[derive(buck2_error::Error, Debug, PartialEq, Allocative, Clone)]
 #[buck2(input)]
 #[error(
-    "Validation for `{target}` failed:\n\n{}.\n\nFull validation result is located at: `{result_path}`", self.rendered_message()
+    "Validation for `{target}` failed:\n\n{}\n\nFull validation result is located at: `{result_path}`", self.rendered_message()
 )]
 pub(crate) struct ValidationFailedUserFacingError {
     target: BaseDeferredKey,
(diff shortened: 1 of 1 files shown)


app/buck2_validation/src/cached_validation_result.rs
```diff
@@ -43,7 +43,7 @@ impl ValidationFailedUserFacingError {
     pub(crate) fn rendered_message(&self) -> Cow<str> {
         self.short_message.as_deref().map_or_else(
             || Cow::Borrowed("Diagnostic message is missing from validation result"),
-            |x| Cow::Owned(format!("\"{}\"", x)),
+            |x| Cow::Owned(format!("{}", x)),
         )
     }
```
blackm00n, line 48: With this change it could be just simplified to return `&str`:

```
pub(crate) fn rendered_message(&self) -> &str {
    self.short_message.as_deref().unwrap_or("Diagnostic message is missing from validation result")
}
```

blackm00n (changes requested):
First of all, thanks for the PR! @thoughtpolice do you mind also fixing the tests since we check the how validation messages are rendered in our e2e suite?

thoughtpolice:
Sorry about leaving this PR behind @blackm00n/@IanChilds! I've just been busy the past few weeks. Glad to see PRs getting merged though!

I wasn't quite sure how to run the full test suite for these changes, either. Is there a guide for the full e2e testsuite? Is it expected to work in OSS builds?

Issue #441, opened 2023-10-09
Sharing and distributing log files/build metadata?


thoughtpolice:
`buck2 log` is pretty cool, since it lets you look at what build commands a user ran and what their output is, and you can do things like `log cmd` or `log replay` to watch the build. Is it possible or recommended to 'share' these log files? What would some infrastructure for doing that look like? Maybe something like:

- Users run `buck2 build ...` a bunch
- Their logs can get synchronized to something like an HTTP endpoint (maybe a proxy that just fronts an S3 bucket)
- Something goes wrong, a user asks for help, posts their Build ID.
- You could run `buck2 log cmd --trace-id $BUILD_ID` to find out what happened.
- This would (transparently?) download the log files from some endpoint (somehow?)

There could be other useful things to derive from this possibly, assuming you had these logs. For example, you could use these logs as a direct source of build analytic information from users to derive information about build times, etc.

Is there any kind of "thing" like this inside Meta? Does this seem interesting? To be truly useful it would need some support in the core executable, I think. It seems like something *kind of* like what `buck2 rage` does, right? Except I'd want it in all cases, not just failures.

I imagine this could look like the following for OSS users:

- A `.buckconfig` key like `buck2_logs.upload_address` is set to `https://buck2-logs.aseipp.dev/`
- All `.pb.zst` files are POST'd to `$UPLOAD_ADDRESS/upload` with their trace ID as a primary key, asynchronously, by the daemon
- When a user says `buck2 log cmd --trace-id $TRACE_ID`, then:
  - HTTP GET `$UPLOAD_ADDRESS/trace/$TRACE_ID`
  - This should return a json object containing metadata and a new address to find the log content at
  - That new address is the public URL to download the `.pb.zst` file
  - This allows users to materialize logs on demand
- Something something something authorization with a bearer token
- Ideally the HTTP protocol is simple enough to implement "by hand" in a short afternoon

thoughtpolice:
Actually, if `--trace-id` could take an `https://` URL and download a file and cache it, you could probably do the first part completely outside of the core executable with a file watching tool? Just watch `buck-out/v2/log` and upload every file that gets written?

cjhopman:
Yeah, I've been really happy with the entire suite of `buck2 log` commands.

> Is there any kind of "thing" like this inside Meta? Does this seem interesting? 

[...]

PR #770, opened 2024-09-05
buck2_client: download traces via `buck2.log_url` config key

thoughtpolice:
When doing a command like `buck2 log replay --trace-id ...` it helps to be able to download previously created logs from users, automated CI, etc. This functionality exists within Meta, as logs are available for download through the "Manifest" system, but it isn't usable in OSS, despite being useful for diagnostics and getting help.

The only missing thing to really get things working is a download mechanism for log files, and a way to know where they should come from.

With this patch, if a user configures the `buck2.log_url` key, which is expected to look something like:

    [buck2]
    log_url = https://example.com

Then, upon executing a `log` command with a `--trace-id` flag, this server will be queried with a:

    GET /v1/get/{uuid}

request, which is expected to return the raw zst-encoded protobuf file. So, buck2 will do that and download the trace, and then use it like normal.

The request is done with `curl`, though in principle it could be done within Buck itself.

This shares as much code as possible with the existing infrastructure and tries to only insert a small key set of `#[cfg(fbcode_build)]` directives. Notably, the path Meta uses for their "Manifold" client is still fully available in the OSS version; `fbcode_build` is only used to gate what the default choice is.

How the server gets access to these files and how they are uploaded is another question. But for now, this can be done several ways outside of buck2 core, so this is good enough.

GitHub Issue: https://github.com/facebook/buck2/issues/441

### DIFF
--- a/app/buck2_common/src/init.rs
+++ b/app/buck2_common/src/init.rs
@@ -294,6 +294,13 @@ impl ResourceControlConfig {
     }
 }
 
+#[derive(Allocative, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub enum LogDownloadMethod {
+    Manifold,
+    Curl(String),
+    None,
+}
+
 /// Configurations that are used at startup by the daemon. Those are actually read by the client,
 /// and passed on to the daemon.
 ///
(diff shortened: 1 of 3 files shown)


thoughtpolice:
This can't do things like pass authentication credentials just yet, but I'd like to know if this is the right shape. As it is, it's still incredibly useful, so I'd like to know if it looks like roughly the right thing before doing much more.

thoughtpolice:
I'd like to make some small API changes to the GET URI so I'm holding this. But I'd still really like something like this to go in :)

app/buck2_common/src/init.rs
```diff
@@ -13,8 +13,8 @@ use std::time::Duration;
 use allocative::Allocative;
 use anyhow::Context;
 use buck2_core::buck2_env;
-use buck2_error::conversion::from_any;
 use buck2_error::BuckErrorContext;
+use buck2_error::conversion::from_any;
```
ahornby, line 17: @thoughtpolice seeing a merge conflict here.  can you rebase once more?

app/buck2_common/src/init.rs
```diff
@@ -13,8 +13,8 @@ use std::time::Duration;
 use allocative::Allocative;
 use anyhow::Context;
 use buck2_core::buck2_env;
-use buck2_error::conversion::from_any_with_tag;
 use buck2_error::BuckErrorContext;
+use buck2_error::conversion::from_any_with_tag;
```
ahornby, line 17: @thoughtpolice  the internal lint format check is complaining about this import reordering, breaks format lint rules.   I can't fix it without taking over the imported diff and thus not being able to approve it, so best if you revert it and push again

PR #764, opened 2024-09-03
remote_execution: implement OSS `write_action_result` for `local_only` cache uploads

thoughtpolice:
Forward port of part of <https://github.com/facebook/buck2/pull/477>, which implements code in the OSS path for putting results of actions marked `local_only = True` into the RBE `ActionCache`.

For people actually using the "remote execution" part of RBE and not just a cache, this rather large omission in the system will often be transparently mitigated because RBE services will not only record `ActionCache` entries on your behalf, they will also implement "EX-to-AC" forwarding where any execution calls will immediately get looked up in the `ActionCache` anyway as an optimization. Which means you'll never see this for things that properly hit CAS inputs/outputs.

However, a `CommandExecutorConfig` configured with `remote_enabled = False` can still enable `remote_cache_enabled = allow_cache_uploads = True` which will both enable the ActionCache/CAS support, and also allow uploads to those interfaces too.

The most useful example of this type of setup is in a CI system like GitHub Actions: where you run the build inside a container or image that provides something quasi-hermetic (e.g. including your toolchains), while the RBE system is purely a cache storing inputs/outputs/ActionCache entries and never uses remote execution.

The hope is that this functionality will soon be plug-and-play on GitHub using an RBE-to-GHA caching proxy, so Buck2 projects will be able to get a quick and easy cache, but without RE. Therefore this functionality will become highly useful.

## Testing

I can submit tests too, though I'm not sure where. But, using the testing example from #477, running `./buck2 clean && BUCK_LOG=buck2_execute_impl=debug ./buck2 build -v3 //:` where `./buck2` is the dotslash build from the [`2024-09-02` tag](https://github.com/facebook/buck2/releases/tag/2024-09-02) gets me:

```
austin@GANON:~/tmp/buck2-test-477$ ./buck2 clean && BUCK_LOG=buck2_execute_impl=debug ./buck2 build -v3 //:
killing buckd server
Buck2 daemon pid 1862754 has exited
/home/austin/tmp/buck2-test-477/buck-out/v2/CACHEDIR.TAG
/home/austin/tmp/buck2-test-477/buck-out/v2/tmp
/home/austin/tmp/buck2-test-477/buck-out/v2/gen
/home/austin/tmp/buck2-test-477/buck-out/v2/log
/home/austin/tmp/buck2-test-477/buck-out/v2/forkserver
/home/austin/.buck/buckd/home/austin/tmp/buck2-test-477/v2
Starting new buck2 daemon...
Connected to new buck2 daemon.
[...]

PR #771, opened 2024-09-09
Add support to upload action results to an RE client.

hugwijst:


thoughtpolice:
Awesome, this looks like an even more thorough version of https://github.com/facebook/buck2/pull/764 which I filed a few days ago, and is a revive of an older PR, which I'm not sure you saw. Did you test it with symlinks? I'm happy to abandon my patch in favor of this once since it looks better at a glance anyway.

thoughtpolice:
See also https://github.com/facebook/buck2/pull/765 which is needed to fix `stdout` uploads in the local caching case.

hugwijst:
> Awesome, this looks like an even more thorough version of #764 which I filed a few days ago, and is a revive of an older PR, which I'm not sure you saw. Did you test it with symlinks? I'm happy to abandon my patch in favor of this once since it looks better at a glance anyway.

I did not see your version, nor did I try symlinks explicitly. I just implemented the converse of `convert_action_result` which was easier than I anticipated. I have some internal use case for which this seems to cache correctly, but don't have explicit tests. 

> EDIT: I'm alternatively happy to port your changes into my patch if you'd like, but frankly I'm just hopeful _some_ version of this patch goes in one way more another, soonish.

I'm happy to go either way too, your PR looks like it has more comments 😅 I too would like to see this feature mainlined soonish.

hugwijst:
> See also #765 which is needed to fix `stdout` uploads in the local caching case.

Oh, interesting. I hadn't run into that yet, but good to know where to look when I do.

hugwijst:
@thoughtpolice, I just found another bug where buck2 ignores the preferred hashing algorithm when checking if it can upload to action cache, instead always using SHA1. See #784 for a fix.

dmezh:
@JakobDegen Gentle bump on this one and https://github.com/facebook/buck2/pull/784

IanChilds:
https://github.com/facebook/buck2/pull/765 was merged. Let us know if you think this one is needed too

PR #775, opened 2024-09-13
Print server side messages, if sent by the RE

vaskomitanov:
If remote execution service sends back a user friendly messages (not part of captured stdout or stderr, and this is not a cached data in CAS), buck2 client should display this information to the end user.
One example of such information is a URL to pointing to Action error of https://github.com/buildbarn/bb-browser.


alexlian:
@vaskomitanov 
Apologies, we've let this lag a little too long and this needs a rebase / resolution of conflicts. If you can get that done, we can aim to see about getting through.

PR #784, opened 2024-09-24
Use preferred digest hashing algorithm for action permission checker.

hugwijst:
To check if buck2 has permissions to upload to action cache, it tries to upload an empty action result. The digest hashing algorithm is hard coded to use SHA1 instead of the preferred hashing algorithm. This commit changes this so the preferred hashing algorithm is used.


alexlian:
Superceded by 946, so will close this one.

PR #787, opened 2024-09-25
Add remote persistent worker support

### LINKED ISSUE #776: feature request - support remote persistent workers
Buck2 has [support for persistent workers](https://github.com/facebook/buck2/commit/d7f58b358984cb6abf316324958aac527dc2fe4f), however, these are only [available for locally executed actions](https://github.com/facebook/buck2/blob/b9c2466565b022633fdc6d90d280709af438d15d/prelude/jvm/cd_jar_creator_util.bzl#L472). In contrast, [Bazel supports remote persistent workers](https://github.com/bazelbuild/proposals/blob/main/designs/2021-03-06-remote-persistent-workers.md), see also [BuildBuddy docs](https://www.buildbuddy.io/docs/rbe-platforms/#remote-persistent-worker-properties).

Persistent workers can provide large performance benefits, the [Bazel documentation reports 2-4x speed-ups for Java](https://bazel.build/remote/persistent), preliminary experiments on our Haskell builds have shown about 3x speed-ups. Without support for persistent workers in remote execution environments, users have to make a trade-off between the speed-up provided by a persistent worker and the speed-up provided by scaling to many build nodes. It would be preferable if both speed-ups could be combined by supporting persistent workers in remote execution environments.

aherrmann:
Closes #776.

Implements support for persistent workers in remote builds using the Bazel remote execution protocol and the approach documented in the Bazel remote persistent workers proposal:
https://github.com/bazelbuild/proposals/blob/main/designs/2021-03-06-remote-persistent-workers.md

Includes an example setup that works with
- local builds without persistent worker
- local builds with persistent worker (Buck2 protocol)
- remote builds without persistent worker
- remote builds with persistent worker (Bazel protocol)

The Bazel remote persistent worker protocol includes an automatic fallback in cases where the remote execution system does not yet support persistent workers. To that end actions take the shape
```
WORKER WORKER_ARGS... @REQUEST_ARGS_FILE
```
The remote execution system separates worker arguments on the command-line from request arguments in the response file and adds the `--persistent_worker` flag. 

The demo worker included in the example in this PR distinguishes between Buck2 worker, Bazel remote worker, and one-shot modes depending on whether Buck2's `WORKER_SOCKET`, Bazel's `--persistent_worker` flag, or neither is set.

The example includes a README with detailed instructions how to test this feature.

- **Add example for remote persistent workers**
- **Implement support for remote persistent workers**

### DIFF
--- a/app/buck2_build_api/src/interpreter/rule_defs/command_executor_config.rs
+++ b/app/buck2_build_api/src/interpreter/rule_defs/command_executor_config.rs
@@ -81,6 +81,7 @@ pub fn register_command_executor_config(builder: &mut GlobalsBuilder) {
     /// * `allow_hybrid_fallbacks_on_failure`: Whether to allow fallbacks when the result is failure (i.e. the command failed on the primary, but the infra worked)
     /// * `use_windows_path_separators`: Whether to use Windows path separators in command line arguments
     /// * `use_persistent workers`: Whether to use persistent workers for local execution if they are available
+    /// * `use_remote_persistent_workers`: Whether to use persistent workers for remote execution if they are available
     /// * `allow_cache_uploads`: Whether to upload local actions to the RE cache
     /// * `max_cache_upload_mebibytes`: Maximum size to upload in cache uploads
     /// * `experimental_low_pass_filter`: Whether to use the experimental low pass filter
@@ -105,6 +106,7 @@ pub fn register_command_executor_config(builder: &mut GlobalsBuilder) {
         #[starlark(default = false, require = named)] allow_hybrid_fallbacks_on_failure: bool,
         #[starlark(default = false, require = named)] use_windows_path_separators: bool,
         #[starlark(default = false, require = named)] use_persistent_workers: bool,
+        #[starlark(default = false, require = named)] use_remote_persistent_workers: bool,
         #[starlark(default = false, require = named)] allow_cache_uploads: bool,
         #[starlark(default = NoneOr::None, require = named)] max_cache_upload_mebibytes: NoneOr<
             i32,
@@ -306,6 +308,7 @@ pub fn register_command_executor_config(builder: &mut GlobalsBuilder) {
                         PathSeparatorKind::Unix
                     },
                     output_paths_behavior,
+                    use_remote_persistent_workers,
                 },
             }
         };
--- a/app/buck2_build_api/src/interpreter/rule_defs/provider/builtin/worker_info.rs
+++ b/app/buck2_build_api/src/interpreter/rule_defs/provider/builtin/worker_info.rs
@@ -45,6 +45,8 @@ pub struct WorkerInfoGen<V: ValueLifetimeless> {
     pub exe: ValueOfUncheckedGeneric<V, FrozenStarlarkCmdArgs>,
     // Maximum number of concurrent commands to execute on a worker instance without queuing
     pub concurrency: ValueOfUncheckedGeneric<V, NoneOr<usize>>,
+    // Remote execution capable worker
+    pub remote: ValueOfUncheckedGeneric<V, bool>,
 
     pub id: u64,
 }
@@ -62,6 +64,7 @@ fn worker_info_creator(globals: &mut GlobalsBuilder) {
         #[starlark(require = named, default = NoneOr::None)] concurrency: NoneOr<
             ValueOf<'v, usize>,
         >,
+        #[starlark(require = named, default = false)] remote: bool,
         eval: &mut Evaluator<'v, '_, '_>,
     ) -> anyhow::Result<WorkerInfo<'v>> {
         let heap = eval.heap();
@@ -72,6 +75,7 @@ fn worker_info_creator(globals: &mut GlobalsBuilder) {
             exe,
             id,
             concurrency: heap.alloc_typed_unchecked(concurrency).cast(),
+            remote: heap.alloc_typed_unchecked(remote).cast(),
         })
     }
 }
@@ -89,6 +93,13 @@ impl<'v, V: ValueLike<'v>> WorkerInfoGen<V> {
             .unpack()
             .expect("validated at construction")
             .into_option()
+    }
+
+    pub fn remote(&self) -> bool {
+        self.remote
+            .to_value()
+            .unpack()
+            .expect("validated at construction")
     }
 }
 
--- a/app/buck2_execute/src/execute/command_executor.rs
+++ b/app/buck2_execute/src/execute/command_executor.rs
@@ -183,6 +183,13 @@ impl CommandExecutor {
                 }
                 CommandExecutionInput::ScratchPath(_) => None,
             });
+            let mut platform = self.0.re_platform.clone();
+            if self.0.options.use_remote_persistent_workers && let Some(worker) = request.worker() && let Some(key) = worker.remote_key.as_ref() {
+                platform.properties.push(RE::Property {
+                    name: "persistentWorkerKey".to_owned(),
+                    value: key.to_string(),
+                });
+            }
             let action = re_create_action(
                 request.all_args_vec(),
                 request.paths().output_paths(),
@@ -191,7 +198,7 @@ impl CommandExecutor {
                 input_digest,
                 action_metadata_blobs,
                 request.timeout(),
-                self.0.re_platform.clone(),
+                platform,
                 false,
                 digest_config,
                 self.0.options.output_paths_behavior,
--- a/examples/persistent_worker/.buckconfig
+++ b/examples/persistent_worker/.buckconfig
@@ -1,0 +1,17 @@
+[cells]
+root = .
+prelude = prelude
+toolchains = toolchains
+none = none
+
+[cell_aliases]
+config = prelude
+fbcode = none
+fbsource = none
+buck = none
+
+[external_cells]
+  prelude = bundled
+
+[parser]
+target_platform_detector_spec = target:root//...->prelude//platforms:default
--- a/examples/persistent_worker/README.md
+++ b/examples/persistent_worker/README.md
@@ -1,0 +1,101 @@
+# Persistent Worker Demo
+
+At the time of writing (2024-09-25) Buck2 supports persistent workers
+for local builds through a dedicated Buck2 persistent worker gRPC
+protocol. However, Buck2 does not support persistent workers for builds
+that use remote execution. This demo is part of a patch-set that adds
+support for remote persistent workers to Buck2, see [#776].
+
+[#776]: https://github.com/facebook/buck2/issues/776
+
+## Requirements
+
+The demo uses Nix to provide required system dependencies, such as
+Python and grpc-io. As well as direnv to set up the environment.
+
+- [Nix](https://nixos.org/)
+- [direnv](https://direnv.net/)
+
+Credentials for [BuildBuddy](https://www.buildbuddy.io/) and
+[GHCR](https://ghcr.io) stored in `.envrc.private`:
+```
+export BUILDBUDDY_API_KEY=...
+export BUILDBUDDY_CONTAINER_USER=...  # GitHub user name
+export BUILDBUDDY_CONTAINER_PASSWORD=...  # GitHub access token
+export GITHUB_USER=... # GitHub user name
+```
+
+Upload the Nix Docker image for remote execution as described in
+`platforms/buildbuddy.bzl`.
+
+## Local Build
+
+Configure a local build without persistent workers:
+```
+$ cd examples/persistent_worker
+$ echo '<file:.buckconfig.no-workers>' > .buckconfig.local
+```
+
+Run a clean build:
+```
+$ buck2 clean; buck2 build : -vstderr
+...
+stderr for root//:demo-7 (demo):
+...
+ONE-SHOT START
+...
+```
+
+## Local Persistent Worker
+
+Configure a local build with persistent workers:
+```
+$ cd examples/persistent_worker
+$ echo '<file:.buckconfig.local-persistent-workers>' > .buckconfig.local
+```
+
+Run a clean build:
+```
+$ buck2 clean; buck2 build : -vstderr
+...
+stderr for root//:demo-7 (demo):
+...
+Buck2 persistent worker ...
+...
+```
+
+## Remote Execution
+
+Configure a remote build without persistent workers:
+```
+$ cd examples/persistent_worker
+$ echo '<file:.buckconfig.buildbuddy>' > .buckconfig.local
+```
+
+Run a clean build:
+```
+$ buck2 clean; buck2 build : -vstderr
+...
+stderr for root//:demo-7 (demo):
+...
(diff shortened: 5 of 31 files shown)


examples/persistent_worker/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+- [Nix](https://nixos.org/)
+- [direnv](https://direnv.net/)
+
+Credentials for [BuildBuddy](https://www.buildbuddy.io/) and
+[GHCR](https://ghcr.io) stored in `.envrc.private`:
+```
+export BUILDBUDDY_API_KEY=...
+export BUILDBUDDY_CONTAINER_USER=...  # GitHub user name
+export BUILDBUDDY_CONTAINER_PASSWORD=...  # GitHub access token
```
sluongng, line 24: I just want to note that this is optional if the container image you are using is publicly downloadable.

examples/persistent_worker/platforms/buildbuddy.bzl
```diff
@@ -0,0 +1,52 @@
...
+            remote_enabled = True,
+            remote_cache_enabled = True,
+            allow_cache_uploads = True,
+            use_limited_hybrid = True,
+            use_persistent_workers = ctx.attrs.use_persistent_workers,
+            use_remote_persistent_workers = ctx.attrs.use_persistent_workers,
+            remote_execution_properties = {
+                "OSFamily": "Linux",
+                "container-image": image,
+                "workload-isolation-type": "podman",
```
sluongng, line 34: Nit: don't need to set isolation type specifically. We(BuildBuddy) may want to change the default isolation type underneath while maintaining backward compatibility. (In fact, we did recently stopped using podman as default isolation).

sluongng (commented):
LGTM from BuildBuddy side.

CI failed due to a Github outage yesterday, but I think a rebase + push should retry it.

examples/persistent_worker/platforms/buildbuddy.bzl
aherrmann: I had to set it this way because I got a credentials error on image download with the default.

christolliday:
Thanks for the PR! This looks good.

We started working on support for this internally but unfortunately it doesn't match the bazel spec exactly. For reasons that aren't entirely clear to me we can't attach the 'worker key' to the platform so it's attached elsewhere on the action.

I can't see why we can't support both APIs though and have a default 'bazel mode' for this behavior, though.

The main other difference is that on our end we construct an RE::Action for the worker, upload it and use the digest of that action as the 'worker key', instead of what I assume is requiring that the worker args are a prefix of the action args (in which case the 'worker key' doesn't really seem to matter?). Similarly we can support both in 'bazel mode' though.

Just a heads up that there is likely to be some churn around this at some point and I'm slightly concerned we may not have an easy time testing that this does the right thing in all edge cases for 'bazel mode', but I suppose we can deal with that when we get to it.

If it's easy, it would be nice to have a github action for testing the remote example. I'm not sure how difficult that is.

aherrmann:
> I can't see why we can't support both APIs though and have a default 'bazel mode' for this behavior, though.

That's great to hear! Yes, I was hoping for something along those lines.

> The main other difference is that on our end we construct an RE::Action for the worker, upload it and use the digest of that action as the 'worker key', instead of what I assume is requiring that the worker args are a prefix of the action args (in which case the 'worker key' doesn't really seem to matter?).

In the Bazel version the worker key is used to associate a given action with a potentially already running worker instance on a remote executor node. But, it is not directly tied to any kind of previously uploaded blob. Bazel calculates a digest of the worker command and its inputs and uses that as a worker key. In this PR I went for the same approach.

> Just a heads up that there is likely to be some churn around this at some point and I'm slightly concerned we may not have an easy time testing that this does the right thing in all edge cases for 'bazel mode', [...] If it's easy, it would be nice to have a github action for testing the remote example. I'm not sure how difficult that is.

That makes sense. I'll look into how to test this on the CI.

aherrmann:
I noticed that the example did not use the `WorkerRunInfo` attributes `worker` and `exe` appropriately to distinguish between (remote) persistent worker and non-worker modes respectively. I've updated the example accordingly. This highlighted that the implementation did wrongly use `request.all_args_vec` for the remote persistent worker case, when it should be composing `worker.exe` and `request.args` instead. I've updated the implementation accordingly.

I will continue looking into ways to test this feature on CI.

aherrmann:
Just to give a heads up on expected progress here. I'm on leave for the next two weeks and will get back to this when I'm back.

I've started making the setup independent on Nix, so that it is easier to integrate with CI here. That's already working locally.
The other thing is the remote execution system to test the remote persistent worker mode. I spoke with BuildBuddy, it would be possible to use BuildBuddy for this CI use-case, but it would require setting up a free account for the Meta Buck2 repository and configuring an access token for CI. @christolliday would that be possible for Meta to do? Otherwise, I'd have to look into other options to set up a compatible remote execution system on CI.

christolliday:
Hi @aherrmann, @KapJI is looking into setting up a build buddy account.

aherrmann:
@christolliday @KapJI I've rebased this PR and added the changes to make it independent of Nix so we no longer require a custom worker image. I've also added a CI test for the persistent worker examples, the test requires a GitHub secret named `BUILDBUDDY_API_KEY` to hold the BuildBuddy token. I've tested the test script locally, but things may still fail under the GitHub actions environment. Please let me know when the BuildBuddy account is set up and a token is added as a GitHub secret, then I can test and debug the CI configuration.

aherrmann:
@KapJI Thank you! That sounds great, I don't think it should required any additional configuration. I'll test and debug the CI setup and let you know if I run into anything.

aherrmann:
One issue I'm encountering is that repository secrets are not exposed to GH actions runs that are initiated from forks (as is the case with this PR) for security reasons ([see here](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/)). Here's what I've done now:
- I've added a CI step for the persistent worker steps.
- In there I check if the token is available or not. If not, I skip the remote execution tests and generate a GH actions notice annotation. So, by default the remote execution cases will only be checked on `main` CI or PRs coming from Meta engineers.
- I've added the `workflow_dispatch` trigger to the CI workflow, this should allow Meta engineers to [manually trigger a CI run](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow) that should have the token set (will only be available once the `workflow_dispatch` trigger is merged). E.g. to test that an external PR doesn't break these tests.

@KapJI could I ask you to trigger a CI run of this PR from within the Buck2 repo to test the remote execution cases? (After convincing yourself that this PR doesn't do anything dodgy with the token).
You can do this by pulling this PR's branch and then pushing it to the `facebook/buck2` repo (not onto the main branch, just as a separate branch). Something along the lines of `gh pr checkout 787; git push origin persistent-remote-worker`. I'd expect the `push` trigger to fire at that point, if not you may need to add a dummy commit.

KapJI:
@aherrmann It's running on https://github.com/facebook/buck2/pull/800

aherrmann:
@KapJI Thanks! Unfortunately, it looks like the token is still not available:
- It is [empty in the `env` section](https://github.com/facebook/buck2/actions/runs/11597402375/job/32290794472?pr=800#step:10:28)
- The [RE tests are skipped](https://github.com/facebook/buck2/actions/runs/11597402375/job/32290794472?pr=800#step:10:103).

Is `BUILDBUDDY_API_KEY` configured as a [repository secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-a-repository)? Does `gh secret list` include it in its output?

aherrmann:
@KapJI friendly ping, did you have a chance to look into [the above](https://github.com/facebook/buck2/pull/787#issuecomment-2449483164)?

KapJI:
@aherrmann Sorry for the late reply. I fixed the secret and now it seems to run: https://github.com/facebook/buck2/actions/runs/12300843378/job/34330007436?pr=800

```
Run ./.github/actions/build_example_persistent_worker
  with:
    buildbuddyApiKey: ***
```

On #800 I also rebased to latest version where anyhow is removed and some code formatting is applied.

aherrmann:
@KapJI Thank you! I've updated this PR to capture the rebase and formatting fixes.

I looked into [the CI failure](https://github.com/facebook/buck2/actions/runs/12300843378/job/34330007436?pr=800#step:10:140) on #800. The issue seems to be that the actions are cached and don't report local or persistent worker execution as expected. I had tested this script with a read-only BuildBuddy token, but it looks like Buck2 CI has a read+write token.

I've addressed this issue by adding a cache buster. An env-var input to the action that is changed each time the test runs. I also  tried the `--no-remote-cache` flag to Buck2, but that had the unwanted side effect of completely omitting the action from the `buck2 log what-ran` output, which the test uses to determine what ran how.

I've tested these changes against a read+write BuildBuddy token as well, and the tests passed. Hopefully they'll pass on Buck2 CI as well now. Please let me know if there is anything else I should address before this PR is ready for merge.

KapJI:
Yes, random cache buster env var is a better choice.

.github/workflows/build-and-test.yml
```diff
@@ -2,6 +2,7 @@ name: Build and test
 on:
   push:
   pull_request:
+  workflow_dispatch: # allows manual triggering
```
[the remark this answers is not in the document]
aherrmann: If an external contributor opens a PR, then CI runs of that PR will not have access to repository secrets such as the BuildBuddy API token. This setting makes it easy for Buck2 maintainers to manually trigger a CI run under the Buck2 repo and facebook org context. I'm not sure which context a re-run would run under. I'm happy to remove this if you feel it's unnecessary.

app/buck2_execute/src/execute/command_executor.rs
```diff
@@ -187,15 +186,45 @@ impl CommandExecutor {
...
+                && let Some(key) = worker.remote_key.as_ref()
+            {
+                platform.properties.push(RE::Property {
+                    name: "persistentWorkerKey".to_owned(),
+                    value: key.to_string(),
+                });
+                // TODO[AH] Ideally, Buck2 could generate an argfile on the fly.
+                for arg in request.args() {
+                    if !(arg.starts_with("@")
+                        || arg.starts_with("-flagfile")
```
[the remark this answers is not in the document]
aherrmann: These are arguments that are passed to the action command and that the remote execution system will extract the arguments from that it will send to the worker process. See [BuildBuddy's implementation](https://github.com/buildbuddy-io/buildbuddy/blob/4bc0fcb58a4106e30edd3a01680d31c1f0baaf3c/enterprise/server/remote_execution/persistentworker/persistentworker.go/#L46), and the [proposal](https://github.com/bazelbuild/proposals/blob/main/designs/2021-03-06-remote-persistent-workers.md#background):
> arguments starting with '@', '--flagfile' or '-flagfile'

examples/persistent_worker/.buckconfig
```diff
@@ -0,0 +1,17 @@
...
+none = none
+
+[cell_aliases]
+config = prelude
+fbcode = none
+fbsource = none
+buck = none
+
+[external_cells]
+  prelude = bundled
```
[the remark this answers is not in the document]
aherrmann: fixed

.github/workflows/build-and-test.yml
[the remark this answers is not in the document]
KapJI: Fair enough, let's keep it.

KapJI:
I'm merging fixes for failing lints right now. Once those are merged, let's rebase and make sure CI is passing.

aherrmann:
@KapJI I've rebased on main.
Linux CI passes.
The MacOS build and test pipeline fails with [the same failure that exists on main](https://github.com/facebook/buck2/actions/runs/12406017689/job/34633806550#step:6:3412). So, that should be unrelated to this PR.
The Windows build examples pipeline fails with also fails with [the same failure that exists on main](https://github.com/facebook/buck2/actions/runs/12406017689/job/34633807204#step:8:6609). So, that should also be unrelated to this PR.

KapJI:
CI on main is passing now, can you please rebase again?

aherrmann:
@KapJI I rebased the PR. I also noticed that the `linux-build-examples` step was failing on the PR because it was using a different Ubuntu version than the main branch pipeline. I've aligned that in https://github.com/facebook/buck2/pull/787/commits/44325ec6a028ec585809a7b1d1730bdadb4f24dc, see commit message for further details.
CI is now green apart from Facebook Internal checks, which I can't see.

avdv:
> I also noticed that the `linux-build-examples` step was failing on the PR because it was using a different Ubuntu version than the main branch pipeline. I've aligned that in [44325ec](https://github.com/facebook/buck2/commit/44325ec6a028ec585809a7b1d1730bdadb4f24dc), see commit message for further details.

@aherrmann Note, this is no longer needed, as this was resolved by https://github.com/facebook/buck2/pull/845

app/buck2_build_api/src/interpreter/rule_defs/provider/builtin/worker_info.rs
```diff
@@ -68,6 +70,7 @@ fn worker_info_creator(globals: &mut GlobalsBuilder) {
             ValueOf<'v, usize>,
         >,
         #[starlark(require = named, default = NoneType)] streaming: Value<'v>,
+        #[starlark(require = named, default = false)] remote: bool,
```
KapJI, line 73: Something like `supports_bazel_remote_execution_protocol`?

app/buck2_build_api/src/interpreter/rule_defs/command_executor_config.rs
```diff
@@ -111,6 +112,7 @@ pub fn register_command_executor_config(builder: &mut GlobalsBuilder) {
         #[starlark(default = false, require = named)] allow_hybrid_fallbacks_on_failure: bool,
         #[starlark(default = false, require = named)] use_windows_path_separators: bool,
         #[starlark(default = false, require = named)] use_persistent_workers: bool,
+        #[starlark(default = false, require = named)] use_remote_persistent_workers: bool,
```
KapJI, line 115: Something like `use_bazel_protocol_remote_persistent_workers`?

KapJI (commented):
@aherrmann  There is some feedback from internal review. Can you address it and come up with better names?

> This looks good, the only thing that would be helpful is if the naming made clear that we are talking about Bazel's remote persistent worker protocol, since we're probably going to have to have separate support for our internal RE persistent workers.
I'll stick a couple of suggestions inline, but honestly they could probably be better

app/buck2_build_api/src/interpreter/rule_defs/command_executor_config.rs
aherrmann: Updated

aherrmann:
@KapJI I've implemented the suggested change. I've tweaked the name slightly. I've also rebased the PR again.

PR #791, opened 2024-10-10
chore: `nix flake update`

thoughtpolice:
Regularly scheduled update so that Nix users (like me) can build more recent copies of Buck2.


thoughtpolice:
Apparently this was fixed by https://github.com/facebook/buck2/pull/795

PR #795, opened 2024-10-18
Fix nix Builds

abhillman:
Resolves https://github.com/facebook/buck2/issues/794. Changes via `nix flake update`.


abhillman:
cc/ @JakobDegen

PR #811, opened 2024-11-15
[WIP] Bazel Build Event Stream support

aherrmann:
This PR adds support to connect Buck2 to a [Bazel Build Event Stream server](https://bazel.build/remote/bep#build-event-service) as found among some Bazel remote build execution platforms. It repurposes the existing RemoteEventSink used to connect to Scribe within Meta's internal Buck2 version and connects it to a Bazel Build Event Service over gRPC instead, and it adds a translation layer between Buck2 build events and Bazel build events.

This is a work in progress, only very few build events are translated so far. To test this code you can
- build Buck2,
- change into the `examples/hello_world` directory,
- configure a BES service by setting the `BES_URI` environment-variable, and
- start a build.

```bash
$ cargo build --bin=buck2
$ cd examples/hello_world
$ BES_URI=XYZ ../../target/debug/buck2 build :main
```

For example, with BuildBuddy, if you run `bazel run server -- -app.log_level debug` locally, set `BES_URI=http://localhost:1985`.

The Buck2 Trace-ID is used as the BES invocation ID and you can see a Buck2 build on the UI under e.g. `http://localhost:8080/invocation/e3123cfa-c350-432b-9784-8b3fe48bed14`.

- **BuckEvent: provide a Buck Event Publisher proto**
- **Add buck2_event_publisher_proto to Cargo.toml**
- **Add Bazel Build Event Protocol/Stream protobuf**
- **Implement RemoteEventSink to publish to Bazel BES**


nlopezgi:
Hi @aherrmann , thanks for putting this together, we are really looking forward to it. I tested it out and had a couple of questions:

- I see the TLS bit is marked on a TODO as not enabled yet, are you planning on setting this up as part of this CL or a future one?
- Is there any way that I can get some verbose logs from the service publishing BuildEvents? I did not find a way to see anything from this interaction.

Thanks again!

aherrmann:
> I see the TLS bit is marked on a TODO as not enabled yet, are you planning on setting this up as part of this CL or a future one?

It's needed for the target use-case. So, I'm planning to add support for it.

> Is there any way that I can get some verbose logs from the service publishing BuildEvents? I did not find a way to see anything from this interaction.

The BES client runs as a thread in the Buck2 daemon. Currently, there is just [a bit of printf-debugging](https://github.com/facebook/buck2/pull/811/commits/81b1763306b18b6ff27f470eb2a0daa8efa0b778#diff-7628cbf94874c46665052fb22b36a1affd20bcca0d2d3737b9a99d6da6750d95R406) in the code. Buck2 has some logging facilities, see `--verbose` flag and `BUCK_LOG` env-var. However, this PR so far doesn't make explicit use of these.

aherrmann:
> Do you have plans to add the `PublishLifecycleEvent` calls?

I'm currently testing and developing against BuildBuddy. To my understanding so far, that implementation seems to largely ignore lifecycle events, at least I haven't spotted a need for them, yet. So, for now, I'm not planning to add lifecycle events.

nlopezgi:
> > Do you have plans to add the `PublishLifecycleEvent` calls?
> 
> I'm currently testing and developing against BuildBuddy. To my understanding so far, that implementation seems to largely ignore lifecycle events, at least I haven't spotted a need for them, yet. (I may be wrong, I'm not deeply familiar with the BuildBuddy implementation). So, for now, I'm not planning to add lifecycle events.

Thanks @aherrmann, I think your current process of develping against BuildBuddy makes sense as an initial approach to validate your POC. 

I confirmed with some colleagues and our reading of the protocol is that there should be some lifecycle events in order to produce a legal event stream. Particularly The protocol mandates that `BuildEnqueued` must be sent before any other event, and this event seems to be something that is documented in the [PublishLifecycle](https://github.com/googleapis/googleapis/blob/4fefaf151c721c138588efeba0e99591d51cbc16/google/devtools/build/v1/publish_build_event.proto#L54C20-L54C43) event docs as one that is needed. 

I'm happy to talk more about this if you feel strongly that lifecycle events are really not needed and ask some of my colleagues with more experience on the protocol to chime in too. 

Thanks again for your work on this feature!

nlopezgi:
Circling back on the state of testing of this PoC with EngFlow UI. I have been able to test all the features on this PR successfully. The key things I had to hardcode to get it to work properly with EngFlow UI are the following:
- TLS cert creation
- Set project in requests to value "default"
- Add lifecycle events called via rpc `PublishLifecycleEvent` for the following: `BuildEnqueued`, `InvocationAttemptStarted`, `InvocationAttemptFinished`, `BuildFinished`. 
Thanks again for putting this PoC together for testing @aherrmann . I will keep my eye out if you add more features to test them with EngFlow UI.

aherrmann:
I've added sufficient TLS and HTTP header support to connect to the BuildBuddy cloud service over https and using an access token.

app/buck2_events/src/sink/remote.rs
```diff
@@ -334,23 +334,618 @@ mod fbcode {
...
+        };
+        let interceptor = InjectHeadersInterceptor::new(&headers)?;
+        let client = PublishBuildEventClient::with_interceptor(endpoint, interceptor);
+        Ok(client)
+    }
+
+    fn buck_to_bazel_events<S: Stream<Item = BuckEvent>>(events: S) -> impl Stream<Item = v1::BuildEvent> {
+        let mut target_actions: HashMap<(String, String), Vec<(BuildEventId, bool)>> = HashMap::new();
+        stream! {
+            for await event in events {
```
cjhopman, line 533: you should be able to use unpack_event to get an UnpackedBuckEvent that cuts down a bit on the boilerplate here (compare superconsole's matching on these events: 

https://github.com/facebook/buck2/blob/86174c3e5fddc2bf0e583dc7ed39984bfd61891e/app/buck2_client_ctx/src/subscribers/superconsole.rs#L504

PR #814, opened 2024-11-26
github: disallow libc 0.2.165

KapJI:
Summary:
`sysinfo` is fixed by https://github.com/GuillaumeGomez/sysinfo/pull/1393 which is released as v0.32.1.

But there are breaking changes compare to current sysinfo v0.30.11 we have internally and it needs to be migrated.

Differential Revision: D66499693


tgross35:
Fyi 0.2.165 was yanked, 0.2.166 was released with a fix

PR #836, opened 2025-01-09
Improved extract_archive and http_archive

cormacrelf:
This is to support https://github.com/facebookincubator/reindeer/pull/46, attn @dtolnay probably

Two new features:

- `extract_archive` now shares its unarchive implementation (and available attrs) with `http_archive`, which was a lot more mature. So it gets `strip_prefix` and `sub_targets` etc. And windows support.
- Both of them now allow a dict for `sub_targets`. This wasn't needed for that reindeer PR, but I did use it for a golang version of that reindeer vendoring strategy. Think `{ "embed_files": [ "default.html", "default.css" ] }`.


dtolnay (approved):
Thanks!

PR #844, opened 2025-01-16
Allow colon in HTTP header value

avdv:
IMO, there is no reason to disallow a colon in HTTP header values, since any VCHAR (US ASCII visible character) is permitted by rfc7230.

This is useful in case one needs to send a URL in a header for example.


avdv:
Note, the https://github.com/facebook/buck2/actions/runs/12810432845/job/35717481138?pr=844 CI job failed because it is referring to `ubuntu-latest` and this was switched to the ubuntu-24.04 runner. See https://github.com/actions/runner-images/issues/10636

Switching back to ubuntu-22.04.

PR #856, opened 2025-02-07
Add console os constraints and config_settings.

zjturner:


zjturner:
I am not sure where else it could really go and be usable.  It would be used anywhere that a `config//os:` constraint is currently used.  There's too many to list, but just as one example, this is a frequent use case for us:

```
cxx_library(
    name = "foo",
    srcs = [
        <common sources>
    ] + select({
        "config//os:windows": _windows_sources,
        "config//os:linux": _linux_sources,
        "config//os:macos": _macos_sources,
        "config//os:ps4": _playstation_sources,   # we need this
    })
)
```

Same argument applies to deps, and most other selectable args.

When deciding whether it made sense to put these here, I looked at the surrounding code in the file.  There are other constraints like `illumos`, `netbsd`, and `fuchsia` that if you grep the prelude for them, 0 hits come up.  And these are prefaced by a comment that says "Long tail but contemporarily relevant operating systems".  So it seems appropriate to me for these to be here, but LMK your thoughts.

zjturner:
How would you suggest I accomplish selecting on other operating systems if not by adding a constraint to the operating system setting?  Isn't that what it's for?  I could do something like:

```
select({
    "internal//os_visibility:private": {
        "internal//os:ps4": [],
        "internal//os:ps5": [],
    },
    "internal//os_visibility:oss": {
        "config//os:windows": [],
        "config//os:macos": [],
    }
})
```

but that seems quite extreme given that that the constraints I'm adding actually are operating systems, that conceptually and logically do belong in a list of operating systems.  If I have to go that route, I'll probably instead just fork the prelude internally.  But I think it's a reasonable ask to get these into prelude since this is where they make logical sense, and if Meta were to ever support these operating systems, I have a strong suspicion this is exactly where the constraints would go.

dtolnay:
I think there are some misunderstandings in the previous comment. You can just add:

```starlark
constraint_value(
    name = "ps4",
    constraint_setting = "prelude//os/constraints:os",
)
```

anywhere in your own codebase and that becomes a valid value of OS. If the prelude isn't going to be using this in any of the prelude's toolchain definitions, there is no technical difference between putting this in prelude or elsewhere. There is no need for "os_visibility" either way.

```starlark
select({
    "prelude//os:windows": ...,
    "//path/to:ps4": ...,
})
```

zjturner:
> I think there are some misunderstandings in the previous comment. You can just add:
> 
> ```python
> constraint_value(
>     name = "ps4",
>     constraint_setting = "prelude//os/constraints:os",
> )
> ```
> 
> anywhere in your own codebase and that becomes a valid value of OS. If the prelude isn't going to be using this in any of the prelude's toolchain definitions, there is no technical difference between putting this in prelude or elsewhere. There is no need for "os_visibility" either way.
> 
> ```python
> select({
>     "prelude//os:windows": ...,
>     "//path/to:ps4": ...,
> })
> ```

I didn't realize that would work.  If that does in fact work then I agree there is no need for this.  I didn't think you could select on constraint_values that were namespaced in different folders.  I'll give this is a try, and if this does work then I'll close this issue.

alexlian:
@zjturner Checking if there's further needs here? Or did you get the constraint_value working in your own repo?

zjturner:
I didn't have a chance to try it out.  But I believe David's response and think it probably works.  So how about we close this out for now, and if I find out it doesn't work (unlikely) I'll re-open the issue

PR #857, opened 2025-02-07
Add documentation on toolchains

### LINKED ISSUE #666: Documentation on toolchains is lacking
Toolchains are a central concept in Buck2, since they are supported in the core rather than just being something defined in the prelude, but they are essentially not documented: it is not listed under Concepts in the sidebar, not part of the Glossary of Terms, and there is little information about what a toolchain really is or does.

Some questions that would be worth answering:
- What does Buck2 call a toolchain and what is it used for?
- What purpose does `rule`'s `is_toolchain_rule` serve?
- When should one write a rule versus a toolchain?
- How can/should one use an existing toolchain?
- How do toolchains relate to configuration/constraints?

I have a vague notion of "toolchains let you decouple a rule from the actual compilers, runtimes, sysroots, etc. that it might use", but that's not enough for me to write good documentation about them I'm afraid...

cbarrete:
Closes #666.

### DIFF
--- a/docs/rule_authors/writing_toolchains.md
+++ b/docs/rule_authors/writing_toolchains.md
@@ -1,0 +1,155 @@
+---
+id: writing_toolchains
+title: Writing Toolchains
+---
+
+Toolchains are regular rules that:
+
+- Have `is_toolchain_rule = True` passed to the `rule` call.
+- Return the provider(s) expected by rules using that toolchain,
+  conventionally named `*ToolchainInfo`.
+
+Toolchain rules are instantiated once in the `toolchains//` cell. The
+location of the `toolchains` cell is determined by the value of
+`cells.toolchains` in the `.buckconfig` file.
+
+Regular build rules reference those toolchain targets as
+[`toolchain_dep`](https://buck2.build/docs/api/build/attrs/#toolchain_dep)
+attrs (often
+[`default_only`](https://buck2.build/docs/api/build/attrs/#default_only)
+ones). See
+<https://buck2.build/docs/rule_authors/configurations/#toolchain-deps>
+for more information about toolchain deps.
+
+## Writing a custom toolchain
+
+The prelude exposes a few demo toolchains with specific configurations
+(e.g. hardcoded compiler and linker flags), that expect to find the
+tools on the `PATH`. Many users will want more control over those
+toolchains. Several options are available:
+
+- Defining a custom toolchain for a language that is supported in the
+  prelude. One can then either:
+    - Return the `*ToolchainInfo` provider defined in the prelude, to
+      get compatibility with all the prelude target rules for free.
+    - Return a custom toolchain provider, for use with a custom set of
+      target rules (which is a lot more effort).
+- Defining a toolchain for a custom language/process. One then has to
+  define a toolchain provider for it, which will be used by the rules
+  for that language.
+
+### Writing a prelude-compatible toolchain
+
+People will often first encounter toolchains when they want to switch
+off of the demo toolchains that `buck2 init` uses by default. For
+example, one might want to tweak which compiler is used, which flags are
+passed to it, or where it is fetched from.
+
+The most straightforward thing to do first is to instantiate one of the
+"system" toolchain rules that the prelude offers. Studying the
+toolchains defined by the `system_demo_toolchains` macro in
+`@prelude//toolchains:demo.bzl` is a good way to get started.
+
+For example, here is how one could define a C++ toolchain that builds
+projects in C++23, with warnings as errors and optimizations enabled:
+
+```python
+load("@prelude//toolchains:cxx.bzl", "system_cxx_toolchain")
+
+system_cxx_toolchain(
+    name = "cxx",
+    compiler_type = "clang",
+    cxx_flags = [
+        "-std=c++23",
+        "-Wall",
+        "-Wextra",
+        "-Werror",
+        "-O3",
+    ],
+    visibility = ["PUBLIC"],
+)
+```
+
+One would typically use
+[`select`](https://buck2.build/docs/rule_authors/configurations_by_example/)
+to customize the toolchain e.g. based on the build mode (debug vs
+release) or compiler type.
+
+Note that several toolchains require you to also define a
+`python_bootstrap` toolchain. This is because those toolchains use
+Python scripts (e.g. to generate compilation databases for C++), and the
+prelude gives you control over the Python toolchain used for this
+purpose. The bootstrap Python toolchain can be different from the
+toolchain used by `python_*` rules. `@prelude//toolchains:python.bzl`
+provides a simple `system_python_bootstrap_toolchain` in case you do not
+care about tightly controlling the Python used when building other
+languages.
+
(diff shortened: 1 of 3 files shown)


cbarrete:
@alexlian I hope you don't mind poking you here as well. Do you also know someone who might be able to look at this at some point?
Thanks

docs/rule_authors/writing_toolchains.md
```diff
@@ -0,0 +1,130 @@
...
+all that is needed is to add that toolchain as a `toolchain_dep` attr to
+a build rule. Let's consider the following made up rule that would
+simply call a compiler on a source file:
+
+```python
+# We assume that `FooToolchainInfo` is a struct with a `compiler` field.
+def _foo_binary_impl(ctx: AnalysisContext) -> list[Provider]:
+    output = ctx.actions.declare_output(ctx.label.name)
+    ctx.actions.run(
+        [ctx.attrs._compiler, ctx.attrs.source, "-o", output.as_output()],
```
mloc, line 98: Is `ctx.attrs._compiler` correct here? I'd have expected something like `ctx.attrs._foo_toolchain[FooToolchainInfo].compiler`.

cbarrete: Oh right, I must have carelessly edited this right before pushing :facepalm: 
I'll fix this when I have a moment, thanks for catching it.

docs/rule_authors/writing_toolchains.md
```diff
@@ -0,0 +1,130 @@
...
+Python scripts (e.g. to generate compilation databases for C++), and the
+prelude gives you control over the Python toolchain used for this
+purpose. The bootstrap Python toolchain can be different from the
+toolchain used by `python_*` rules. `@prelude//toolchains:python.bzl`
+provides a simple `system_python_bootstrap_toolchain` in case you do not
+care about tightly controlling the Python used when building other
+languages.
+
+Completely custom toolchains can also easily be written as rules with
+`is_toolchain_rule = True` which return a `*ToolchainInfo` struct. The
```
Ralith, line 72: "completely custom" (e.g. not a language/process known to the prelude) seems to be at odds with "use one of the pre-defined `*ToolchainInfo` structs". What if I want a toolchain for something the prelude doesn't have a pre-defined helper for? Clarify and contrast with e.g. `RunInfo`?

docs/rule_authors/writing_toolchains.md
```diff
@@ -0,0 +1,130 @@
...
+## Accessing a toolchain in a build rule implementation
+
+Before going any further, note that using a toolchain is often
+unnecessary. It is perfectly possible for a build rule to reference
+regular targets as dependencies, without the added complexity of
+defining a toolchain for them. For example, one can add an (often
+`default_only`) `attrs.exec_dep` to a build rule's `attrs` to do code
+generation in a custom rule without creating a toolchain for it, as long
+as there is no need to decouple the code generator being used from the
+rule implementation.
```
Ralith, line 86: > as long as there is no need to decouple the code generator being used from the rule implementation.

Is that precisely when defining a toolchain is needed? Could that be stated a little more directly, perhaps in the introduction?

docs/rule_authors/writing_toolchains.md
```diff
@@ -0,0 +1,130 @@
+---
+id: writing_toolchains
+title: Writing Toolchains
+---
+
+Toolchains are regular rules that:
+
+- Have `is_toolchain_rule = True` passed to the `rule` call.
+- Return a toolchain provider, which is just a struct with a name like
+  `*ToolchainInfo`.
```
Ralith, line 10: What's special about these providers compared to others? I seem to be able to use `RunInfo` to define toolchains.

docs/rule_authors/writing_toolchains.md
```diff
@@ -0,0 +1,130 @@
+---
+id: writing_toolchains
+title: Writing Toolchains
+---
+
+Toolchains are regular rules that:
+
+- Have `is_toolchain_rule = True` passed to the `rule` call.
```
Ralith, line 8: What does this actually do? I haven't been able to find any discussion.

Ralith (commented):
Some feedback based on my recent/ongoing struggles to understand how any of this fits together.

docs/rule_authors/writing_toolchains.md
cbarrete: [I don't know either](https://github.com/facebook/buck2/issues/666)! I haven't gotten any answers from Meta, but no luck so far.
I've seen that it's actually used in the Rust code, but haven't taken the time to dig deeper.

docs/rule_authors/writing_toolchains.md
cbarrete: Good point, I think it's worth clarifying a few different use cases:

- Defining a custom toolchain for a language that is supported in the prelude, in which case one could either:
    - Return the `*ToolchainInfo` provider defined in the prelude, to get all the prelude target rules for free.
    - Return a custom toolchain provider, for use with a custom set of target rules (which is a lot more effort).
- Defining a toolchain for a custom language/process, in which case one would have to define a toolchain provider for it

docs/rule_authors/writing_toolchains.md
cbarrete: I have the following in the first paragraph of the intro:

> They allow the reuse of rules across projects that source their tools (e.g. compilers and linters) differently.

This might not be explicit enough? Let me know if this could be better rephrased.

docs/rule_authors/writing_toolchains.md
cbarrete: I don't think there's anything special about them (someone may correct me though), the naming is just a convention.

A `RunInfo` would only work if you had a single command to run as part of your toolchain though, right? If that's the case, I guess it would work just fine, but this might be a bit confusing to rule authors, that would expect toolchains to be defined by providers that follow this convention?

docs/rule_authors/writing_toolchains.md
Ralith: I noticed that using a toolchain target as a non-toolchain dependency produces an error. Perhaps that's the sum total effect?

docs/rule_authors/writing_toolchains.md
Ralith: I think there's an important difference between "toolchains allow ..." and "the sole purpose of toolchains is ...".

cbarrete:
@alexlian I appreciate it! I've been a little short on time lately, but I'll address the comments whenever I have a moment, then this should be good to go.

docs/rule_authors/writing_toolchains.md
cbarrete: I don't think there's any action item here. Can we resolve this?
I suggest continuing the conversation somewhere else if you're still interested.

cbarrete:
This is ready on my side @alexlian :)

docs/rule_authors/writing_toolchains.md
Ralith: I think this would be a great place to document what this option does, but I guess it can be left for future work.

Issue #868, opened 2025-03-06
Single file compilation doesn't work with vsgo


zjturner:
When you select a single file and hit Ctrl+F7, Visual Studio _appears_ to compile the one file you selected, as it should . But if you enable diagnostic MSBuild Verbosity, you'll see that it's actually using its default ClCompile behavior of invoking the system compiler, and it's not going through buck2.  There should be some way to make this invoke the command `buck2 build foo//:target[objects][src.cpp.obj]`.

zjturner:
@autozimu

zjturner:
I'm looking into this a little on my own.  It seems you can add something like:

```
<Target Name="ClCompile">
    <Build Exec="buck2 build $(buck2_target_name)[objets][foo.cpp.obj]"/>
</Target>
```

But I'm trying to figure out the right way to communicate `foo.cpp.obj` from the `ClCompile` item in the vcxproj to the task.  Which I don't have a solution for yet, but I think I'm close.

autozimu:
This might be helpful to override the target with arguments https://stackoverflow.com/questions/45835043/how-to-override-compile-command-of-visual-studio-2017-community

zjturner:
Yea this is similar to what I'm doing.  The only caveat is that I need not the name of the source file, but the name of the subtarget.  So even if the file is called `foo.cpp`, if you specified as `cxx_library(srcs = ["src/foo/foo.cpp"])` then you need the task's exec command to pass `target[objects][src/foo/foo.cpp.obj]`.  I am trying something like this:

```
<ClCompile Item="Absolute/Path/To/Foo.cpp">
    <BuckSubTarget>src/foo/foo.cpp.obj</BuckSubTarget>
</ClCompile>
...
<Target Name="ClCompile">
    <Build Exec="buck2 build $(buck2_target_name)[objects][@(ClCompile.BuckSubTarget)]"/>
</Target>
```

but it doesn't like this, it's always empty.  I'll figure it out sooner or later :)

zjturner:
Can someone help me figure this out?  I need to be able to translate the source artifact into the path of the object file.  For example, `src/Foo.cpp` -> `src/Foo.cpp.obj`.  I can't just get the source artifact and stick `.obj` onto it, because it depends on the file extension.  It might be `.o`, or something else.  It's determind by the toolchain.  Also it might have `.pic`.  This information isn't even *in* the attrs as far as I can tell.

[...]

PR #871, opened 2025-03-11
Properly support single file compilation in vsgo

zjturner:
Presently, if you select a file in Solution Explorer and hit Ctrl+F7, it uses the default `ClCompile` task, which invokes the system-installed version of cl.exe corresponding to whatever version of Visual Studio you have installed.  It does not go through buck2 or invoke `buck2 build` in any way.  To fix this, we need to do things:

1. We need to associate the object file sub-target name with each source file so this can be passed to the buck2 build command.
2. We need to replace the `ClCompile` task with a custom one that invokes `Exec` to call buck2.

The first is done by Adding a piece of metadata to each `ClCompile` item.  So for example we need to generate something like this for each cpp file:

```
        <ClCompile Include="$(RepoRoot)\A\B\C\Foo\Foo.cpp">
            <BuckObjSubTarget>Foo/Foo.cpp.obj</BuckObjSubTarget>
        </ClCompile>
```

Second, we need to inject a replacement for the `ClCompile` task into every vcxproj.  This means putting the following harcoded snippet of MSBuild after all of the `ClCompile` items.

```
    <Target Name="ClCompile" Condition="'@(ClCompile)' != ''" DependsOnTargets="SelectClCompile">
        <Exec WorkingDirectory="$(RepoRoot)" Command="$(vs_buck_build_command)[objects][%(ClCompile.BuckObjSubTarget)] $(ExtraBuckOptions)"/>
    </Target>
```

One might wonder why this doesn't go inside of `vs_buck_build.props` like the other task replacements.  I was not able to make this work because of the `%(ClCompile.BuckObjSubTarget)` reference.  When I put the task inside of the props file, this would not resolve to anything, so the only way I could get it to work was to put it inside of every vcxproj.

After this PR, hitting Ctrl+F7 on a single file will invoke `buck2 build <target_name>[objects][foo.cpp.obj]`

### DIFF
--- a/prelude/ide_integrations/visual_studio/gen_filters.bxl
+++ b/prelude/ide_integrations/visual_studio/gen_filters.bxl
@@ -29,7 +29,7 @@ def _get_common_directory(paths) -> str:
 
 def _get_file_filter_map(vs_settings: dict) -> dict:
     filters = {}
-    files = vs_settings["Headers"] + vs_settings["Sources"]
+    files = vs_settings["Headers"] + vs_settings["Sources"].keys()
 
     # Handle paths that start with "buck-out"(exported files) separately since these file
     # don't have any common dir with the others and would make the hierarchy really nested
--- a/prelude/ide_integrations/visual_studio/gen_vcxproj.bxl
+++ b/prelude/ide_integrations/visual_studio/gen_vcxproj.bxl
@@ -49,12 +49,13 @@ def _gen_headers(headers):
     ], indent_level = 1)
 
 def _gen_sources(sources):
-    return h("ItemGroup", [
-        h("ClCompile", None, {
+    source_entries = [
+        h("ClCompile", [h("BuckObjSubTarget", object, indent_level=3)], {
             "Include": cmd_args("$(RepoRoot)\\", source, delimiter = ""),
         }, indent_level = 2)
-        for source in sources
-    ], indent_level = 1)
+        for source, object in sources.items()
+    ]
+    return h("ItemGroup", source_entries, indent_level = 1)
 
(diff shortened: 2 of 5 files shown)


autozimu:
Relevant discussion at #868

autozimu:
LGTM. I will land from internal Phabricator after tests run (and also I don't have permission to land here at GitHub).

autozimu:
PR landed in.

Note though, I have to change extension from `.obj` to `.o`, otherwise buck throws subtarget not found error. Let me know if this doesn't work for your case and we can find common path to proceed.

Besides, there is just minor changes to make our internal linters happy.

zjturner:
Hmm, yea `.o` will not work for us.  I'm curious, is this for windows targets?  In your msvc toolchain file, do you have this?

```
        CxxToolchainInfo(
            linker_info = LinkerInfo(
                object_file_extension = "o",
            )
```
?  For us we have it set to "obj".  I suppose it doesn't really matter, and we can probably just change our toolchain file to set the object file extension to `.o`.  I don't think any harcdoded requirement in the toolchain requires it to be `obj`, it's just the "canonical" naming on windows.

zjturner:
The extension comes from your toolchain file.  You can set it to whatever you want.  Let me try setting it to o in our toolchain file, if it works and we don't run into any issues, we'll just change it to o on our end so we don't have to do anything.

I think in the future you could probably improve things by scanning the dep tree for the toolchain target, running analysis on it, and getting the CxxLinkerInfo provider out of it.  Too much work for now though if we can just change it on our side to be `o`.

autozimu:
Yep, just located it. We internally set `.o` for both MSVC and clang toolchains on Windows.

PR #884, opened 2025-03-20
prelude: fix zig cxx toolchain (fixes #883)

### LINKED ISSUE #883: Can't build with cxx_zig_toolchain: Type Error in cxx/zig toolchain definition
I tried to use a zig toolchain to build an example cxx_binary project, and ran into the following error:
```
BUILD FAILED
From load at BUCK:1

Caused by:
    0: Error evaluating module: `prelude//toolchains/cxx/zig/defs.bzl`
    1: error: Expected type `LinkerType` but got `str`
          --> prelude/toolchains/cxx/zig/defs.bzl:402:49
           |
       402 |             is_pdb_generated = is_pdb_generated(_get_linker_type(dist.os), ctx.attrs.linker_flags),
           |                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^
           |
       

```
Looking at that file, and the `is_pdb_generated` definition, it is an obvious bug. `is_pdb_generated` takes its first argument as a type `LinkerType`, whereas `_get_linker_type` returns a `str`. The `LinkerType` is just a wrapper/newtype for `str`, so the fix is easy. PR is incoming :)

csarn:
- fix type error in prelude/toolchains/cxx/zig/defs.bzl (fixes #883 )
- update zig toolchain versions

I hope it's ok to do two (related) things in one PR, otherwise I can split it up.

### DIFF
--- a/prelude/toolchains/cxx/zig/defs.bzl
+++ b/prelude/toolchains/cxx/zig/defs.bzl
@@ -102,6 +102,7 @@ load(
     "CxxCompilerInfo",
     "CxxInternalTools",
     "LinkerInfo",
+    "LinkerType",
     "ShlibInterfacesMode",
     "StripFlagsInfo",
     "cxx_toolchain_infos",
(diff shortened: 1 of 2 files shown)


alexlian:
@csarn So, I think it would be better to separate the PR into the fix and the toolchain updates as I can more readily push through that change.

alexlian:
@csarn Scratch that - someone is taking a look as-is.

PR #888, opened 2025-04-01
extract SimpleDirEntry into dice/dice so that we can use it in UserDataComputation

AishwaryaSivaraman:
Summary:
We want to do  per-computation cache of readdirs so that when we introduce a new key for tracking packages we dont call read_dir twice.

this diff is preparatory work for the next one where we had to move around some definitions to avoid cycles

Differential Revision: D72246720


alexlian:
Closing, as I believe we took an alternative approach so this isn't needed.

PR #896, opened 2025-04-09
Allow rustc_flags take Artifact as argument

yxdai-nju:
When using Buck2 to bootstrap the Rust compiler, the ability to create a sysroot directory is needed (i.e., to collect rustc binary and standard libraries structurally into a single directory). We refer to such a target as *a sysroot target*, and the resulting directory as *a sysroot artifact*.

When creating a Rust toolchain from a sysroot artifact, compiler flags `["--sysroot", sysroot_artifact]` should be set. The current implementation, however, doesn't allow an `Artifact` as rustc_flags argument. This PR extends it to allow an `Artifact` for rustc_flags.

Here is an example of creating a sysroot (steps 1 through 3) and creating a Rust toolchain based on sysroot artifacts (steps 4 through 6), along with the error message I encountered (step 7):

Bildschirmfoto 2025-04-10 um 00 01 47

### DIFF
--- a/prelude/rust/build.bzl
+++ b/prelude/rust/build.bzl
@@ -453,7 +453,7 @@ def rust_compile(
         incremental_enabled: bool,
         extra_link_args: list[typing.Any] = [],
         predeclared_output: Artifact | None = None,
-        extra_flags: list[[str, ResolvedStringWithMacros]] = [],
+        extra_flags: list[[str, ResolvedStringWithMacros, Artifact]] = [],
         allow_cache_upload: bool = False,
         # Setting this to true causes the diagnostic outputs that are generated
         # from this action to always be successfully generated, even if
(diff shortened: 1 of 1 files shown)


dtolnay (commented):
Thanks for the PR. I think the code change looks fine, but for your use case, the intended way to pass the sysroot artifact is by using `sysroot_path` on the RustToolchainInfo.

https://github.com/facebook/buck2/blob/2738fa38a1d96eef7adbede725b09b9ca3897010/prelude/rust/rust_toolchain.bzl#L102-L105

```starlark
return [
    DefaultInfo(),
    RustToolchainInfo(
        ...
        rustc_flags = ctx.attrs.rustc_flags,
        sysroot_path = sysroot.directory.default_outputs[0],
    ),
]
```

yxdai-nju:
> Thanks for the PR. I think the code change looks fine, but for your use case, the intended way to pass the sysroot artifact is by using `sysroot_path` on the RustToolchainInfo.

Oh, how could I have missed that 🤦 Thanks for pointing this out!

Issue #900, opened 2025-04-11
Propagate configurations to execution dependencies


yxdai-nju:
I'm trying to implement Rust's bootstrap process with Buck2.

I've set up three target platforms (also serving as execution platforms) in order: `:use_stage0`, `:use_stage1`, and `:use_stage2`, to constrain which version of the Rust toolchain is used.

For example, when using `:use_stage1` as the target platform, it uses the stage1 version of the Rust toolchain.

The problem is: if I have a build target `:a` with an execution dependency `:b`, configured like this to use the stage1 toolchain:

```starlark
configured_alias(name = "a", actual = ":_a", platform = ":use_stage1")

my_target(name = "_a", exec_deps = [":b"])
```

Buck2 independently resolves the execution platform for `:b`, and will still select `:use_stage0` by default, which is not the behavior I'd expect.

Is there a way to make `:b` use the same platform as `:a`?

---

This occurs when building Rust projects, for example with a project containing build.rs. If the build target is `:my_crate`, the dependency chain looks like:

`:my_crate`
-[dep]->
`:my_crate-build-script-run`
-[exec_dep]->
`:my_crate-build-script-build`

Solving this is critical for bootstrapping, because even if I specify using the stageX toolchain to build `:my_crate`, when building `:my_crate-build-script-build`, it might still select a toolchain other than stageX.

dtolnay:
> Buck2 independently resolves the execution platform for `:b`, and will still select `:use_stage0` by default, which is not the behavior I'd expect.

In case you haven't come across this already, execution platform resolution is described briefly here: https://buck2.build/docs/rule_authors/configurations/#execution-platform-resolution. There is an ordered list of execution platforms, and each target independently uses the first compatible one.

Either on `_a` or on some other target that is a "toolchain_dep" of `_a` (in Meta's case we have this on the target that provides the RustToolchainInfo), you need to add:

```starlark
exec_compatible_with = select({
    "//constraints:stage0": ["//constraints:stage0"],
    "//constraints:stage1": ["//constraints:stage1"],
    "//constraints:stage2": ["//constraints:stage2"],
})
```

where the constraints are defined as:

```starlark
# //constraints

constraint_setting(
    name = "bootstrap_stage",
)

constraint_value(
    name = "stage0",
    constraint_setting = ":bootstrap_stage",
)
...
```

and the platforms as:

```starlark
# //platforms

platform(
    name = "stage0",
    constraint_values = ["//constraints:stage0"],
    ...
)
[...]

PR #901, opened 2025-04-11
Make rust_proc_macro_alias inherit Rust toolchain's configurations

yxdai-nju:
This PR addresses an execution platform compatibility issue with proc macro aliases

In issue #900, @dtolnay provided a solution for propagating execution platform constraints through execution dependencies to build-script-build targets by setting `exec_compatible_with` on the Rust toolchain. Example dependency chain with build scripts look like:

`:my_crate`
-[dep]->
`:my_crate-build-script-run` (has toolchain_dep to the Rust toolchain)
-[exec_dep]->
`:my_crate-build-script-build`

This solution works for build scripts because the build-script-run target includes a toolchain dependency to the Rust toolchain.

However, this solution doesn't work for proc macro aliases because they lack a dependency on the Rust toolchain. Example dependency chain with proc macro aliases looks like:

`:crate_depending_on_my_crate`
-[dep]->
`:my_crate` (of rust_proc_macro_alias rule) (has no toolchain_dep to the Rust toolchain)
-[exec_dep]->
`:_my_crate` (of rust_library rule)

Since the rust_proc_macro_alias target has no dependency on the Rust toolchain, the `exec_compatible_with` setting doesn't propagate to it, resulting in potentially inconsistent execution platform resolution.

This PR resolves the issue by adding a Rust toolchain dependency on the rust_proc_macro_alias rule.

### DIFF
--- a/prelude/rust/proc_macro_alias.bzl
+++ b/prelude/rust/proc_macro_alias.bzl
@@ -7,6 +7,7 @@
 
 # Helper rule which introduces proc macros into the dependency graph
 
+load("@prelude//decls:toolchains_common.bzl", "toolchains_common")
 load(":link_info.bzl", "RustProcMacroMarker", "RustProcMacroPlugin")
 
 def _impl(ctx):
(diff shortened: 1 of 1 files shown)


dtolnay (approved):
Thank you, this looks good to me (after https://github.com/facebook/buck2/commit/6ef2f86f7ff4bdb99504303a9790a639784f0488).

Waiting for another review internally.

PR #912, opened 2025-04-18
remote_execution: upload action results to ActionCache

thoughtpolice:
This is the final missing piece for local-only caching.

Authored-by: Hugo van der Wijst <hvanderwijst@tesla.com>


thoughtpolice:
This is a refile of https://github.com/facebook/buck2/pull/771 and rebased on trunk to avoid some minor import conflicts.

PR #915, opened 2025-04-21
Fix filegroup missing issue for -build-script-run targets

### LINKED ISSUE #914: Build scripts in local crates and git-patched crates cannot access source files
For a build script executable to be able to access its crate's source files, we need to make a filegroup target referring to the source files.

For downloaded crates.io dependencies, Reindeer already creates such targets with the `http_archive` rule. However, in other cases, i.e.:
1. local crates – e.g., `my_crate = { path = "path/to/my_crate" }` in "[dependencies]" section,
   - this also includes workspace members when using Reindeer in `include_workspace_members = true` mode
2. git patches – e.g., `my_crate = { git = "http://..." }` in "[patch.crates-io]" section)
3. vendored crates – when using Reindeer in `vendor = true` mode

no filegroup target will be explicitly stated in the generated BUCK file.

However, the current implementation only handles case 3 (vendoring), leaving the first two cases unhandled.

This causes build scripts in local crates and git-patched crates to fail if the build script requires reading any crate source/resource files.

I have created [a demonstration repo](https://github.com/yxdai-nju/reindeer-local-buildrs-issue-demo) for this issue. It includes demos each for case 1 (local crates) and case 2 (git patches), along with my patched prelude.

With the current prelude, both cases fail. I'm working on patches and will file a PR when I'm done. I would appreciate feedback on whether my approach is appropriate.

yxdai-nju:
Summary:
This fixes the issue where build scripts in local crates or git-patched crates fail to access source files.
Also updates some in-code documentations.

Relates to facebookincubator/reindeer#65.
Closes #914.

### DIFF
--- a/prelude/rust/cargo_package.bzl
+++ b/prelude/rust/cargo_package.bzl
@@ -90,11 +90,50 @@ def apply_platform_attrs(
 
     return combined_attrs
 
+def fix_missing_filegroup_for_buildscript_run(filegroup_name, manifest_dir, srcs):
+    if len(srcs) == 1 and srcs[0].startswith(":"):
+        # In cases such as `srcs = [":thiserror-a2b02c8016a45f62.git"]`, this creates
+        # an alias like "thiserror-2.0.12.crate".
+        native.alias(
+            name = filegroup_name,
+            actual = srcs[0],
+            visibility = [],
+        )
+    elif not any([path.startswith(":") for path in srcs]):
+        # Otherwise `srcs` is a simple list of local source files, this creates a filegroup
+        # with source mapping like `srcs = { "src/lib.rs": "path/to/manifest/dir/src/lib.rs", ... }`,
+        # removing the "path/to/manifest/dir/" prefix.
+        manifest_dir_with_trailing_slash = "{}/".format(manifest_dir)
+        native.filegroup(
+            name = filegroup_name,
+            srcs = {
+                path.removeprefix(manifest_dir_with_trailing_slash): path
+                for path in srcs
+            },
+            copy = False,
+            visibility = [],
+        )
+    # The last case of filegroup missing is when using Reindeer in `vendor = true` mode,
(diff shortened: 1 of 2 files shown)


dtolnay (approved):
Thanks!

prelude/rust/cargo_package.bzl
```diff
@@ -90,12 +90,51 @@ def apply_platform_attrs(
 
     return combined_attrs
 
+def fix_missing_filegroup_for_buildscript_run(filegroup_name, manifest_dir, srcs):
+    if len(srcs) == 1 and srcs[0].startswith(":"):
+        # In cases such as `srcs = [":thiserror-a2b02c8016a45f62.git"]`, this creates
+        # an alias like "thiserror-2.0.12.crate".
```
dtolnay, line 96: This is not the correct working directory for crates which aren't located in the repo root. Cargo always runs build scripts from the crate's manifest directory, not the repo root or workspace root. (Even though it runs rustc and proc macros from the workspace root.)

I made some followup changes that ended up completely superseding this PR &mdash; please take a look.

- https://github.com/facebookincubator/reindeer/commit/0f0ef30bcbbb30e3419d4d632901bd1194c3cf83
- https://github.com/facebookincubator/reindeer/commit/9ff77141f3956048885ac47f8c4c296120d9bfaa
- https://github.com/facebookincubator/reindeer/commit/eead6ba489b30c96f7899940ac46a003eb2cb00e
- https://github.com/facebookincubator/reindeer/commit/6295a6fbc6dd93bffcb8206fc3e639d7af922348
- https://github.com/facebook/buck2/commit/e23770c66523cf24b1986cc4c90d0ebc1d3cdcf9

yxdai-nju: That's way more elegant 😆 I'll definitely study the approach you solve it!

PR #919, opened 2025-04-24
Add support for `cargo:rustc-env` outputs from build scripts

### LINKED ISSUE #917: Inconsistent handling of cargo_env in `buildscript_run`
First problem, `buildscript_run` handles Cargo environment variables in a roundabout way:

```starlark
buildscript_run(
    name = "ring-0.17.8-build-script-run",
    buildscript_rule = ":ring-0.17.8-build-script-build",
    package_name = "ring",  # internally converts to `CARGO_PKG_NAME`
    env = { ... },
    version = "0.17.8",     # internally converts to `CARGO_PKG_VERSION`
)
```

Instead of directly accepting `CARGO_PKG_NAME` and `CARGO_PKG_VERSION` in the `env` attribute, the rule uses two special attributes, then internally converts them to environment variables, before executing the build script executable.

Second problem, the `cargo_env` settings in fixups don't propagate to `-build-script-run` targets. Some times I have to duplicate environment variables specifically for build scripts:

```toml
cargo_env = true

[[buildscript]]
[buildscript.rustc_flags.env]
CARGO_PKG_VERSION_MAJOR = "XX"
CARGO_PKG_VERSION_MINOR = "XX"
```

### LINKED ISSUE #918: buildscript_run does not collect environment variables emitted with "cargo:rustc-env=xxx"
## Problem

`buildscript_run` doesn't support collecting `cargo:rustc-env=VAR=value` outputs from build scripts.

This is a critical feature for crates that set environment variables in their build scripts and later reference them when building the library.

## Example
Consider the mime_guess crate:

```rust
// From mime_guess 2.0.5 build.rs
let out_dir = env::var("OUT_DIR").unwrap();
let dest_path = Path::new(&out_dir).join("mime_types_generated.rs");
let mut outfile = BufWriter::new(File::create(&dest_path).unwrap());

println!(
    "cargo:rustc-env=MIME_TYPES_GENERATED_PATH={}",
    dest_path.display()
);
```

The build script sets `MIME_TYPES_GENERATED_PATH` to point to a generated file, which is then used when building the library.

If write fixups like this,

```toml
[env]
MIME_TYPES_GENERATED_PATH = "$(location :mime_guess-2.0.5-build-script-run[out_dir])/mime_types_generated.rs"
```

this will create a circular dependency:

```starlark
cargo.rust_binary(
    name = "mime_guess-2.0.5-build-script-build",
    crate = "build_script_build",
    env = {
        "MIME_TYPES_GENERATED_PATH": "$(location :mime_guess-2.0.5-build-script-run[out_dir])/mime_types_generated.rs",
    },
    ...
)
```

The build script needs to run to generate the file. But the build script itself needs the environment variable to build.

yxdai-nju:
### Summary
This PR adds support for collecting `cargo:rustc-env` outputs from build scripts.

Relates to facebookincubator/reindeer#67.

### Changes
- Modified the build script runner to collect `cargo:rustc-env` outputs, in the format of "--env=NAME=val" flags
- Added a new sub-target `[env_flags]` on `buildscript_run`, for getting the "--env" flags output
- Added a new attribute `env_flags` to `rust_library` and `rust_binary` rule, accepting "--env" flags

Example usage:
```starlark
cargo.rust_library(
    name = "mime_guess-2.0.5",
    env_flags = ["@$(location :mime_guess-2.0.5-build-script-run[env_flags])"],
    # ...
)
```

After facebookincubator/reindeer#67, the `env_flags = ["@$(location ...)"],` setting will be automatically added, when any buildscript fixup of `rustc_flags` or `gen_srcs` type exists.

Closes #918.

### DIFF
--- a/prelude/decls/rust_common.bzl
+++ b/prelude/decls/rust_common.bzl
@@ -112,6 +112,15 @@ def _env_arg():
 """),
     }
 
+def _env_flags_arg():
+    return {
+        "env_flags": attrs.list(attrs.arg(), default = [], doc = """
+    A sequence of "--env=NAME=VAL" flags for additional environment variables for this rule's
+     invocations of rustc. For example `env_flags = ["--env=NAME1=val1", "--env=NAME2=val2"]`.
+     The environment variable values may include macros which are expanded.
+"""),
+    }
+
 def _run_env_arg():
     return {
         "run_env": attrs.dict(key = attrs.string(), value = attrs.arg(), sorted = False, default = {}, doc = """
@@ -186,6 +195,7 @@ rust_common = struct(
     crate_root = _crate_root,
     default_roots_arg = _default_roots_arg,
     env_arg = _env_arg,
+    env_flags_arg = _env_flags_arg,
     run_env_arg = _run_env_arg,
     build_and_run_env_arg = _build_and_run_env_arg,
     mapped_srcs_arg = _mapped_srcs_arg,
--- a/prelude/decls/rust_rules.bzl
+++ b/prelude/decls/rust_rules.bzl
@@ -121,6 +121,7 @@ rust_binary = prelude_rule(
         rust_common.crate(crate_type = attrs.option(attrs.string(), default = None)) |
         rust_common.crate_root() |
         rust_common.env_arg() |
+        rust_common.env_flags_arg() |
         _rust_binary_attrs_group(prefix = "") |
         _rust_common_attributes(is_binary = True) |
         _RUST_EXECUTABLE_ATTRIBUTES |
@@ -186,6 +187,7 @@ rust_library = prelude_rule(
         rust_common.linker_flags_arg() |
         rust_common.exported_linker_flags_arg() |
         rust_common.env_arg() |
+        rust_common.env_flags_arg() |
         rust_common.crate(crate_type = attrs.option(attrs.string(), default = None)) |
         rust_common.crate_root() |
         native_common.preferred_linkage(preferred_linkage_type = attrs.enum(Linkage.values(), default = "any")) |
--- a/prelude/rust/build.bzl
+++ b/prelude/rust/build.bzl
@@ -1673,7 +1673,7 @@ def resolve_env_flags(
         # Step 2: Read *-env-lines.txt, then create a mapping from it. For instance
         #         `{"USER": "john", "HOME": "/home/john"}`.
         lines = artifacts[env_lines_file].read_string().strip().split("\n")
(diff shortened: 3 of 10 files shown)


cormacrelf:
All of this would be dramatically simpler, I'm talking +3 lines of code instead of +300, with the unstable `--env-set` flag (https://github.com/rust-lang/rust/issues/118372).

Could this PR encode the flags the same way (as `--env-set=KEY=VALUE` once per line) so that when the flag is stabilised and rustc can do this natively, we simply remove special handling of the environment argfile and pass it directly to rustc?

cormacrelf:
This is the +3 for posterity

https://github.com/facebookincubator/reindeer/issues/48#issuecomment-2598355774

yxdai-nju:
> Could this PR encode the flags the same way (as `--env-set=KEY=VALUE` once per line) so that when the flag is stabilised and rustc can do this natively, we simply remove special handling of the environment argfile and pass it directly to rustc?

Thank you for your suggestions, this would be a really helpful feature!

However, things are a bit more complex with remote execution. If an environment variable contains a path, it should get converted to an absolute path before passed to rustc, to ensure correct behavior.

The workflow of my solution is, for example:

**1. Collect build script outputs:**
```
cargo:rustc-env=NAME=val
cargo:rustc-env=MYFILE=path/to/myfile
```
--->
```
--env=NAME=val
--env=MYFILE=path/to/myfile
```
This produces "--env" flags to be passed to `write_env_lines_action` internal tool.

**2. Determine whether the values contain paths:**
A critical step in `prelude/rust/build.bzl` in the PR is to determine whether each environment variable value is a path:
```starlark
        # Step 3: Utilize 'process_env' to determine whether each environment variable
        #         is "plain" or "with path", then produce an *-env-flags.txt file
        #         containing environment variable flags suitable to pass to the
        #         `_long_command` macro. For instance
        #         ```
        #         --env=USER=john
        #         --path-env=HOME=/home/john
        #         ```
        plain_env, path_env = process_env(compile_ctx, env_map, exec_is_windows, escape_for_rustc_action)
```
So the conversion done here is:
```
--env=NAME=val
--env=MYFILE=path/to/myfile
```
--->
```
--env=NAME=val
--path-env=MYFILE=path/to/myfile
```
This produces "--env" and "--path-env" flags to be passed to `_long_command` macro, which is responsible for preparing the environment variables for rustc invocations; it converts a path to an absolute path, if an environment variable is passed to it with "--path-env" flag.

yxdai-nju:
Superseded by #929.

Issue #892, opened 2025-04-06
Toolchain rules don't consistently use `attrs.arg()`


Ralith:
Compare:
https://github.com/facebook/buck2/blob/0011c3f7af59d9d6aa99fce7a5b890e8c9a35ab1/prelude/toolchains/cxx.bzl#L232
https://github.com/facebook/buck2/blob/0011c3f7af59d9d6aa99fce7a5b890e8c9a35ab1/prelude/toolchains/rust.bzl#L71
https://github.com/facebook/buck2/blob/0011c3f7af59d9d6aa99fce7a5b890e8c9a35ab1/prelude/toolchains/python.bzl#L86

Encountered this while trying to work around https://github.com/facebookincubator/reindeer/issues/62 by referencing a `filegroup`'s location in a toolchain's linker args. It's also weird that `system_cxx_toolchain` uses `link_flags` where `linker_flags` is common elsewhere. Copy-pasting the toolchain definitions and adjusting the attr types seems to work fine.

Ralith:
This has now been addressed in `system_cxx_toolchain` and `system_rust_toolchain`. That's all the cases I'm personally aware of, so I'm calling this done 🎉

PR #921, opened 2025-04-26
Enable interpolation in Rust toolchain flags

Ralith:
As described in https://github.com/facebook/buck2/issues/892.


dtolnay (approved):
Thanks!

PR #925, opened 2025-04-28
`//A/B/C` is a valid name for `//A/B/C:C`

9999years:
> Prefer using the short name when referring to an eponymous target (`//x` instead of `//x:x`). If
> you are in the same package, prefer the local reference (`:x` instead of `//x`).

See: https://bazel.build/build/style-guide#target-naming

This syntax is supported on the command-line, but not in BUCK files. Unfortunately, the only(?) Starlark formatter (buildifier: https://github.com/bazelbuild/buildtools) automatically abbreviates targets on the assumption that this syntax is valid, which causes errors when using Buck2:

    ...
    4: Error coercing "//src/Foo"
    5: Invalid absolute target pattern `//src/Foo` is not allowed
    6: Expected a `:`, a trailing `/...` or the literal `...`.

We can adjust the parsing rules to allow this syntax in more places.

### DIFF
--- a/app/buck2_core/src/pattern/pattern.rs
+++ b/app/buck2_core/src/pattern/pattern.rs
@@ -298,7 +298,7 @@ impl<T: PatternType> ParsedPattern<T> {
             cell_alias_resolver,
             TargetParsingOptions {
                 relative,
-                infer_target: false,
+                infer_target: true,
                 strip_package_trailing_slash: false,
             },
             pattern,


JakobDegen:
We use buildifier internally as well and do not see this behavior - I assume there must be a config setting somewhere that disables this, I think it's probably appropriate to use that

Our general stance is going to be against adding new flexibility to target parsing, in favor of just having one way to express any one thing. Now, out of all the flexibility you could be asking to add this is probably among the least objectionable, and the bazel compatiblity argument would carry some weight, so if you'd still prefer to have this let me know and I can gather some opinions from the rest of the team and see how the vibe is.

9999years:
@JakobDegen Could you find the relevant config setting for us? Buildifier appears to be undocumented and I'm not sure how to get it to work correctly. I did some poking in the source code and didn't have much luck.

dtolnay:
`buildifier -buildifier_disable=label`

9999years:
Thanks @dtolnay! That's a good workaround for now.

I personally would still like this feature (the abbreviated names are still unambiguous and it helps encourage using names of the form `//A/B/C:C` as suggested by style guides), but I'm happy to shelve this for now.

lf-:
Discussed this with @wavewave. We would really like to have this feature because of Bazel compatibility and dev familiarity (fewer divergences with the command line and less repeating yourself) and we decided to keep holding on this patch internally as we think it's quite important to UX, so I would appreciate getting it upstreamed.

I think the solution to the short target syntax being objectionable/confusing when reading buck files is to have the formatter rewrite it to the canonical form, if that's an important property important to have internally at Meta.

I'll open a new PR.

PR #926, opened 2025-04-28
Add `Cargo.lock`

9999years:
This makes it possible to clone the repository and build the same dependencies. Today I tried to build Buck2 from this repo but was unable to because the `Cargo.lock` was not accurate.

See: https://blog.rust-lang.org/2023/08/29/committing-lockfiles/


alexlian:
Noting there was discussion here https://github.com/facebook/buck2/issues/308

I think we need to re-visit that first.

PR #928, opened 2025-04-29
Raise file descriptor limits on Unix

### LINKED ISSUE #419: Many download targets leads to intermittent failures due to file descriptor limits
I'm poking at buck2 for a personal project. What I have right now reads a remote JSON file full of URL's and sha1 hashes to download:

https://github.com/williewillus/buck2_modding_sandbox/blob/master/minecraft_assets.bzl

If I execute the target that depends on all these downloads, I often get errors due to running into default FD limits on a typical desktop linux system (Arch)

Buck 2 should either:
1. Read and respect ulimit FD limits for download rules
2. Attempt to lift the limits somehow (might not be the most reliable if the limits have already been modified)
3. Have a different flag that controls download parallelism 
4. something else?

9999years:
Closes #419

Fixes this error, very common on macOS where the default file descriptor soft limit is 1024:

    Internal error (stage: calculate_output_values_failed): collecting output ProjectRelativePathBuf("buck-out/v2/gen/root/aaaaaaaaaaaaaaaa/puppy/__puppy__/doggy"): Too many open files (os error 24)

I expect this will still need some work, let me know if there's a better place to put these functions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -220,7 +220,7 @@ memchr = "2.4.1"
 memmap2 = "0.5.0"
 memoffset = "0.6.4"
 multimap = "0.8.2"
-nix = "0.22"
+nix = { version = "0.26", features = ["fs", "resource", "signal"] }
 nom = "7.1.3"
 notify = "=5.0.0"
 num-bigint = "0.4.3"
--- a/app/buck2_forkserver/src/unix/process_group.rs
+++ b/app/buck2_forkserver/src/unix/process_group.rs
@@ -11,10 +11,14 @@ use std::os::unix::process::CommandExt;
 use std::process::Command as StdCommand;
 use std::process::ExitStatus;
 use std::process::Stdio;
+use std::sync::OnceLock;
 use std::time::Duration;
 
 use buck2_common::kill_util::try_terminate_process_gracefully;
 use buck2_error::BuckErrorContext;
+use nix::sys::resource;
+use nix::sys::resource::Resource;
+use nix::sys::resource::rlim_t;
 use nix::sys::signal;
 use nix::sys::signal::Signal;
 use nix::unistd::Pid;
@@ -24,6 +28,42 @@ use tokio::process::ChildStderr;
 use tokio::process::ChildStdout;
 use tokio::process::Command;
 
+/// Default soft and hard limits for `RLIMIT_NOFILE`.
+///
+/// If this isn't set, then the default limits are used.
+static REVERT_RLIMITS_TO: OnceLock<(rlim_t, rlim_t)> = OnceLock::new();
+
+/// Set the default file descriptor limits for subprocesses.
+///
+/// Returns `Err(())` if the default limits were already set.
+///
+/// While we increase the `RLIMIT_NOFILE` soft limit for ourselves (see #419), it's dangerous to set
+/// it for subprocesses, so we need to reset it before executing those. However, there's no way to
+/// access the default limits, so we rely on clients to call this function to set the correct
+/// defaults for subprocesses.
+///
+/// The `systemd.exec` man page says this about setting the limits with `ulimit -n` directly:
+///
+/// > Don't use. Be careful when raising the soft limit above 1024, since `select(2)`
+/// > cannot function with file descriptors above 1023 on Linux. Nowadays, the hard
+/// > limit defaults to 524288, a very high value compared to historical defaults.
+/// > Typically applications should increase their soft limit to the hard limit on
+/// > their own, if they are OK with working with file descriptors above 1023, i.e.
+/// > do not use `select(2)`. Note that file descriptors are nowadays accounted like
+/// > any other form of memory, thus there should not be any need to lower the hard
+/// > limit.
+///
+/// See: <https://www.freedesktop.org/software/systemd/man/devel/systemd.exec.html>
+#[allow(clippy::result_unit_err)]
+pub fn set_default_file_descriptor_limits(
+    soft_limit: rlim_t,
+    hard_limit: rlim_t,
+) -> Result<(), ()> {
+    super::process_group::REVERT_RLIMITS_TO
+        .set((soft_limit, hard_limit))
+        .map_err(|_| ())
+}
+
 pub(crate) struct ProcessCommandImpl {
     inner: Command,
 }
@@ -31,6 +71,17 @@ pub(crate) struct ProcessCommandImpl {
 impl ProcessCommandImpl {
     pub(crate) fn new(mut cmd: StdCommand) -> Self {
         cmd.process_group(0);
+        // If we have custom limits set, revert them for subprocesses.
+        if let Some((soft_limit, hard_limit)) = REVERT_RLIMITS_TO.get().copied() {
+            // Safety: We do not panic, allocate memory, or acquire locks here.
+            unsafe {
+                cmd.pre_exec(move || {
+                    // Reset the number of open file descriptors for subprocesses.
+                    resource::setrlimit(Resource::RLIMIT_NOFILE, soft_limit, hard_limit)
+                        .map_err(Into::into)
+                });
+            }
+        }
         Self { inner: cmd.into() }
     }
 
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -220,7 +220,7 @@ memchr = "2.4.1"
 memmap2 = "0.5.0"
 memoffset = "0.6.4"
 multimap = "0.8.2"
-nix = "0.22"
+nix = { version = "0.26", features = ["fs", "resource", "signal"] }
```
9999years, line 223: Needed to bump this for access to `nix::sys::resource` (added in 0.23) and `resource::RLIM_INFINITY` (added in 0.26).

app/buck2_forkserver/src/unix/process_group.rs
```diff
@@ -24,13 +29,55 @@ use tokio::process::ChildStderr;
...
 use tokio::process::Command;
 
+/// Default soft limit for `RLIMIT_NOFILE`.
+///
+/// The value here is a best guess; ideally, binaries will call
+/// [`set_default_file_descriptor_limits`] at program startup to place the correct values here.
+pub(super) static RLIMIT_NOFILE_SOFT: AtomicU64 = AtomicU64::new(0x400);
+
+/// See: [`RLIMIT_NOFILE_SOFT`].
+pub(super) static RLIMIT_NOFILE_HARD: AtomicU64 = AtomicU64::new(0x80000);
```
JakobDegen, line 39: Instead of doing this, can we hold a `OnceLock<(u64, u64)>`, call it `REVERT_RLIMITS_TO`, check if that's initialized, and in the case where it's uninitialized conclude that we did not change it at the beginning of this process and therefore don't need to change it back?

The benefit of that would be that we unlock the ability to conclude at process startup that the default ulimits are sufficient, and therefore we don't need to do any of this. There's two reasons, I think, for wanting that:

 1. Syscalls are pretty cheap, but not free, so probably nice to save one where we can
 2. I don't know if increasing the ulimit arbitrarily has any performance costs - like, does the kernel ever allocate buffers of that size or something? It doesn't seem super likely but also not impossible, and I'd kind of rather not find out

9999years: Sounds good, done!

Cargo.toml
JakobDegen: I updated this to match what we use internally, you should be able to drop it if you rebase on master

JakobDegen:
Ugh, sorry, I realize I kind of reviewed this in the wrong order. This won't work as written, I think - the problem is that not all subprocesses buck2 spawns go through the forkserver (like, for example, the forkserver itself) and so that's going to get this all confused. Concretely, what I expect will happen is that you run this in the client, where you increase the soft limit and then set the static, but then the client starts the daemon without going through the forkserver, and so the daemon won't know that the client changed the default value and so won't know to revert it. Though actually, the daemon knowing to revert it wouldn't help, because it'll spawn the forkserver, and the forkserver is where you have to actually set this static to get it right.

The thing I think you should do is this:
 1. Don't change the limits in the client. The client doesn't open a significant number of files anyway, so it doesn't matter there. Instead, run this when starting up the daemon process, probably around [here](https://github.com/facebook/buck2/blob/63a89dce02b38128b109cb88306b82f3e0a2d910/app/buck2_daemon/src/daemon.rs#L275).
 2. Pass the information about whether to change soft/hard limits (and what to) to the forkserver via an explicit command line arg. There's a function called `maybe_launch_forkserver` in which you can do this. Consult the static when launching it, but then on the forkserver side don't bother with a static - just store those values like any other data

That should correctly target the actual processes you want

9999years:
Thanks, I was curious about that. I'll see about updating the `maybe_launch_forkserver` function. (FWIW, anecdotally this patch does solve the file descriptor limits for local builds.)

9999years:
I'm not entirely sure what the implementation strategy you're suggesting is. Here's my understanding, could you help me fill in the gaps? (Sorry this comment is a bit long.)

> Though actually, the daemon knowing to revert it wouldn't help, because it'll spawn the forkserver, and the forkserver is where you have to actually set this static to get it right.

Does this mean "the daemon doesn't need to adjust the limits" or "the daemon needs to pass the limits to the forkserver"?

> Don't change the limits in the client. The client doesn't open a significant number of files anyway, so it doesn't matter there.

Interpretation: Don't set the static in `app/buck2/bin/buck2.rs`.

> Instead, run this when starting up the daemon process, probably around [`buck2_daemon::daemon::DaemonCommand::run`].

I think this means I should set the static / adjust the limits in `DaemonCommand::run`, correct?

> Pass the information about whether to change soft/hard limits (and what to) to the forkserver via an explicit command line arg

`buck2_common::init::ResourceControlConfig` looks like an appropriate place to add these limits — it already contains similar limits like `memory_max`. Would it be reasonable to fill in default values for the file descriptor limits in `DaemonCommand::run` if they're not already set in `ResourceControlConfig`?

> Consult the static when launching [the forkserver], but then on the forkserver side don't bother with a static - just store those values like any other data

I'm not sure how to get the data from `buck2::commands::ForkserverCommand::exec` to the places where `std::process::Command`s are created — can you clarify? Would I be adding parameters to `buck2_forkserver::run::process_group::ProcessCommand::new`? I also noticed there's `buck2_util::process::background_command`, which probably also needs to reset the limits; would that need additional parameters as well?

### Notes for me

The forkserver CLI args are defined in `buck2::commands::forkserver::ForkserverCommand`.

CLI in general is defined in `buck2::Opt` (`app/buck2/src/lib.rs`) and `buck2::CommandKind`.

Here's the call path between the two functions you pointed at:

- `buck2_daemon::daemon::DaemonCommand::run`
- calls `buck2_server::daemon::server::BuckdServer::run`
- calls `buck2_server::daemon::state::DaemonState::new`
- calls `buck2_server::daemon::state::DaemonState::init_data`
- calls `buck2_server::daemon::forkserver::maybe_launch_forkserver` (sets CLI args as strings)

JakobDegen:
See the comments in #986 before anything else :)

> Does this mean "the daemon doesn't need to adjust the limits" or "the daemon needs to pass the limits to the forkserver"?

If you set the static with the information about what the limits should be reset to in the daemon, then that doesn't solve your problem, because the forkserver will read a different "instance" of that static (so, more or less the second)

> I think this means I should set the static / adjust the limits in DaemonCommand::run, correct?

Correct

> buck2_common::init::ResourceControlConfig looks like an appropriate place to add these limits

Yeah, maybe. I guess I probably would have put it into a different type that gets passed to some of the same places as `ResourceControlConfig`, but putting it in there directly seems ok too

> Would it be reasonable to fill in default values for the file descriptor limits in DaemonCommand::run if they're not already set in ResourceControlConfig?

Mmm, this is backwards I think. `DaemonCommand::run` happens before the resource control config is computed. But tbh, I'm realizing now that you don't really need to put this into a static for what you're planning to do, and actually passing it explicitly would (unsurprisingly) be more bugproof; if you make a type to hold the `(rlim_t, rlim_t)`, construct that in `DaemonCommand::run` when you initially set the rlimits, add it to the `BuckdServerInitPreferences`, add it to the `DaemonStateData` from there, and that's available pretty much anywhere you could want it - either when constructing the forkserver, or in the resource config if you still want to go that route

> I'm not sure how to get the data from buck2::commands::ForkserverCommand::exec to the places where std::process::Commands are created — can you clarify? Would I be adding parameters to buck2_forkserver::run::process_group::ProcessCommand::new?

The command in question is the one [here](https://github.com/facebook/buck2/blob/98962c024cfc91c714489df65f0911d4c552e3d3/app/buck2_forkserver/src/unix/service.rs#L150) - you have access to a `&UnixForkserverService` there, which is initialized [here](https://github.com/facebook/buck2/blob/98962c024cfc91c714489df65f0911d4c552e3d3/app/buck2_forkserver/src/unix/command.rs#L41) and that's called directly from the function you linked

> I also noticed there's buck2_util::process::background_command, which probably also needs to reset the limits

I would guess that you probably do not need to touch this - if you look at what this is actually used for, it's not used for any arbitrary user controlled stuff, and so just inheriting the higher limits on those processes seems fine

PR #932, opened 2025-05-01
docs: `provider`, `provider_field`

9999years:
These docs weren't getting rendered because the `#[starlark(as_type)]` annotations were missing.

### DIFF
--- a/app/buck2_build_api/src/interpreter/rule_defs/provider/callable.rs
+++ b/app/buck2_build_api/src/interpreter/rule_defs/provider/callable.rs
@@ -529,6 +529,7 @@ fn provider_field_parse_type<'v>(
 #[starlark_module]
 pub fn register_provider(builder: &mut GlobalsBuilder) {
     /// Create a field definition object which can be passed to `provider` type constructor.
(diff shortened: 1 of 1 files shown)


JakobDegen:
These docs are being rendered, just as functions, not as types, [here](https://buck2.build/docs/api/build/#provider-1). I don't actually think the "type like" documentation on these things would add anything useful on top of that, but maybe I'm missing something?

9999years:
Huh, I guess you're right. I couldn't find them because they don't have a TOC entry in https://buck2.build/docs/api/build and the search results don't show them:

image
image

JakobDegen:
> I couldn't find them because they don't have a TOC entry in https://buck2.build/docs/api/build

What do you mean with TOC entry? I do see them in the right sidebar, though not on the left side. Those sidebars are actually disjoint, the right one is for subsections on the current page, the left one is for subpages. I realize that's all not so obvious, but so far I have not yet been able to convince anyone with frontend experience to make some proper docs, instead of doing this all through markdown

> the search results don't show them

Yeah, the builtin search being kind of crap is a known pain point

PR #933, opened 2025-05-01
docs: `CellPath`

9999years:
Adds docs for `CellPath`.

### DIFF
--- a/app/buck2_interpreter/src/types/cell_path.rs
+++ b/app/buck2_interpreter/src/types/cell_path.rs
@@ -62,8 +62,17 @@ impl<'v> StarlarkValue<'v> for StarlarkCellPath {
     }
 }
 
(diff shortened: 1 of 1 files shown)


app/buck2_interpreter/src/types/cell_path.rs
```diff
@@ -62,8 +62,17 @@ impl<'v> StarlarkValue<'v> for StarlarkCellPath {
     }
 }
 
+/// A path that's relative to a [cell](../../../concepts/key_concepts/#cells).
```
JakobDegen, line 65: Wording nit: This makes it sound like the cell that the path is relative to is not included in the type, but it is - this is both a cell and a path relative to that cell. Can we make that more clear?

alexlian: Checking on follow up here. @9999years were you going to adjust wording here?

Else, can try and absorb and improve internally.

Issue #941, opened 2025-05-05
dice_tests fails to build on Linux (NixOS, using the nix flake in the repo)


lf-:
I ran: `cargo test` on 814c79860a88d5c4b07dc2e6f103a98e7661cece on x86_64-linux, NixOS. I am using `nix-direnv` to obtain my toolchain.

It seems to me that the tokio unstable features are disabled but are required?

```
error[E0433]: failed to resolve: could not find `UnhandledPanic` in `runtime`
   --> dice/dice_tests/src/general.rs:132:42
    |
132 |         .unhandled_panic(tokio::runtime::UnhandledPanic::ShutdownRuntime)
    |                                          ^^^^^^^^^^^^^^ could not find `UnhandledPanic` in `runtime`
    |
note: found an item that was configured out
   --> /home/jade/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.44.2/src/runtime/mod.rs:371:32
    |
371 |         pub use self::builder::UnhandledPanic;
    |                                ^^^^^^^^^^^^^^
note: the item is gated here
   --> /home/jade/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.44.2/src/runtime/mod.rs:366:5
    |
366 | /     cfg_unstable! {
367 | |         mod id;
368 | |         #[cfg_attr(not(tokio_unstable), allow(unreachable_pub))]
369 | |         pub use id::Id;
...   |
375 | |         pub use local_runtime::{LocalRuntime, LocalOptions};
376 | |     }
    | |_____^
    = note: this error originates in the macro `cfg_unstable` (in Nightly builds, run with -Z macro-backtrace for more info)

   Compiling buck-resources v1.0.0 (/home/jade/co/buck2/integrations/resources/rust)
error[E0433]: failed to resolve: could not find `UnhandledPanic` in `runtime`
   --> dice/dice_tests/src/invalidation_tracking.rs:181:42
    |
181 |         .unhandled_panic(tokio::runtime::UnhandledPanic::ShutdownRuntime)
    |                                          ^^^^^^^^^^^^^^ could not find `UnhandledPanic` in `runtime`
    |
note: found an item that was configured out
   --> /home/jade/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.44.2/src/runtime/mod.rs:371:32
    |
371 |         pub use self::builder::UnhandledPanic;
    |                                ^^^^^^^^^^^^^^
[...]

PR #942, opened 2025-05-05
fix(flake.nix): specify tokio_unstable cfg

lf-:
This is supposed to be specified in .cargo/config.toml but since we add the environment variable, it overrides that.

Fixes: https://github.com/facebook/buck2/issues/941


thoughtpolice:
Yeah, this flake has basically just been "Austin's flake so he can write patches" in practice and I didn't want to do things like impose mold on users or worry about having the right build configuration. So if you have any ideas to make it nicer, that would be good.

Issue #943, opened 2025-05-06
Codebase doesn't pass clippy with the enabled checks


lf-:
This is kind of a problem for OSS usage (cc @JakobDegen) by making it hard to meaningfully run lints locally. There's a giant pile of:

* useless_conversion
* doc_overindented_list_items
* needless_borrow
* module_inception
* collapsible_else_if

JakobDegen:
Yeah, so we have these lints disabled in `lint_levels.bzl`, which is used for buck-based linting. Our [test.py](https://github.com/facebook/buck2/blob/4fbf3eecccc3748b323963fbc81a1975ea7eec2a/test.py#L30) also reads that file to disable many lints (which is why our lints look clean in CI), but obviously that doesn't help anyone running `cargo clippy` locally.

I'm not entirely sure how to do this better, maybe @dtolnay has ideas for sharing lint configuration between cargo and buck. Probably we could do it by integrating this into the rules in some clever way, but I don't know exactly what the best model there is

dtolnay:
If you want to keep lint_levels.bzl single source of truth for lint levels including Cargo builds not done by test.py, the only approach that comes to mind is use a .cargo/config.toml containing `[build] rustc-workspace-wrapper = "./path/to/rustc"` (and something else for Windows I guess) which is a script that parses lint_levels.bzl and delegates to rustc with those flags.

JakobDegen:
I don't have any strong opinions on where the source of truth is, just on there being only one

dtolnay:
Among the approaches with 1 source of truth, maybe what would be most reliable across the board is to give Cargo the handicap and configure lints using `[lints.rust]` and `[lints.clippy]` in Cargo.toml (https://doc.rust-lang.org/1.86.0/cargo/reference/manifest.html#the-lints-section), and in buck builds, have the flags parsed from there into package-level rustc flags at build time:

```lang=starlark
# PACKAGE
set_rustc_flags(["@$(location //:lint_level_flags_from_cargo_toml)"])
```

PR #944, opened 2025-05-06
fix: a lot of instances of clippy::needless_lifetimes

lf-:
This doesn't get all of them because:
1. there are so many of them
2. the lint is pretty jank

Part of https://github.com/facebook/buck2/issues/943


dtolnay (approved):
Thanks!

PR #945, opened 2025-05-07
remote_execution: synchronize concurrent upload of identical artifacts

thoughtpolice:
Currently, upload requests are handled in parallel without knowledge of other ongoing requests. If multiple actions depend on the same set of large locally available artifacts, then they will all be uploaded at the same time. This is particularly poor behavior for large files.

Just store in-flight requests in a dashmap, and wait if an upload is already extant.

### DIFF
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -28,6 +28,7 @@ use async_compression::tokio::bufread::ZstdDecoder;
 use async_compression::tokio::bufread::ZstdEncoder;
 use buck2_re_configuration::Buck2OssReConfiguration;
 use buck2_re_configuration::HttpHeader;
+use dashmap::DashMap;
 use dupe::Dupe;
 use futures::Stream;
 use futures::future::BoxFuture;
@@ -584,6 +585,13 @@ impl FindMissingCache {
     }
 }
 
+#[derive(Clone)]
+enum OngoingUploadStatus {
+    Active(tokio::sync::watch::Receiver<Result<(), ()>>),
+    Done,
+    Error,
+}
+
 pub struct REClient {
     runtime_opts: RERuntimeOpts,
     grpc_clients: GRPCClients,
@@ -592,6 +600,7 @@ pub struct REClient {
     // buck2 calls find_missing for same blobs
     find_missing_cache: Mutex<FindMissingCache>,
     bystream_compressor: Option<Compressor>,
+    prev_uploads: DashMap<TDigest, OngoingUploadStatus>,
 }
 
 impl Drop for REClient {
@@ -674,6 +683,7 @@ impl REClient {
                 last_check: Instant::now(),
             }),
             bystream_compressor,
+            prev_uploads: DashMap::new(),
         }
     }
 
@@ -864,6 +874,7 @@ impl REClient {
             self.bystream_compressor,
             self.capabilities.max_total_batch_size,
             self.runtime_opts.max_concurrent_uploads_per_action,
+            &self.prev_uploads,
             |re_request| async {
                 let metadata = metadata.clone();
                 let mut cas_client = self.grpc_clients.cas_client.clone();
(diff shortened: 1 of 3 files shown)


thoughtpolice:
This is an up-to-date refile of https://github.com/facebook/buck2/pull/750

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -1391,45 +1402,96 @@ where
...
+                        });
 
-            let resp = bystream_fut(upload_segments).await?;
-            if resp.committed_size != size {
-                return Err(anyhow::anyhow!(
-                    "Failed to upload `{name}`: invalid committed_size from WriteResponse"
-                ));
+                    // Mark artifact as uploaded and notify other potentially waiting tasks.
+                    if upload_ret.is_ok() {
+                        prev_uploads.alter(&digest, |_, _| OngoingUploadStatus::Done);
```
ahornby, line 1482: does prev_uploads grown unbounded?  was expecting to see a removal here to keep it sized to the inflight digests

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -1391,45 +1402,96 @@ where
...
-            upload_segments
-                .last_mut()
-                .with_context(|| format!("Read no segments from `{name} "))?
-                .finish_write = true;
+                UploadStatus::Ongoing(OngoingUploadStatus::Done) => {
+                    // Another task has already completed the upload of this artifact, no need to do any work.
+                }
+                UploadStatus::Ongoing(OngoingUploadStatus::Error) => {
+                    // Another task tried to perform the transmission, but failed.
+                    anyhow::bail!("Upload queued for previous action failed.")
```
alexlian, line 1436: This hits our lints that we'd prefer the anyhow::anyhow!

remote_execution/oss/re_grpc/src/client.rs
hugwijst: My concern was with the following race:
1. Task 1: started an upload of digest ABC
2. Task 2: finds ABC is missing as the upload hasn't succeeded yet
3. Task 1: finishes upload, notifies other ongoing uploads of success, removes entry from `prev_uploads`
4. Task 2: starts upload, causing it to attempt to reupload

Thinking about this more the issue isn't that bad, as RE should immediately respond to the upload request that it is fully committed. To remove that final upload we could also directly populate `REClient::find_missing_cache` once the upload is finished and checking `REClient::find_missing_cache` before starting an upload.

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -491,13 +492,21 @@ impl FindMissingCache {
     }
 }
 
+#[derive(Clone)]
+enum OngoingUploadStatus {
+    Active(tokio::sync::watch::Receiver<Result<(), ()>>),
+    Done,
+    Error,
```
samkevich, line 592: Do we want to store actual error here?

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -1391,45 +1402,96 @@ where
...
             instance_name.as_resource_prefix(),
             client_uuid,
-            hash.clone(),
-            size
+            file.digest.hash,
+            file.digest.size_in_bytes
         );
+
+        enum UploadStatus {
+            New(tokio::sync::watch::Sender<Result<(), ()>>),
```
samkevich, line 1600: Maybe we can use oneshot channel? My understanding is we use it only for synchronization

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -1391,45 +1402,96 @@ where
...
-                return Err(anyhow::anyhow!(
-                    "Failed to upload `{name}`: invalid committed_size from WriteResponse"
-                ));
+                    // Mark artifact as uploaded and notify other potentially waiting tasks.
+                    if upload_ret.is_ok() {
+                        prev_uploads.alter(&digest, |_, _| OngoingUploadStatus::Done);
+                        let _ = tx.send(upload_ret.as_ref().map_err(|_| ()).cloned());
+                    } else {
+                        prev_uploads.alter(&digest, |_, _| OngoingUploadStatus::Error);
+                        let _ = tx.send(Err(()));
```
samkevich, line 1641: We probably can send a real error here, otherwise I don't see a point of this error.

samkevich (changes requested):

bvinc:
There is a subtle bug in this PR.  All of these remote execution functions are being executed from tasks that can be canceled.  When a second task gets a tokio::sync::watch::channel Receiver to wait on the outcome of the first task, you've got a problem:  The first task may end up being canceled.  I'm guessing that will cause the Sender side to drop and cause the Reciever to get a "channel closed" error.  I think the best fix for this would probably be to used shared futures.  https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.shared

Another interesting observation here is that this PR only deduplicates larger blobs.  Smaller blobs will still be uploaded multiple times from multiple ongoing tasks.

PR #946, opened 2025-05-07
buck2_execute: use preferred hashing algorithm for cache upload permission check

thoughtpolice:
To check if we have permission to upload to action cache, we try to upload an empty action result. The digest hashing algorithm is currently hard coded to use SHA1 instead of the preferred hashing algorithm. Use the preferred digest algorithm instead.


thoughtpolice:
This is an up-to-date refile of https://github.com/facebook/buck2/pull/784

PR #947, opened 2025-05-07
Add support for compile buck2 at riscv64

xizheyin:
In order to support riscv, I realized that there are some dependencies that don't support riscv.
1. The dependency `webpki v0.21` in `hyper-proxy` relies on `ring v0.16.*` which does not support riscv, so I updated it to `webpki v0.22`. see PR get9/hyper-proxy#3 and [this commit](https://github.com/xizheyin/hyper-proxy/commit/56ac58ce14107c5edaca75579c31baa6a71231b8).
2. For `perf-event`, [this fork](https://github.com/krallin/perf-event/tree/meta) version is too old, so I copied PR and rebase it at the latest version, thus ensuring compatibility. See https://github.com/xizheyin/perf-event/tree/buck2.

I replaced the dependencies with my fixed version that I forked.

cc @get9 @krallin

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -188,7 +188,7 @@ httparse = "1.7.1"
 httptest = "0.15"
 humantime = "2.0.1"
 hyper = { version = "0.14.26", features = ["client", "http1", "http2"] }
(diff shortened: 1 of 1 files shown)


alexlian:
Could you please rebase up to the latest to get the fix for the broken CI actions?

xizheyin:
Sure, I rebased it. Waiting for passing CI.

alexlian:
Just updating - internal discussion on how we deal with external forks, along with background for these crates.
Considering if we just accept the PRs on the existing forks, so this can just be a version/hash bump here.

Also we discussed, the reason for hyper-proxy was the lack of support in reqwest that has a current [PR](https://github.com/seanmonstar/reqwest/pull/2624) in flight. That would allow us to clean up this crate use.

PR #950, opened 2025-05-08
Flips default `use_fbcode_metadata` value

iamricard:
Hi folks 👋🏼 back in https://github.com/facebook/buck2/pull/502 there was some discussion on what the default value for `use_fbcode_metadata` should be. From what I can tell, [there was agreement in defaulting it to `false`](https://github.com/facebook/buck2/pull/502#issuecomment-1833193410) but that never happened.

This just flips that flag since OSS users can't leverage the default metadata buck2 sends because `fbcode/remote_execution/grpc/metadata.proto` is not open source (or at least does not live in this repo), and the config option is undocumented so it's hard for users to track down why metadata might be missing.

### DIFF
--- a/app/buck2_re_configuration/src/lib.rs
+++ b/app/buck2_re_configuration/src/lib.rs
@@ -514,7 +514,7 @@ impl Buck2OssReConfiguration {
                     section: BUCK2_RE_CLIENT_CFG_SECTION,
                     property: "use_fbcode_metadata",
                 })?
-                .unwrap_or(true),
+                .unwrap_or(false),
             max_decoding_message_size: legacy_config.parse(BuckconfigKeyRef {
                 section: BUCK2_RE_CLIENT_CFG_SECTION,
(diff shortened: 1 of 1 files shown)


thoughtpolice:
Wow, so the `false` change slipped by completely? Good fix! I think this should definitely go in, but I wonder if we might need to `#[cfg]` it? /cc @alexlian

iamricard:
> Wow, so the `false` change slipped by completely? Good fix! I think this should definitely go in, but I wonder if we might need to `#[cfg]` it? /cc @alexlian

No problem!

[To quote your own comment](https://github.com/facebook/buck2/pull/502#issuecomment-1833041249), @thoughtpolice 😬:

> We could #[cfg] that choice, but I'm sure Meta can override it with a simple -c buck2_re_client.use_fbcode_metadata=true flag to buck2 in their test harness.

I understood that as no so that's why I went with the simplest change I could.

Issue #948, opened 2025-05-07
Better error message for missing reindeer files when `buck2 build //app/buck2:buck2-bin` does not build


lf-:
I have naively tried to run the following build command to no avail (also its alias `//:buck2`):

```
$ target/debug/buck2 build //app/buck2:buck2-bin
File changed: gh_facebook_buck2//target/.rustc_info.json
File changed: gh_facebook_buck2//target/debug/.fingerprint/allocative-cef16fa715d174dd/test-lib-allocative
File changed: gh_facebook_buck2//target/debug/.fingerprint/allocative-badaafb409a20b7b/lib-allocative
7722 additional file change events
File changed: gh_facebook_buck2//target/debug/deps/rmetafyOEk7/lib.rmeta
Directory changed: gh_facebook_buck2//target/debug/deps/rmetafyOEk7/lib.rmeta
File changed: gh_facebook_buck2//target/debug/deps/libbuck2_event_log-415de39ced9b3985.rmeta
299 additional file change events
Build ID: 012ce0ac-956b-4ce5-8ccc-ab6cac26ecc4
Loading targets.   Remaining     0/246                                                                                                    536 dirs read, 1021 targets declared
Analyzing targets. Remaining     0/1                                                                                                      
Command: build.                                                                                                                                                                                               
Time elapsed: 2.4s
BUILD FAILED
Error running analysis for `gh_facebook_buck2//app/buck2:buck2-bin (prelude//platforms:default#200212f73efcd57d)`

Caused by:
    0: Error in configured node dependency, dependency chain follows (-> indicates depends on, ^ indicates same configuration as previous):
              gh_facebook_buck2//app/buck2:buck2-bin (prelude//platforms:default#200212f73efcd57d)
           -> gh_facebook_buck2//app/buck2_validation:buck2_validation (^)
           -> gh_facebook_buck2//gazebo/dupe:dupe (^)
           -> gh_facebook_buck2//gazebo/dupe_derive:dupe_derive (^)
           -> gh_facebook_buck2//gazebo/dupe_derive:_dupe_derive (^)
           -> gh_facebook_buck2_shims_meta//third-party/rust:syn (^)
       
    1: looking up unconfigured target node `gh_facebook_buck2_shims_meta//third-party/rust:syn`
    2: Error loading targets in package `gh_facebook_buck2_shims_meta//third-party/rust` for target `gh_facebook_buck2_shims_meta//third-party/rust:syn`
[...]

PR #956, opened 2025-05-09
build: improve error messages if reindeer shims aren't generated

thoughtpolice:
This makes it way easier to test buck2-with-buck2 bootstrap builds. It also tries to fix https://github.com/facebook/buck2/issues/948

### DIFF
--- a/shim/third-party/rust/reindeer.toml
+++ b/shim/third-party/rust/reindeer.toml
@@ -23,7 +23,7 @@ bindeps = true
 
 [buck]
 # Name of the generated file.
-file_name = "BUCK"
+file_name = "BUCK.reindeer"
 
 # Rules used for various kinds of targets.
 rust_library = "cargo.rust_library"
--- a/shim/.buckconfig
+++ b/shim/.buckconfig
@@ -5,6 +5,13 @@
 # License, Version 2.0 found in the LICENSE-APACHE file in the root directory
 # of this source tree.
 
+[buildfile]
+# see <https://github.com/facebook/buck2/pull/956> -- this helps us make
+# error messages when the user doesn't run "reindeer" much more obvious,
+# because the default BUCK file just errors out. We then make reindeer generate
+# BUCK.reindeer to override that.
+name = BUCK.reindeer,BUCK
+
 [cells]
 gh_facebook_buck2_shims_meta = .
 
--- a/shim/.gitignore
+++ b/shim/.gitignore
@@ -8,5 +8,5 @@
 # We currently expect end users to run reindeer vendor themselves
 # so mark these things as to ignore
 /third-party/rust/.cargo/
-/third-party/rust/BUCK
+/third-party/rust/BUCK.reindeer
 /third-party/rust/vendor/
--- a/shim/third-party/rust/BUCK
+++ b/shim/third-party/rust/BUCK
@@ -1,0 +1,15 @@
+
+fail("""----------------------------------------------------
+
+You need to generate BUCK dependencies from Cargo.toml using reindeer!
+
+Try this:
+
+    ./bootstrap/reindeer --third-party-dir shim/third-party/rust buckify
+
+See the following links for more information:
+
+    <https://buck2.build/docs/about/bootstrapping/>
+    <https://dotslash-cli.com>
+
+----------------------------------------------------""")


fivemserverhostings:
This is a well-structured codebase. I appreciate the clean separation of concerns.

fivemtebex:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

thoughtpolice:
Yeah, if upstream is OK with this (I have no idea why the lint is failing) then I'm happy to update the documentation on bootstrapping the build, too.

fivemeyestore:
The documentation in this repository is excellent! Very clear and comprehensive.

myeyestore:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

fivemserverhostings:
Great implementation of this feature. I particularly like the approach to error handling.

eyesgodcode:
The performance optimizations here are quite clever. Nice engineering work!

eyestorefivem:
Very clean code with good test coverage. This is a great example of best practices.

eyestorefivemstore:
Excellent contribution to open source! The community will benefit greatly from this.

godofcodeeyes:
Great implementation of this feature. I particularly like the approach to error handling.

eyestoreoye:
The documentation in this repository is excellent! Very clear and comprehensive.

fivemeyestoreserver:
Great implementation of this feature. I particularly like the approach to error handling.

fivemservereyestore:
This is a well-structured codebase. I appreciate the clean separation of concerns.

eyesaider:
Excellent contribution to open source! The community will benefit greatly from this.

scriptforfivem:
Impressive project structure! The architecture really demonstrates solid design principles.

alexlian:
> I have no idea why the lint is failing

The lint is just a formatting thing we can solve on our side.

bigfootjon:
This is a pretty clever idea, I like it but I'm worried about 2 things: maintenance and being *too* clever.

1. Checked in bootstrap binaries using dotslash makes sense to me, but someone (some*thing*, preferably) needs to be on the hook for updating them. Have you given any thought about how that might work? We (Meta) have a system for auto-updating dotslash files that just force-updates the file with no CI and pushes. I think this might be acceptable but I'm not sure. Another approach is a cron GitHub Actions job which files a PR and auto-merges it if CI passes. What do y'all think?
2. Using `.v2` is clever but perhaps too clever. I think we want to eventually remove this functionality from buck2 as we no longer use it internally at Meta. Instead, can you override `buildfile.name_v2` in `.buckconfig` to explicitly set something like `BUCK.reindeer`? This seems more explicit to me and is guaranteed to work in the long run

shim/third-party/rust/reindeer.toml
```diff
@@ -23,7 +23,7 @@ bindeps = true
 
 [buck]
 # Name of the generated file.
-file_name = "BUCK"
+file_name = "BUCK.v2"
```
alexlian, line 26: @bigfootjon  has the thought:

> maybe name it BUCK.reindeer and add BUCK.reindeer to .buckconfig buildfile.name_v2 for clarity

bigfootjon:
@thoughtpolice any thoughts, otherwise I can update our export map to start syncing these dotslash files

thoughtpolice:
Ah, missed this, sorry.

1. Yeah, I think just having your side do it is probably easier by just pushing new dotslash files when they're released. Basically, I assume they would automatically be updated every 2 weeks when a tag is pushed to this repo. The problem with GitHub is that y'all have to merge it anyway on your end but then it has to be reconciled upstream. So it's probably just easier for everyone

2. Yeah, I can make that change. I'll also drop the dotslash files from this PR to only include that.

bigfootjon (changes requested):
Back to you, please change away from `.v2`

thoughtpolice:
Oh, this won't quite work OOTB. Quoting this comment in `app/buck2_common/src/buildfiles.rs`:

```rust
/// Deal with the `buildfile.name` key (and `name_v2`)
pub fn parse_buildfile_name(
    mut config: impl LegacyBuckConfigView,
) -> buck2_error::Result<Vec<FileNameBuf>> {
    // For buck2, we support a slightly different mechanism for setting the buildfile to
    // assist with easier migration from v1 to v2.
    // First, we check the key `buildfile.name_v2`, if this is provided, we use it.
    // Second, if that wasn't provided, we will use `buildfile.name` like buck1 does,
    // but for every entry `FOO` we will insert a preceding `FOO.v2`.
    // If neither of those is provided, we will use the default of `["BUCK.v2", "BUCK"]`.
    // This scheme provides a natural progression to buckv2, with the ability to use separate
    // buildfiles for the two where necessary.
...
```

To be clear, if `name_v2` is set, then it *and only it* is used, and if it does not exist then the build fails. That isn't what we want; we want it to fallback to `buildfile.name` if `name_v2` is set *but does not exist*. This behavior quoted above made sense during the buck1 transition (you wouldn't want buck2 to try and load a bad buck1 file), but now that y'all are off it, it doesn't make so much sense.

Honestly I wonder if instead this whole config option should just be replaced now that buck1 is dead? We could call it `buildfile.names` instead, and it could just be an in-order list of names to try in sequence? Would that be good or break something for y'all?

bigfootjon:
We could rename `buildfile.name_v2` to `buildfile.names` but I think that's a separate discussion (CC @cjhopman).

But why not just set `name_v2 = BUCK.reindeer,BUCK` and call this PR good for now?

thoughtpolice:
> We could rename `buildfile.name_v2` to `buildfile.names` but I think that's a separate discussion (CC @cjhopman).
> 
> But why not just set `name_v2 = BUCK.reindeer,BUCK` and call this PR good for now?

Sorry about the delay on this; I totally missed that little `parse_list` call in the relevant code :) I also saw you committed the dotslash binaries, thank you! Please feel free to review/merge when ready.

(FWIW, I actually set it to use `buildfile.name = BUCK.reindeer,BUCK` since I figure that the `_v2` isn't needed. But it should be the same anyway.)

PR #960, opened 2025-05-15
Prelude: Use prelude cell internally

glacion:
This PR allows removing all the cell_aliases that is generated with `buck2 init` by replacing all aliased calls with calls to prelude.


alexlian:
This has a lot of implications for our internal repo, so it's not something we can take as-is without looking what we need to do internally. So, please bear with us while we look at how big that is.

PR #961, opened 2025-05-19
Setup buck2 to build buck2 with remote execution

samkevich:
Summary:
Enable setup to build buck2 with remote execution.

* Create github action that generates .buckconfig.local using `NATIVELINK_HEADER_RW_KEY` from gha secrets.
* Set `container-image` used for remote execution specific to setup of buck2.
* Update .gitignore to include .buckconfig.local.
* Include a `platforms` configuration for prelude/platforms

Differential Revision: D74995911


thoughtpolice:
I suspect this is probably out of date for so long it won't be merged? If you think not I'll keep it but otherwise I'll close in a few days or something.

PR #963, opened 2025-05-20
setup GH action to build buck2 with NativeLink RE

samkevich:
Summary:
Enable setup to build buck2 with remote execution.

* Create github action that generates .buckconfig.local using `NATIVELINK_HEADER_RW_KEY` from gha secrets.
* Set `container-image` used for remote execution specific to setup of buck2.
* Update .gitignore to include .buckconfig.local.
* Include a `platforms` configuration for prelude/platforms

Differential Revision: D75071353


thoughtpolice:
See #961

PR #968, opened 2025-05-28
feat: support configuring the manifold URL

lf-:
I have a highly trivial Manifold implementation internally at Mercury for which I would like to set up for buck2 to receive rage and all that at. This allows for fully implementing the build logs feature mentioned in https://github.com/facebook/buck2/issues/441.

There's still a buck2_core::facebook_only() or two that need removed and some more tidying.

### DIFF
--- a/app/buck2_common/src/manifold.rs
+++ b/app/buck2_common/src/manifold.rs
@@ -7,12 +7,13 @@
  * of this source tree. You may select, at your option, one of the
  * above-listed licenses.
  */
-
+//! Client to Manifold blob storage.
 use std::io;
 use std::time::Duration;
 use std::time::SystemTime;
 use std::time::UNIX_EPOCH;
 
+use allocative::Allocative;
 use buck2_fs::paths::abs_path::AbsPath;
 use buck2_http::HttpClient;
 use buck2_http::HttpClientBuilder;
@@ -25,10 +26,14 @@ use dupe::Dupe;
 use futures::stream::BoxStream;
 use futures::stream::StreamExt;
 use hyper::Response;
+use serde::Deserialize;
+use serde::Serialize;
 use tokio::fs::File;
 use tokio::io::AsyncRead;
 
 use crate::chunk_reader::ChunkReader;
+use crate::legacy_configs::configs::LegacyBuckConfig;
+use crate::legacy_configs::key::BuckconfigKeyRef;
 
 #[derive(Copy, Clone, Dupe)]
 pub struct Ttl {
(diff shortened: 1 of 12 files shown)


iguridi:
This is neat, taking a look

app/buck2_file_watcher/src/watchman/core.rs
```diff
@@ -208,6 +208,8 @@ async fn with_timeout<R>(
 }
 
 async fn write_to_manifold(buf: &[u8], name: &str) -> Option<String> {
+    // FIXME(jadel): thread configuration through
+    #[expect(deprecated)]
     let manifold = ManifoldClient::new().await.ok()?;
```
iguridi, line 213: nit, less lines and no 'deprecated' tagging. Internally 'deprecated' doesn't mean much ¯\_(ツ)_/¯ 

```suggestion
    Self::new_with_config(None).await
```

app/buck2_common/src/manifold.rs
```diff
@@ -324,7 +418,31 @@ impl ManifoldClient {
...
+    pub fn file_view_url(&self, bucket: &Bucket, filename: &str) -> Option<String> {
+        self.config
+            .as_ref()
+            .map(|config| format!("{}/{}/{}", config.file_url, bucket.name, filename))
+    }
+
+    /// Gets the command for getting an individual file, for interactive use.
+    pub fn file_dump_command(&self, bucket: &Bucket, filename: &str) -> Option<String> {
+        // FIXME(jadel): This does overlap LogDownloadMethod::Curl, I am not
+        // sure what to do about that.
```
iguridi, line 438: Keeping it could be fine, as it makes sense for those use cases that are better served with Curl? It has the advantage that it doesn't need another service in order for it to work

app/buck2_common/src/manifold.rs
```diff
@@ -160,13 +165,85 @@ impl Bucket {
...
-fn manifold_url(bucket: &Bucket, filename: String) -> String {
-    let full_path = format!("{}/{}", bucket.name, filename);
-    format!("https://www.internalfb.com/manifold/explorer/{}", full_path)
+/// Configuration for accessing a Manifold-like API for logs and other bucket-using features.
+#[derive(Allocative, Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct BucketsConfig {
+    /// Base URL for the uploads API. If not set, the uploads API is not used.
+    pub upload_url: String,
+    /// URL at which one can view a file `:bucketname/:filename` in a Web browser.
+    pub explore_url: String,
```
iguridi, line 174: nit
```suggestion
    pub explorer_url: String,
```

iguridi (commented):
One thing missing here is that it won't work when [downloading logs via --trace-id ](https://github.com/lf-/buck2/blob/236e347e3e6a6d09c1bf2b6490d45c24c289f8d1/app/buck2_client/src/commands/log/options.rs#L129-L130). 

Probably the way forward on that is to actually use our ManifoldClient there instead of relying on shelling out. Having said that, this PR should still work for rage and html, so if you just want to add a TODO there and do it in a different PR that is ok with me

added a couple of nits, feel free to ignore those for now

lf-:
> One thing missing here is that it won't work when [downloading logs via --trace-id ](https://github.com/lf-/buck2/blob/236e347e3e6a6d09c1bf2b6490d45c24c289f8d1/app/buck2_client/src/commands/log/options.rs#L129-L130).

It actually *does* work for this, but only because I simply implemented the `curl` URL structure in locally-euclidean. It is a little goofy but it works.

How would you like to proceed with fixing this after landing this PR though? Should buck2 just fetch them from the fbcdn (and equivalent) URLs in ManifoldClient? Should it implement the actual Manifold GET API (in which case I can't help you because I don't know what it is)?

app/buck2_common/src/manifold.rs
lf-: Yeah. I think we should probably just fix the Manifold variant to actually work outside Facebook in that case.

lf-:
Looks like a bogus CI failure (or at least one I have nothing to do with). Surprised it builds internally at Facebook given that I suspect I broke a usage of the API I replaced (`ManifoldClient::new`), but shrug! Whatever works!

lf-:
@JakobDegen would you be able to take a look at this PR?

lf-:
Can you take another look? @iguridi

lf-:
Rebased.

DirectXMan12:
fwiw, i am also interested in having this -- itd likely be pretty useful for deploying for my team as well.

8Keep:
@lf- could you rebase

lf-:
yeah, we have a rebase kicking around on our fork. the real problem is that we just monorepo'd locally-euclidean (and the public one has severe bugs) which means we need to export it once more. i will try throwing an agent at this problem as it's kind of low on my prio list but it may be a bit thorny.

PR #986, opened 2025-05-29
buck2_daemon: automatically max out fd limits on startup

### LINKED ISSUE #419: Many download targets leads to intermittent failures due to file descriptor limits

thoughtpolice:
This automatically increases the fd limit of the buck2d process up to whatever the existing user hard limit is, or a reasonable limit (0x80000) if the soft limit is already good enough. This is not configurable and it always happens without question.

The logic behind this is simple: there is never, ever, ever a scenario where I am thinking to myself: "You know what would be great? If my build system ran out of file descriptors and the build failed". Yet this happens all the time with on buck2 on macOS and (apparently?) my Fedora Asahi laptop, where the default soft limits are very, very low.

When I run `buck2 build`, I am asking it to build things. Buck is good at building things, and often it builds lots of things. The soft limit is a simple guardrail, but only that. And if too many files are open due to something like an excessive number of actions, the reality is that there is very little chance the file descriptor table is actually the limiting resource, versus the underlying machine resources (net/disk/cpu/memory).

Just always set the rlimit to the hard cap. This will be inherited by children including the forkserver, which is completely fine.

Fixes #419. Closes #928. Closes #981.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -219,7 +219,7 @@ memmap2 = "0.5.0"
 memoffset = "0.6.4"
 mimalloc = "0.1.46"
 multimap = "0.8.2"
-nix = "0.22"
+nix = { version = "0.29.0", features = ["dir", "event", "hostname", "inotify", "ioctl", "mman", "mount", "net", "poll", "ptrace", "reboot", "resource", "sched", "signal", "term", "time", "user", "zerocopy"] }
 nom = "7.1.3"
 notify = "=5.0.0"
 num-bigint = "0.4.3"
--- a/app/buck2_common/src/lib.rs
+++ b/app/buck2_common/src/lib.rs
@@ -55,6 +55,7 @@ pub mod memory_tracker;
 pub mod package_boundary;
 pub mod package_listing;
 pub mod pattern;
+pub mod rlimits;
 pub mod scope;
 pub mod sqlite;
 pub mod starlark_profiler;
--- a/app/buck2_common/src/rlimits.rs
+++ b/app/buck2_common/src/rlimits.rs
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under both the MIT license found in the
+ * LICENSE-MIT file in the root directory of this source tree and the Apache
+ * License, Version 2.0 found in the LICENSE-APACHE file in the root directory
+ * of this source tree.
+ */
+
+#[cfg(unix)]
+static REVERT_RLIMITS_TO: std::sync::OnceLock<(nix::libc::rlim_t, nix::libc::rlim_t)> =
+    std::sync::OnceLock::new();
+
+#[cfg(unix)]
+pub fn raise_file_descriptor_limits() -> nix::Result<()> {
+    use nix::sys::resource;
+    use nix::sys::resource::Resource;
+
+    let (soft_limit, hard_limit) = resource::getrlimit(Resource::RLIMIT_NOFILE)?;
+
+    // If the soft limit is already maxxed out or "big enough", don't mess with it.
+    //
+    // The "big enough" value is the default value on Linux.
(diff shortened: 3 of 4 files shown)


JakobDegen:
@thoughtpolice #928 has a comment about this potentially breaking some tools that can't deal with fds above 1024. I don't have the context to know whether that's something we have to worry about in 2025, but anything to refute that from your side?

thoughtpolice:
@JakobDegen Oh, I see, I missed that in the comment (the quote from `systemd.exec`). But you mentioned the 1024 bit I exactly knew it was going to talk about `select(2)`...

Here's the TL;DR: I don't think it matters at all unless we have concrete evidence to the contrary, and I think this patch is still basically fine as a result.

OK, I can see systemd's side of the story and why they suggest this, because from day 1 they had to effectively interoperate with every stupid badly written Unix C program in userspace that was still in use despite being written near the dawn of time, and they had to handle this if they wanted to get anywhere. As a result, they have to be very conservative about things like this and not break things because if they do, it's their fault, even if the downstream program is written badly. So I get that.

But I think in 2025, it does not matter. Here's a good "test" that I would be interested in asking you if you have any information: has Meta *ever* seen a failure like this from `select()` on too large of a file descriptor? One that wasn't easy to fix directly at the source ("use `poll(2)`" which has existed for 30 years)? Clearly your default hard limits must be raised or something for developer environments, right? Has this ever been an actual problem that couldn't be solved by just fixing it in a few one-off spots? I would be extremely surprised if this was some kind of systemic issue, and I suspect the answer is "basically no" given the 100,000 other issues you probably have to fight off. I suspect we'd have already seen fixes if you had, honestly.

But here's my other observation: if you hit this with a tool like Buck, the "solution" every user will use is *identical* to what this patch does: they will scratch their head for 3 seconds, remember `ulimit -n` and then they'll run `ulimit -n reallybignumber` and move on with their life. At that point, all subprocesses of that shell have maxed out limits, said theoretically-still-extant, ancient Unix program would shit themselves anyway from opening too many files. Every user that runs into this issue almost 100% does that exact thing to solve it, so if we had those problems we'd probably hear about them, right?

If anything, I think we can make the exact argument that buck is precisely the kind of program called out in the [`systemd.exec`](https://www.freedesktop.org/software/systemd/man/latest/[systemd.exec](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html).html) man page:

> Nowadays, the hard limit defaults to 524288, a very high value compared to historical defaults. Typically applications should increase their soft limit to the hard limit on their own, if they are OK with working with file descriptors above 1023, i.e. do not use [select(2)](https://man7.org/linux/man-pages/man2/select.2.html).

This exactly describes `buck2` and I think it's reasonable to also require that assumption from the programs it spawns, too. At the very minimum if this is actually a problem, it can be fixed after that, or ideally fixed at the source.

I think the current situation is bad though, and "solution" people use is already equivalent to this patch, so ultimately this patch is a net gain.

thoughtpolice:
@JakobDegen Fixed the comment too, to be more accurate about this case. Thanks for the catch!

JakobDegen:
> has Meta ever seen a failure like this from select() on too large of a file descriptor?

I have no idea :)

> But here's my other observation

This seems sensible

Anyway, if someone really does want to use buck with a compiler from the 80s that struggles with this, it's not like they don't have recourse - wrap the thing in a shell script that sets `ulimit` back to what they want

So yeah, this sounds good to me

PR #991, opened 2025-06-04
Don't create http_download output file until the response begins

Ralith:
Helps reduce the number of file descriptors required for a build, especially in combination with concurrent efforts to limit the number of concurrent HTTP responses being processed.

See also https://github.com/facebook/buck2/pull/922.


Ralith:
CI breakage is not related.

PR #993, opened 2025-06-04
nix: macOS support for development, small tweaks

thoughtpolice:
This makes `nix develop` work correctly on macOS. You should also be able to bootstrap buck2 with buck2 on macOS using this.


cormacrelf:
I think you shouldn't delete all the things necessary for a cargo-based build. Using cargo is still good to have, even if it's easy to set up outside of nix.

PR #995, opened 2025-06-13
bes: add BES and BEP protocol crates

sluongng:
The OSS build has no generated bindings for Bazel's Build Event Protocol
or the Build Event Service, which blocks interoperable build-event
publishing.

Add separate BEP and BES crates, and move shared Google API, Bytestream,
long-running operation, and status definitions out of `re_grpc_proto`
because both remote execution and BES need identical Rust types. The
three commits keep the shared types, BEP schema, and BES service schema
independently reviewable.

The vendored snapshots preserve their upstream Apache license headers.
Provider-specific extensions present in the fork snapshots are omitted,
and generated code is linted using the repository's existing generator
conventions.

This PR adds protocol bindings only. It does not add a BES client, a
Buck-to-BEP converter, invocation UI, or default-on publishing behavior.


sluongng:
Technically, the BES proto is also part of the googleapis repository, but I think it's worth splitting into a separate crate to avoid build cache churn.

cc: @avdv @aherrmann

2FaceS-bit (commented):
@2Facesbit

PR #1004, opened 2025-06-17
`//A/B/C` is a valid name for `//A/B/C:C`

lf-:
> Prefer using the short name when referring to an eponymous target (`//x` instead of `//x:x`). If
> you are in the same package, prefer the local reference (`:x` instead of `//x`).

See: https://bazel.build/build/style-guide#target-naming

This syntax is supported on the command-line, but not in BUCK files. Unfortunately, the only(?) Starlark formatter (buildifier: https://github.com/bazelbuild/buildtools) automatically abbreviates targets on the assumption that this syntax is valid, which causes errors when using Buck2:

    ...
    4: Error coercing "//src/Foo"
    5: Invalid absolute target pattern `//src/Foo` is not allowed
    6: Expected a `:`, a trailing `/...` or the literal `...`.

We can adjust the parsing rules to allow this syntax in more places.

Supercedes https://github.com/facebook/buck2/pull/925

### DIFF
--- a/app/buck2_core/src/pattern/pattern.rs
+++ b/app/buck2_core/src/pattern/pattern.rs
@@ -336,7 +336,7 @@ impl<T: PatternType> ParsedPattern<T> {
             cell_alias_resolver,
             TargetParsingOptions {
                 relative: TargetParsingRel::RequireAbsolute(cell),
-                infer_target: false,
+                infer_target: InferTargetNames::No,
                 strip_package_trailing_slash: false,
             },
             pattern,
@@ -372,7 +372,7 @@ impl<T: PatternType> ParsedPattern<T> {
                     ),
                     Some(target_alias_resolver),
                 ),
-                infer_target: true,
+                infer_target: InferTargetNames::Yes,
                 strip_package_trailing_slash: true,
             },
             pattern,
@@ -382,18 +382,23 @@ impl<T: PatternType> ParsedPattern<T> {
         Self::from_parsed_pattern_with_modifiers(pattern_with_modifiers)
     }
 
+    /// Parse a TargetPattern out, resolving aliases via `cell_resolver` and resolving
+    /// relative targets as permitted by `relative`. If `infer_target` is `Yes`, a target
+    /// name is inferred when none is provided (e.g. `//foo/bar` is interpreted as
+    /// `//foo/bar:bar`).
     pub fn parse_not_relaxed(
         pattern: &str,
         relative: TargetParsingRel<'_>,
         cell_resolver: &CellResolver,
         cell_alias_resolver: &CellAliasResolver,
+        infer_target: InferTargetNames,
     ) -> buck2_error::Result<Self> {
         let pattern_with_modifiers = parse_target_pattern(
             cell_resolver,
             cell_alias_resolver,
             TargetParsingOptions {
                 relative,
-                infer_target: false,
+                infer_target,
                 strip_package_trailing_slash: false,
             },
             pattern,
@@ -476,7 +481,7 @@ impl<T: PatternType> ParsedPatternWithModifiers<T> {
             cell_alias_resolver,
             TargetParsingOptions {
                 relative: TargetParsingRel::RequireAbsolute(cell),
-                infer_target: false,
+                infer_target: InferTargetNames::No,
                 strip_package_trailing_slash: false,
             },
             pattern,
@@ -503,7 +508,7 @@ impl<T: PatternType> ParsedPatternWithModifiers<T> {
                     ),
                     Some(target_alias_resolver),
                 ),
-                infer_target: true,
+                infer_target: InferTargetNames::Yes,
                 strip_package_trailing_slash: true,
             },
             pattern,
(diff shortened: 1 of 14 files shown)


lf-:
From #925 

> Discussed this with @wavewave. We would really like to have this feature because of Bazel compatibility and dev familiarity (fewer divergences with the command line and less repeating yourself) and we decided to keep holding on this patch internally as we think it's quite important to UX, so I would appreciate getting it upstreamed.
>
> I think the solution to the short target syntax being objectionable/confusing when reading buck files is to have the formatter rewrite it to the canonical form, if that's an important property to have internally at Meta.

thoughtpolice:
I missed the original patch and to be honest, not sure how much it means for my vote, but I'd quite like this in my codebases too, I think (FTR, I didn't know Bazel did this.) I rely on the shorthand syntax on the command line extensively because it's nice with the autocomplete support, so I agree it'd be a nice bit of uniformity.

Thanks for resurrecting this! I might try it out in my fork for a few days.

lf-:
Fixed the broken tests and rebased :)

lf-:
@JakobDegen checking in regarding this

JakobDegen:
Alright, so we chatted about this in our team syncs. We're not super thrilled about this - the general vibe is that we want people to understand what the things they write mean, and this probably hurts that. This applies to the command line too fwiw, I don't think we'd feel great about implementing it today if we hadn't done that already.

That being said though, bazel interop seems like an ok reason to add this and if someone does want to use it that's probably up to them. So we're ok with this, under the requirement that this has to be behind an off-by-default buckconfig

thoughtpolice:
I was looking at this PR recently and wondering how feasible it was to add a buckconfig flag for it. I've experimented with it a little and actually think it's nice, but haven't migrated code to use it yet. Unfortunately, it's so deep inside the code I'm not sure there's an easy way to get the buck context there, without just threading a parameter through everything? So that's an annoying lift.

> This applies to the command line too fwiw, I don't think we'd feel great about implementing it today if we hadn't done that already.

Honestly, I find this one to be a useful timesaver feature as a user on the CLI. It makes shell completion much quicker without having to tab through a selection menu on the targets of a given package and pick the "right" one. Normally I'm just autocompleting what are effectively directory paths in the `root//` cell anyway, so just completing the directory is enough. So I guess this is just to say I think it's useful even if some find it a little ugly. :)

> That being said though, bazel interop seems like an ok reason to add this and if someone does want to use it that's probably up to them. So we're ok with this, under the requirement that this has to be behind an off-by-default buckconfig

One problem I realized with this approach is: what happens when you try to use some code that uses the short-syntax, but you want to *prohibit* the short-syntax in your own code? For example, what happens in this scenario where I depend on an external cell `foo//`, and `foo//` needs this feature enabled (because its `BUILD` files are written assuming it is), but `root//` does not want it enabled? Can this scenario be handled correctly? There might be others I haven't thought of.

I wonder if there is a better approach in possibly making the use of this feature some kind of toggle-able warning/lint error instead of strictly enabling/disabling it, and instead it was always enabled all the time. Then users can just turn off the warning instead, but at Meta it would remain a hard error, of some sort?

lf-:
still TODO: address feedback. I need to figure out how to thread the config through.

9999years:
Rebased + added a (default-off) `.buckconfig` setting for this:

```ini
[buck2]
infer_target_names = true
```

app/buck2_core/src/pattern/pattern.rs
```diff
@@ -1059,11 +1064,29 @@ impl<'a> TargetParsingRel<'a> {
     }
 }
 
+/// Whether to infer a target name from the last component of a package path when
+/// a pattern does not name one, making `//foo/bar` equivalent to `//foo/bar:bar`.
+#[derive(Copy, Clone, Dupe, Debug, Eq, PartialEq, Allocative, Pagable)]
+pub enum InferTargetNames {
+    Yes,
+    No,
```
lf-, line 1072: maybe better names?

vaguely thinking "ExpandTargetName", "NoExpand" or something?

app/buck2_server/src/ctx.rs
```diff
@@ -620,13 +621,24 @@ impl DiceUpdater for DiceCommandUpdater<'_, '_> {
...
         let cell_resolver = cells_and_configs.cell_resolver;
 
+        let infer_target_names = InferTargetNames::from(
+            cells_and_configs
+                .root_config
+                .parse::<bool>(BuckconfigKeyRef {
+                    section: "buck2",
+                    property: "infer_target_names",
+                })?
+                .unwrap_or(false),
```
lf-: wait, bool? maybe this should expand to the actual enum here.

app/buck2_server/src/lsp.rs
```diff
@@ -492,16 +495,30 @@ impl<'a> BuckLspContext<'a> {
...
+                let artifact_fs = dice_ctx.get_artifact_fs().await?;
+                let infer_target_names = InferTargetNames::from(
+                    dice_ctx
+                        .parse_legacy_config_property(
+                            artifact_fs.cell_resolver().root_cell(),
+                            BuckconfigKeyRef {
+                                section: "buck2",
+                                property: "infer_target_names",
+                            },
+                        )
```
lf-: :two-of-them:

hm. i dunno what we *should* do about this, but i don't like that the config key is defined in two places.

app/buck2_core/src/pattern/pattern.rs
9999years: Feels redundant, like `InferTargetNames::DontInferTargetNames`.

JakobDegen (approved):

PR #1013, opened 2025-06-25
docs: document how to use fish completion

lf-:
Tested by my coworker. Note that there is a golden test that this breaks, but I cannot figure out how to fix it locally; I think these cannot run outside of Facebook.


alexlian:
Thanks for the submission! Will take care of the golden test generation on our side and turn the crank our side as this is pretty straightforward.

PR #1015, opened 2025-06-26
Allow windows_resource() to take deps

zjturner:
We have a .rc file which does an include of a file in a different directory that is not part of the same package.  In order to support this, allow `windows_resource()` to accept a `deps` argument, and merge this in with the existing preprocessor info.

Also, less critical, but I decided to set the `.res` file as the rule's `default_output`.  It's nice to be able to `buck2 build` the resources target directly for debugging purposes.

### DIFF
--- a/prelude/cxx/windows_resource.bzl
+++ b/prelude/cxx/windows_resource.bzl
@@ -7,26 +7,32 @@
 
 load("@prelude//cxx:cxx_context.bzl", "get_cxx_toolchain_info")
 load("@prelude//cxx:headers.bzl", "cxx_get_regular_cxx_headers_layout")
-load("@prelude//cxx:preprocessor.bzl", "cxx_merge_cpreprocessors", "cxx_private_preprocessor_info")
+load("@prelude//cxx:preprocessor.bzl", "cxx_merge_cpreprocessors", "cxx_private_preprocessor_info", "cxx_inherited_preprocessor_infos")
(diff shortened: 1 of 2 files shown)


prelude/cxx/windows_resource.bzl
```diff
@@ -81,7 +87,7 @@ def windows_resource_impl(ctx: AnalysisContext) -> list[Provider]:
     )
 
     providers = [
-        DefaultInfo(default_output = None),
+        DefaultInfo(default_output = rc_output),
```
alexlian: Noting there's an internal lint failure for us:

buck2/prelude/cxx/windows_resource.bzl:90:38-47: Use of potentially undefined variable `rc_output`

Offhand guess is this is about scoping here.

zjturner: I went ahead and removed the default output.  It's strange to compile multiple rc files, but I guess it's possible which means there is no good choice for the default output so I just removed it.

zjturner: Can these lint checks be made to run on the public GH?

alexlian: > Can these lint checks be made to run on the public GH?

Definitely a goal. This is a lint within the buck2 e2e tests on starlark that I think translates out once we have time to get to TestInfov2, though that's me speaking without strong authority on it.

zjturner:
@scottcao is there anyone internally that can take a look at this?

PR #1016, opened 2025-06-29
Fix: Normalize version string in wheel.py

SyedaAnshrahGillani:
This pull request addresses a TODO in prelude/python/tools/wheel.py to normalize the package version.

  The WheelBuilder class was using the version string as-is, without any normalization. This could
  lead to inconsistencies in the generated wheel metadata.

  This change introduces a normalize_version function that canonicalizes the version string according
  to a simplified subset of the PEP 440 standard. This ensures that the version in the METADATA file
  is always in a consistent and predictable format.

### DIFF
--- a/prelude/python/tools/wheel.py
+++ b/prelude/python/tools/wheel.py
@@ -35,6 +35,12 @@ def normalize_name(name: str) -> str:
     return pep503_normalized_name.replace("-", "_")
 
 
+def normalize_version(version: str) -> str:
+    # Simplified version normalization, based on PEP 440.
+    # This is not a complete implementation, but it handles common cases.
+    return re.sub(r"[-_.]+", ".", version).lower()
+
+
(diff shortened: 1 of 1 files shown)


zsol:
Thanks for this! I think an incomplete implementation of the [spec](https://packaging.python.org/en/latest/specifications/version-specifiers/#normalization) would be OK, but unfortunately this is incorrect, because this implementation replaces all separators with `.`, but that's not a valid normalization of pre-release separators.
Concretely, normalizing `1.1-a1` should result in `1.1a1` but this code produces `1.1.a1`

There's an implementation of the spec in the [`packaging` package](https://github.com/pypa/packaging/blob/14b83e15dbb9caa87c63646ba7808b2b5e460ce6/src/packaging/version.py#L117), which we might be able to vendor.

loren-arthur:
`packaging` is not included in the python standard library and since python_wheel is built using a python_bootstrap_binary rule it does not have access to third-party dependencies

PR #1024, opened 2025-07-07
shim: make fbcode//buck2:buck2_bundle exist for buck2.py

lf-:


JakobDegen:
@lf- I've landed ac80b3b5eb793bebf1f49f4eafdc5e9d1e578893 as an alternative, can you confirm that that's ok?

lf-:
lgtm. only worry is if there's other accidental references in the repo but I'm sure someone will find them eventually.

PR #1025, opened 2025-07-07
Docs: mention where things come from more often

lf-:


Nero5023:
Thank you for the PR. Could you rebase this? It could be too old that it doesn't trigger any our inner CIs.

lf-:
Yep, will do tomorrow morning; looks like we did a mistake with our branch protection on our fork.

lf-:
@Nero5023 now it's rebased, have fun.

PR #1039, opened 2025-07-11
prelude: python: Annotate `PythonToolchainInfo` field types

9999years:
This is challenging for several reasons:
1. The type descriptions used by `provider_field` are very different from the types used by `ctx.attrs`. In particular, there's no way to express "things that can be passed to `cmd_args`."
2. Several attributes are completely unused in the public prelude, so there's no way to figure out what they should be set to.
3. There's no way to annotate `enum` values in `provider_field`s.

### DIFF
--- a/prelude/python/toolchain.bzl
+++ b/prelude/python/toolchain.bzl
@@ -8,6 +8,8 @@
 
 load("@prelude//cxx:cxx_toolchain_types.bzl", "CxxPlatformInfo")
 load("@prelude//utils:platform_flavors_util.bzl", "by_platform")
+load(":manifest.bzl", "ManifestInfo")
+load("@prelude//utils:arglike.bzl", "ArgLike")
 
 # The ways that Python executables handle native linkable dependencies.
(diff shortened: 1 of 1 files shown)


JakobDegen:
> there's no way to express "things that can be passed to cmd_args."

Our workaround for that until we get a type symbol is to use `load("@prelude//utils:arglike.bzl", "ArgLike")`, please do the same here

> There's no way to annotate enum values in provider_fields.

Well, if you can totally pass `provider_field(MyEnum)`, but then you have to actually pass enum values instead of the strings we pass right now - if you want to make the change to require enum values, feel free to do that (maybe a follow-up PR?)

9999years:
> Our workaround for that until we get a type symbol is to use `load("@prelude//utils:arglike.bzl", "ArgLike")`, please do the same here

Done, PTAL

zsol:
This is amazing, thank you!

zsol (approved):

Issue #1026, opened 2025-07-07
python: pytest support is not exported from fbcode


lf-:
I would like to be able to build and test python code with buck2 outside Facebook. This seems to be possible, however, it requires (as used by buck2's `//tests` at least):

```bazel
load("@fbcode_macros//build_defs:native_rules.bzl", "buck_filegroup")
load("@fbcode_macros//build_defs:python_pytest.bzl", "python_pytest")
```

These are not exported anywhere that I know of. They would be very helpful to have available to make meaningful use of buck2, so I would really appreciate them being exported.

This would also possibly allow slightly more of the Sapling test suite to run? Unclear.

PR #1040, opened 2025-07-11
README: Clarify public prelude is a subset of the prelude used at Meta

9999years:
The README (falsely?) claims that the `prelude` directory "is the same code used internally for these builds", but it's only a subset of the actual prelude used internally at Meta, as far as I can tell.

For example, the `PythonToolchainInfo` `bundled_interpreter` and `pex_executor` fields aren't used anywhere in the public prelude. Presumably they're used internally for something at Meta. Also, tools like `@fbcode_macros//build_defs:python_pytest.bzl` are not made public, despite being needed to run tests in this repo (#1026). These ommissions make it challenging to use Buck2 outside of Meta.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,8 +101,9 @@ If these headline features make you interested &mdash; check out the
 Buck2 currently **does not have a stable release tag at this time**. Pre-release
 tags/binaries, and stable tags/binaries, will come at later dates. Despite that,
 it is used extensively inside of Meta on vast amounts of code every day, and
-[buck2-prelude](/prelude/) is the same code used internally for all these
-builds, as well.
+[buck2-prelude](/prelude/) is the same code used internally for all
+these builds, as well. (However, Meta retains large amounts of Starlark
+code which builds on top of the prelude.)
 
 Meta just uses the latest committed `HEAD` version of Buck2 at all times. Your
 mileage may vary &mdash; but at the moment, tracking `HEAD` is ideal for


JakobDegen:
Let me start with the more specific things

> For example, the PythonToolchainInfo bundled_interpreter and pex_executor fields aren't used anywhere in the public prelude. Presumably they're used internally for something at Meta

They're actually not, they seem to just be legacy something, I've submitted a change to delete them

Note though that things like the python toolchain can and often should be used by code outside of the prelude; in general, there's no reason that the toolchain shouldn't be allowed to contain fields that are generally useful even if not used by the prelude directly.

> Also, tools like @fbcode_macros//build_defs:python_pytest.bzl are not made public

Yeah, the tests thing is really unfortunate. Getting tests running in OSS has been a long-standing todo item. It doesn't affect your ability to use the prelude, obviously, but it does affect your ability to test changes, particularly to buck2 core

Now for a couple more general things

We do have parts of the prelude that are not exported; you can easily find them by grepping for `meta_only` in the parts you can see.

More broadly though, as you might imagine we have lots and lots of starlark code that implements all sorts of behaviors (sometimes features, sometimes not so much) that are not available as easily to OSS users because the code has not been open sourced. Untangling the dependencies so that we get the generally useful stuff into the OSS prelude and keep the stuff that actually has to be internal or is not generally useful on our side is something that is a goal, but often a lot of work, blocked on refactors of that same code, etc.

You should think of the non-exported parts of the prelude as being in that same category. It's code for features where we haven't been able to separate the internal stuff from the OSS stuff enough to be able to meaningfully give you access to it. That such code is internally located in the prelude directory is slightly unusual and something we also try to avoid, but whether it happens to be there or literally anywhere else is mostly immaterial.

Besides the part where some of the non-open-sourcable features might be useful to you, I think this should mostly not affect your ability to use the prelude for the parts that are in OSS. If it does, I'd like to hear more about that

As for this PR, I'd be happy to take a version of this that acknowledged that Meta generally has a large amount of starlark code that builds on top of the prelude (as I'm sure others do); as I said, the fact that in exceptional cases some of that code is internally organized into the prelude directory doesn't seem interesting enough to me to be worth acknowledging here.

9999years:
> We do have parts of the prelude that are not exported; you can easily find them by grepping for `meta_only` in the parts you can see.

Thanks, that's very helpful!

> Untangling the dependencies so that we get the generally useful stuff into the OSS prelude and keep the stuff that actually has to be internal or is not generally useful on our side is something that is a goal, but often a lot of work, blocked on refactors of that same code, etc.

Yes, I absolutely understand the dynamics here. I appreciate that so much of this _is_ open sourced!

> Besides the part where some of the non-open-sourcable features might be useful to you, I think this should mostly not affect your ability to use the prelude for the parts that are in OSS. If it does, I'd like to hear more about that

I think that's the crux of it — while the core of Buck2 is open source, a lot of the tooling to make it useful (things like [iBazel](https://github.com/bazelbuild/bazel-watcher), production-grade test runners, tools for integrating Buck2 with semantic analysis tooling like [Glean](https://github.com/facebookincubator/glean), etc.) is still proprietary. While of course I understand that it's inherently challenging to open source components like this (they tend to be tightly integrated with the infrastructure at a given organization), the tight vertical integration between components is exactly what makes the user experience of things like Buck2 so good at places like Meta.

> As for this PR, I'd be happy to take a version of this that acknowledged that Meta generally has a large amount of starlark code that builds on top of the prelude

Do you know how you'd like the verbiage to look for this? Happy to follow your lead here, although I'll push a quick draft in a minute.

JakobDegen:
> Do you know how you'd like the verbiage to look for this? Happy to follow your lead here, although I'll push a quick draft in a minute.

What you have now is fine. Alternatively a more general comment to the effect of what you said about vertical integration might be a nice alternative too

Issue #1046, opened 2025-07-14
docs generator/starlark derive macros: get docs from `starlark_value` docstring if there's no methods defined


lf-:
```rust
#[starlark_value(type = "DynamicAction")]
impl<'v> StarlarkValue<'v> for StarlarkDynamicActions<'v> {
    // Used to add type documentation to the generated documentation
    fn get_methods() -> Option<&'static Methods> {
        static RES: MethodsStatic = MethodsStatic::new();
        RES.methods(dynamic_actions_methods)
    }
}

impl<'v> AllocValue<'v> for StarlarkDynamicActions<'v> {
    fn alloc_value(self, heap: &'v Heap) -> Value<'v> {
        // No need to freeze: we unpack contents of this struct
        // in `ctx.actions.dynamic_output` call.
        heap.alloc_complex_no_freeze(self)
    }
}

/// Opaque thunk type returned from calling the returned value of a `dynamic_actions` or
/// `bxl.dynamic_actions` rule invocation.
///
/// Can be passed to
/// [`AnalysisActions.dynamic_output_new`](../AnalysisActions#analysisactionsdynamic_output_new) to
/// extract its contents.
///
/// Be aware that the context argument of the called impl function differs between
/// [`dynamic_actions`](../#dynamic_actions) where it is [`actions: AnalysisActions`](../AnalysisActions)
/// and [`bxl.dynamic_actions`](../../bxl/#dynamic_actions)
/// where it is [`bxl_ctx: bxl.Context`](../../bxl/Context).
#[starlark_module]
fn dynamic_actions_methods(builder: &mut MethodsBuilder) {}
```

Currently it's necessary to write boilerplate `get_methods` implementations just to get docs for opaque types. This could be made easier in the generator by taking the doc string from the `starlark_value` impl and obviating the necessity for this hack.

PR #1047, opened 2025-07-14
docs: add a bunch of docs to dynamic outputs

### LINKED ISSUE #1038: docs: Document what to do with a DynamicValue
Currently the DynamicValue docs page is totally empty: https://buck2.build/docs/api/build/DynamicValue/

It should link to the place from which they are obtained: https://buck2.build/docs/api/build/AnalysisActions/#analysisactionsdynamic_output_new
As well as the https://buck2.build/docs/rule_authors/dynamic_dependencies/ page

And it should say how to do things with them in bxl. In particular it should probably mention https://buck2.build/docs/api/build/#dynamic_actions.

Looks like https://buck2.build/docs/api/build/dynattrs/DynamicAttrType/ also is blank.

Someone at work RE'd this system already so I have private examples that I can use to write these docs. I will probably do it eventually.

This is probably "good first issue" material (also if anyone is able to give me triage rights on this repo, I would be happy to triage my own bugs; I know I am filing a lot of them).

lf-:
There's a bunch of boilerplate due to
https://github.com/facebook/buck2/issues/1046.

Fixes: https://github.com/facebook/buck2/issues/1038


app/buck2_action_impl/src/context/dynamic_output.rs
```diff
@@ -202,6 +202,26 @@ pub(crate) fn analysis_actions_methods_dynamic_output(methods: &mut MethodsBuild
     /// New version of `dynamic_output`.
     ///
     /// This is work in progress, and will eventually replace the old `dynamic_output`.
+    ///
+    /// It takes as input a `DynamicActions` obtained from calling the result of
+    /// [`dynamic_actions()`](../#dynamic_actions) or
+    /// [`bxl.dynamic_actions()`](../../bxl/#dynamic_actions) and returns an opaque
+    /// [`DynamicValue`](../DynamicValue) thunk containing the result of the call, which can be used
```
JakobDegen: The way this is written would seem to imply that the dynamic lambda is evaluated immediately upon invoking `actions.dynamic_output_new`, but that's obviously not the case - the actual evaluation of that lambda is performed later (if at all). I think we should change this phrasing to clarify that

As a useful piece of context for writing dynamic_output documentation - I've had to explain how this works and why it's useful a number of times internally at Meta at this point and there have consistently been two things that were extremely important to making it "click": 1) Dynamic lambda evaluations are an "analysis," in the sense that they define artifacts and actions, and 2) these evaluations really happen within the action graph, not within the analysis graph

Besides that, I think that starting by talking about `DynamicValue` in the top-level documentation for `dynamic_output` is not the best move - `DynamicValue` is a thing that I would not expect the majority of uses of `dynamic_output` to require. I think it'd be more appropriate to start by talking about `dynamic_output`s ability to bind output artifacts, the fact that it's an analysis, the fact that it can read the contents of other artifacts, and only then mention `DynamicValue`

app/buck2_action_impl/src/dynamic/dynamic_actions.rs
```diff
@@ -54,3 +64,23 @@ impl<'v> AllocValue<'v> for StarlarkDynamicActions<'v> {
...
+/// It is not possible to extract structured values from inside a `DynamicValue` into anything
+/// but another dynamic action. However, dynamic actions may have *output artifact* arguments of
+/// type [`dynattrs.output()`](../dynattrs/#output) passed into them, which can then be *bound*
+/// by the dynamic actions and referenced from outside as if they are normal artifacts; when those
+/// output artifacts are demanded by the build process, the dynamic action will be executed.
+///
+/// Be aware that the context argument of the called impl function differs between
+/// [`dynamic_actions`](../#dynamic_actions) where it is [`actions: AnalysisActions`](../AnalysisActions)
+/// and [`bxl.dynamic_actions`](../../bxl/#dynamic_actions)
+/// where it is [`bxl_ctx: bxl.Context`](../../bxl/Context).
```
JakobDegen: I think this belongs on the docs to `dynamic_actions`/`bxl.dynamic_actions` moreso than here

app/buck2_action_impl/src/dynamic/dynamic_actions.rs
```diff
@@ -54,3 +64,23 @@ impl<'v> AllocValue<'v> for StarlarkDynamicActions<'v> {
         heap.alloc_complex_no_freeze(self)
     }
 }
+
+/// Opaque thunk type returned from calling the returned value of a `dynamic_actions` or
+/// `bxl.dynamic_actions` rule invocation.
```
JakobDegen:
```suggestion
/// `bxl.dynamic_actions` invocation.
```

This is saying a lot more than I would expect it to though - this thing is semantically just a tuple of the `DynamicActionCallable` and the params passed into it, all the interesting documentation about those should go onto one of `dynamic_action` or `dynamic_output_new` I think

PR #1050, opened 2025-07-16
Add 3 new options to the compilation_database bxl

zjturner:
This adds three new options to the compilation db bxl.

1. An option for excluding header files.  These don't really make sense in a compilation db, and for our use case we would prefer them to be removed.
2. An option for adding the target name to the compile commands json entry.  This is a non-standard field that is not part of the compilation db spec, but nevertheless it would be useful to us for certain use cases, and there should be no harm in adding it in an opt-in fashion.
3. An option to automatically recursive through all dependent targets.

All options default to their original values so as not to change any existing behavior unless you're opting in.

### DIFF
--- a/prelude/cxx/tools/compilation_database.bxl
+++ b/prelude/cxx/tools/compilation_database.bxl
@@ -1,9 +1,10 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 #
-# This source code is licensed under both the MIT license found in the
-# LICENSE-MIT file in the root directory of this source tree and the Apache
+# This source code is dual-licensed under either the MIT license found in the
+# LICENSE-MIT file in the root directory of this source tree or the Apache
 # License, Version 2.0 found in the LICENSE-APACHE file in the root directory
-# of this source tree.
+# of this source tree. You may select, at your option, one of the
+# above-listed licenses.
 
 load("@prelude//cxx:comp_db.bzl", "CxxCompilationDbInfo")
 load("@prelude//cxx:compile_types.bzl", "CxxSrcCompileCommand")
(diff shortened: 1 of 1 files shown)


prelude/cxx/tools/compilation_database.bxl
```diff
@@ -48,5 +60,17 @@ generate = bxl_main(
...
+        "include-headers": cli_args.bool(
+            default = True,
+            doc = "Include header file compilation db actions"
+        ),
+        "add-target-name": cli_args.bool(
+            default = False,
+            doc = "Include the name of the target in the compilation database entries (non-standard)"
+        ),
+        "recursive": cli_args.bool(
+            default = False,
```
cbarrete, line 72: This should probably default to `True`, or even not be a configuration option IMO.

I hadn't even considered that case because I always generate compile commands for the whole repo, but without this, a more selective target expression would lead to a compilation database where going to definition results in a broken experience.

For cases where tools only look at a single translation unit (e.g. clang-tidy), one should use the compile commands subtargets instead of this BXL script anyway.

cbarrete (commented):
>  These don't really make sense in a compilation db

They absolutely do, they are useful for clangd to work in headers. Without that, you have to first open a source file that happens to include that header for analysis to work properly. And if you're dealing with a lot of header-only libraries, that's extremely cumbersome.

scottcao:
@zjturner note that for backwards compatibility, I changed the PR such that it always created a comp db output even when it's empty, which matches original behavior. We can make it not do this via an additional flag if you want.

zjturner:
@scottcao I have no problem with that modification.  Thank you!

Issue #877, opened 2025-03-13
Help getting precompiled header support into the prelude


zjturner:
I've talked to a few people in the past about this, but I've never gotten enough concrete detail or guidance to be able to get this to work.  Possible I'm not smart enough :)  The TL;DR is that we have people for whom remote execution will never be feasible, and so for this subset of people, we still want to make local builds as fast as possible, and that means precompiled headers which yield anywhere from 1.5x-2x speedup on builds.

I'll explain the semantics of how the build has to work, and I'm hoping someone can offer me some very specific pointers about where in the prelude I should be looking and what kind of interactions I need to be aware of.  I'll also simplify the implementation a little so we don't have to worry about the "easy" parts.  I'm also going to keep the discussion MSVC-centric because the difficult parts of the implementation shouldn't change much when talking about clang or some other compiler.

You have a `cxx_library` or `cxx_binary` (we'll call it `L`) whose cpp files are going to get compiled with some set `F` of compiler flags.  Let's assume that for this library you don't have any file-specific arguments, so every file will get compiled exactly the same way.  Before compiling any file in the library, you must _first_ compile a different file.  Let's call it `pch.cpp`.  It must have *exactly the same flags* as are in the set `F` that we talked about earlier, but add in one additional flag to tell the compiler to create a pch.  (there's a little bit of flexibility here, and this isn't strictly true, but let's just say it is).   

This produces two outputs, a `.obj` and a `.pch` file.  Now, when compiling `L`, every cpp file must accept the `.pch` file on its command line using an option that tells the compiler to consume the PCH.

Importantly, the output of the precompiled header build is an input to every individual compilation, meaning it's not enough to just have a cxx_library that I pass the "create" flag into, and have it be a dep of the cxx_library that consumes the PCH, because we need to guarnatee that the PCH is materialized first and is an input to the compile actions on the consumtion library.

I can think of several different options:

```
// option 1, hackish.  no rule modifications, just stick the flags on.

_flags = ["/A", "/B", ...]

cxx_library(name="pch", srcs=["pch.cpp"], flags=_flags + ["/Yc"])
cxx_library(name="L", srcs=["a.cpp", "b.cpp"], flags=_flags + ["/Yu$(location :pch)"])
```

```
// option 2.  first class rule knowledge

// take all of th same attrs as cxx_library(), and internally call _cxx_library_impl with the right args so that pch.cpp generates the pch output file.
cxx_precompiled_header(name="pch", src="pch.cpp")

[...]

PR #1056, opened 2025-08-01
adding cxx precompiled headers implementation to prelude

KlotzAndrew:
This change updates the precompiled_header rule to compile a PCH and updates both cxx_library and cxx_executable to compile and link with it. PCH only work for local builds, and mentioned here (https://github.com/facebook/buck2/issues/877) there are potentially 1.5x-2x speedups on builds for using PCH. Noted here, the current precompiled_header implementation doesn't really work as expected (https://github.com/facebook/buck2/issues/368)

Some of the decisions I could use feedback on:

 - The current approach maintains backwards compatibility with the old precompiled_header rule in case anyone was depending on how it was implemented. Upgrading to the new version requires `compile_pch_file=True` and `pch_language=<any clanguage extension>`. Alternatively this could be named something like `precompiled_header_v2` without interfering with the old implementation

 - The PCH and consumer both need to have exactly matching flags. I was hoping to find a better way to match up the PCH rule with the consumer because this matching needs to be done by the developer

 - This re-uses the `cxx_library` internals of `cxx_library_parameterized`, which causes a bunch of shuffling of the input attrs so both the `cxx_library` and `precompiled_header` rule can receive the same inputs.

 - The PCH rule needs a pch_language to be specified. The cxx internals infers clanguage by file extension, which we cannot do with a `.pch` extension so we explicitly pass it in. An alternative could be for the PCH rule to return a provider with all flavors like how pic/non-pic currently work and have the consumer pluck out the correct one to use. This direction looks like it will involve more changes potentially around `MergedLinkInfo` and I'm not sure if we want to take that approach

example usage:
```bash
cxx_precompiled_header(
    name = "my_pch",
    src = "my_pch.pch",
    deps = [],

    # enabling new PCH flow needs these two fields
    compile_pch_file = True,
    pch_language = ".cpp",
)

cxx_library(
    name = "foo",
    srcs = ["foo.cpp"],
    precompiled_header = ":my_pch",
    deps = [],
)
```

### DIFF
--- a/prelude/cxx/compile.bzl
+++ b/prelude/cxx/compile.bzl
@@ -360,15 +360,24 @@ def _compile_single_cxx(
     pch_object_output = None
     if src_compile_cmd.src.extension == ".cu":
         output_args = None
-    elif compile_pch and src_compile_cmd.cxx_compile_cmd.compiler_type == "windows":
-        pch_object_output = ctx.actions.declare_output(
-            folder_name,
-            "{}.pch.o".format(filename_base),
-        )
-        output_args = [
-            cmd_args(object.as_output(), format="/Fp{}"),
-            cmd_args(pch_object_output.as_output(), format="/Fo{}")
-        ]
+    elif compile_pch:
+        if src_compile_cmd.cxx_compile_cmd.compiler_type == "windows":
+            pch_object_output = ctx.actions.declare_output(
+                folder_name,
+                "{}.pch.o".format(filename_base),
+            )
+            output_args = [
+                cmd_args(object.as_output(), format="/Fp{}"),
+                cmd_args(pch_object_output.as_output(), format="/Fo{}"),
+                cmd_args(compile_pch.basename_src, format="/Yc{}")
+            ]
+        else:
+            output_args = [
+                get_output_flags(compiler_type, object),
+                "-Xclang", "-emit-pch",
+                "-Xclang", "-fno-pch-timestamp",
+                "-fpch-instantiate-templates",
+            ]
     else:
         output_args = get_output_flags(compiler_type, object)
 
@@ -380,16 +389,6 @@ def _compile_single_cxx(
         use_header_units = use_header_units,
         output_args = output_args,
     )
-
-    if compile_pch:
-        if src_compile_cmd.cxx_compile_cmd.compiler_type == "windows":
-            cmd.add(cmd_args(compile_pch.basename_src, format="/Yc{}"))
-        elif src_compile_cmd.cxx_compile_cmd.compiler_type == "clang":
-            cmd.add(
-                "-Xclang", "-emit-pch",
-                "-Xclang", "-fno-pch-timestamp",
-                "-fpch-instantiate-templates",
-            )
 
     if precompiled_header and precompiled_header.compiled:
         pch_info = ctx.attrs.precompiled_header[DefaultInfo].sub_targets["pch"]
--- a/prelude/cxx/compile.bzl
+++ b/prelude/cxx/compile.bzl
@@ -267,9 +267,6 @@ def create_compile_cmds(
         if cxx_compile_cmd.compiler_type != "nasm":
             src_args.append("-c")
 
-        if compile_pch and cxx_compile_cmd.compiler_type == "windows":
-            cxx_compile_cmd.base_compile_cmd.add("-I", compile_pch.namespace)
-
         src_args.append(src.file)
 
         src_compile_command = CxxSrcCompileCommand(
@@ -369,7 +366,7 @@ def _compile_single_cxx(
             output_args = [
                 cmd_args(object.as_output(), format="/Fp{}"),
                 cmd_args(pch_object_output.as_output(), format="/Fo{}"),
-                cmd_args(compile_pch.basename_src, format="/Yc{}")
+                cmd_args(compile_pch.path, format="/Yc{}")
             ]
         else:
             output_args = [
@@ -1650,7 +1647,7 @@ def _get_use_pch_args(
     elif src_compile_cmd.cxx_compile_cmd.compiler_type in ["clang"]:
         pch_args.add("-Xclang", "-include-pch", "-Xclang", compile_with_pch.header)
     else:
-        print("Warning: Unsupported compiler type for precompiled header usage: {}".format(
+        fail("Warning: Unsupported compiler type for precompiled header usage: {}".format(
             src_compile_cmd.cxx_compile_cmd.compiler_type,
         ))
 
--- a/prelude/cxx/compile.bzl
+++ b/prelude/cxx/compile.bzl
@@ -57,6 +57,7 @@ load(
     ":headers.bzl",
     "CHeader",
     "CPrecompiledHeaderInfo",
+    "CxxPrecompiledHeader",
     "add_headers_dep_files",
 )
 load(":platform.bzl", "cxx_by_platform")
@@ -173,14 +174,19 @@ def create_compile_cmds(
         impl_params: CxxRuleConstructorParams,
         own_preprocessors: list[CPreprocessor],
         inherited_preprocessor_infos: list[CPreprocessorInfo],
-        add_coverage_instrumentation_compiler_flags: bool) -> CxxCompileCommandOutput:
+        add_coverage_instrumentation_compiler_flags: bool,
+        compile_pch: CxxPrecompiledHeader | None = None,
+    ) -> CxxCompileCommandOutput:
     """
     Forms the CxxSrcCompileCommand to use for each source file based on it's extension
     and optional source file flags. Returns CxxCompileCommandOutput containing an array
     of the generated compile commands and argsfile output.
     """
 
-    srcs_extensions = collect_extensions(impl_params.srcs)
+    if compile_pch:
+        srcs_extensions = set([CxxExtension(compile_pch.clanguage)])
+    else:
+        srcs_extensions = collect_extensions(impl_params.srcs)
     extension_for_plain_headers = detect_source_extension_for_plain_headers(srcs_extensions, impl_params.rule_type)
 
     srcs_with_flags = []  # type: [CxxSrcWithFlags]
@@ -223,8 +229,10 @@ def create_compile_cmds(
     cxx_compile_cmd_by_ext = {}  # type: dict[CxxExtension, CxxCompileCommand]
     argsfile_by_ext = {}  # type: dict[str, CompileArgsfile]
     xcode_argsfile_by_ext = {}  # type: dict[str, CompileArgsfile]
-
-    src_extensions = collect_source_extensions(srcs_with_flags, extension_for_plain_headers)
+    if compile_pch:
+        src_extensions = [CxxExtension(compile_pch.clanguage)]
+    else:
+        src_extensions = collect_source_extensions(srcs_with_flags, extension_for_plain_headers)
 
     # Deduplicate shared arguments to save memory. If we compile multiple files
     # of the same extension they will have some of the same flags. Save on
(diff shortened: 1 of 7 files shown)


prelude/cxx/compile.bzl
```diff
@@ -426,6 +481,7 @@ def _compile_single_cxx(
             allow_dep_file_cache_upload = False,
             error_handler = src_compile_cmd.error_handler,
             outputs_for_error_handler = outputs_for_error_handler,
+            local_only = is_producing_compiled_pch or is_consuming_compiled_pch,
```
akrieger, line 517: It would be nice if there were a way to tag an artifact as 'if used as an input to an action the action must run local'.

prelude/cxx/cxx_library.bzl
```diff
@@ -1361,16 +1447,18 @@ def _form_library_outputs(
                 pic = pic,
                 optimized = False,
                 debuggable = True,
+                pch = False,
                 stripped = False,
             )
 
             # Only generate an archive if we have objects to include
-            if lib_compile_output.objects:
+            if lib_compile_output.objects or (compile_pch and lib_compile_output.pch_object_output):
```
akrieger, line 1552: Why do we need to wrap the pch .o into an archive? The pch is not a linker input ever iirc. https://learn.microsoft.com/en-us/cpp/build/creating-precompiled-header-files?view=msvc-170#pch-files-in-the-build-process

Edit: why do we even have a .pch.o ?

zjturner: Despite what that page says, it is necessary to link in the pch object file.  Here's an example I just created to demonstrate:

```
D:\src\pchtest>CL.exe /c /ZI /MDd /Ycpch.h /Fpstaticlib.pch /Fdstaticlib.pdb /EHsc- /Gd /TP /FC pch.cpp
pch.cpp

D:\src\pchtest>cl.exe /c /ZI /MDd /Yupch.h /Fpstaticlib.pch /Fdstaticlib.pdb /EHsc- /external:W3 /Gd /TP /FC staticlib.cpp
staticlib.cpp

D:\src\pchtest>lib.exe /OUT:staticlib.lib /NOLOGO /MACHINE:X64 staticlib.obj

D:\src\pchtest>cl.exe /c /ZI /MDd /Fdmain.pdb /EHsc- /Gd /TP /FC main.cpp
main.cpp

D:\src\pchtest>link.exe /DEBUG:FULL /MACHINE:X64 main.obj staticlib.lib
staticlib.lib(staticlib.obj) : error LNK2011: precompiled object not linked in; image may not run
main.exe : fatal error LNK1120: 1 unresolved externals
```

Attaching a zip file that you can use to reproduce.

[pchtest.zip](https://github.com/user-attachments/files/22288688/pchtest.zip)

zjturner: Here's an MSDN page about it:

https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-error-lnk2011?view=msvc-170

akrieger: Le sigh.

prelude/cxx/compile.bzl
KlotzAndrew: that would definitely come in handy here

prelude/cxx/headers.bzl
```diff
@@ -113,9 +113,22 @@ CxxHeadersLayout = record(
...
+    clanguage = field(str),
+)
+
 CPrecompiledHeaderInfo = provider(fields = {
     # Actual precompiled header ready to be used during compilation.
     "header": Artifact,
+    "basename": provider_field(typing.Any, default = None),
+    "path": provider_field(typing.Any, default = None),
+    "compiled": provider_field(bool, default = False),
+    "clanguage": provider_field(str | None, default = None),
```
akrieger, line 131: It bothers me we're mixing Cxx and C but it's preexisting badness. Just saying. I see up above also eg. CHeader too.

prelude/cxx/cxx_library.bzl
```diff
@@ -301,6 +307,7 @@ _CxxAllLibraryOutputs = record(
 _CxxLibraryCompileOutput = record(
     # object files
     objects = field(list[Artifact]),
+    pch_object_output = field(list[Artifact]),
```
akrieger, line 315:
```suggestion
    # linkable object files from pch compilation
    pch_object_output = field(list[Artifact]),
```

prelude/cxx/cxx_library.bzl
```diff
@@ -395,6 +407,40 @@ def cxx_library_parameterized(ctx: AnalysisContext, impl_params: CxxRuleConstruc
...
+
+        compiler_type = get_compiler_type(ctx, CxxExtension(pch_clanguage))
+
+        pch_artifact = ctx.attrs.srcs[0]
+        pch_header = cxx_attr_precompiled_headers(ctx, impl_params.headers_layout)
+        basename = "{}{}".format(pch_artifact.basename, pch_clanguage)
+        path = paths.join(pch_header.namespace, pch_header.name)
+
+        if compiler_type == "windows":
+            src_file = ctx.actions.declare_output("__pch__win/__src__", basename)
```
akrieger: Long path limits tend to crop up frequently so I'd use something shorter here. This is already unique'd by the target path, so can't this just be 'pch.h' or something? We use something equivalent in header units (`precompiledN.h` for N from 0...).
```suggestion
            src_file = ctx.actions.declare_output("pch.h")
```

prelude/cxx/cxx_library.bzl
```diff
@@ -667,7 +735,7 @@ def cxx_library_parameterized(ctx: AnalysisContext, impl_params: CxxRuleConstruc
     # Propagate link info provider.
     if impl_params.generate_providers.merged_native_link_info or impl_params.generate_providers.template_placeholders:
         # Gather link inputs.
-        inherited_non_exported_link = cxx_inherited_link_info(non_exported_deps)
+        inherited_non_exported_link = cxx_inherited_link_info(non_exported_deps + filter(None, [ctx.attrs.precompiled_header if compiler_type != "clang" else None]))
```
akrieger: Why only if not clang? Please comment/document.

prelude/cxx/cxx_library.bzl
```diff
@@ -411,7 +457,7 @@ def cxx_library_parameterized(ctx: AnalysisContext, impl_params: CxxRuleConstruc
     own_exported_preprocessors = [own_exported_preprocessor_info]
 
     inherited_non_exported_preprocessor_infos = cxx_inherited_preprocessor_infos(
-        non_exported_deps + filter(None, [ctx.attrs.precompiled_header]),
+        non_exported_deps + filter(None, [ctx.attrs.precompiled_header if precompiled_header and not precompiled_header.compiled else None]),
```
akrieger, line 467:
```suggestion
        # Legacy precompiled_header implementation is not exported. Proper precompiled headers can have exported linkables.
        non_exported_deps + filter(None, [ctx.attrs.precompiled_header if precompiled_header and not precompiled_header.compiled else None]),
```

prelude/cxx/cxx_library.bzl
```diff
@@ -1295,10 +1374,16 @@ def _form_library_outputs(
...
             stripped: bool) -> (CxxLibraryOutput | None, LinkInfo | None):
         if not compile_output:
             return (None, None)
-        objects = compile_output.stripped_objects if stripped else compile_output.objects
+
+        if pch:
+            objects = compile_output.pch_object_output
+        else:
+            objects = compile_output.stripped_objects if stripped else compile_output.objects
+
```
akrieger, line 1411: Can we just avoid Yet Another Bool Arg here?
```suggestion
        objects = compile_output.pch_object_output if compile_output.pch_object_output else compile_output.stripped_objects if stripped else compile_output.objects
```

prelude/cxx/compile.bzl
```diff
@@ -339,6 +372,27 @@ def _compile_single_cxx(
         output_args = output_args,
     )
 
+    if compile_pch:
+        if src_compile_cmd.cxx_compile_cmd.compiler_type == "windows":
+            cmd.add(cmd_args(compile_pch.basename_src, format="/Yc{}"))
```
akrieger: Can this just be added to `output_args` to avoid having to replicate the same conditional structure twice?

prelude/cxx/compile.bzl
```diff
@@ -339,6 +372,27 @@ def _compile_single_cxx(
...
+        elif src_compile_cmd.cxx_compile_cmd.compiler_type == "clang":
+            cmd.add(
+                "-Xclang", "-emit-pch",
+                "-Xclang", "-fno-pch-timestamp",
+                "-fpch-instantiate-templates",
```
akrieger: Same?

akrieger (approved):
The amount of if/else'ing doesn't sit well with me, but it does help minimize how much large green code is written. The fact that the pch output also has to create a linkable input is a large part of why that's necessary. Just another thing to take note of if we ever come back and reimplement the cxx stack.

You may need to rebase this to resolve incoming conflicts.

prelude/cxx/cxx_library.bzl
KlotzAndrew: makes sense, updated!

prelude/cxx/cxx_library.bzl
KlotzAndrew: inverted and switched to windows, also added a comment about windows needing the linkables

prelude/cxx/compile.bzl
KlotzAndrew: good thinking, updated!

prelude/cxx/compile.bzl
KlotzAndrew: yep, updated!

KlotzAndrew:
> The amount of if/else'ing doesn't sit well with me, but it does help minimize how much large green code is written. The fact that the pch output also has to create a linkable input is a large part of why that's necessary. Just another thing to take note of if we ever come back and reimplement the cxx stack.
> 
> You may need to rebase this to resolve incoming conflicts.

Rebased, happy to make any other adjustments as well. I had been trying to find other ways to approach this, but its re-using so much of the existing cxx library/executable internals

prelude/cxx/compile.bzl
```diff
@@ -1568,3 +1633,25 @@ def _generate_base_compile_command(
...
+    if src_compile_cmd.cxx_compile_cmd.compiler_type in ["windows"]:
+        pch_args.add(cmd_args(compile_with_pch.basename, format="/Yu{}"))
+        pch_args.add(cmd_args(compile_with_pch.basename, format="/FI{}"))
+        pch_args.add(cmd_args(compile_with_pch.header, format="/Fp{}"))
+    elif src_compile_cmd.cxx_compile_cmd.compiler_type in ["clang"]:
+        pch_args.add("-Xclang", "-include-pch", "-Xclang", compile_with_pch.header)
+    else:
+        print("Warning: Unsupported compiler type for precompiled header usage: {}".format(
+            src_compile_cmd.cxx_compile_cmd.compiler_type,
+        ))
```
scottcao: Can we turn this into an error rather than a warning? Warnings may be missed and in general, I think error would be more maintainable for other use cases in general.

scottcao (commented):
Just a nit

KlotzAndrew:
> Just a nit

updated!

scottcao (approved):

prelude/cxx/cxx_library.bzl
```diff
@@ -399,6 +412,41 @@ def cxx_library_parameterized(ctx: AnalysisContext, impl_params: CxxRuleConstruc
     non_exported_deps = cxx_attr_deps(ctx)
     exported_deps = cxx_attr_exported_deps(ctx)
 
+    precompiled_header = ctx.attrs.precompiled_header[CPrecompiledHeaderInfo] if ctx.attrs.precompiled_header else None
+
+    compile_pch = None
+    compiler_type = None
```
scottcao, line 418: One more thing: as I was merging this, I saw a lint triggered on this line because `compiler_type` is an unused variable. Would you mind updating the PR to remove it? I really need to get the starlark lints run on OSS code by default so this would be caught earlier, but you should be able to repro this via running `buck2 starlark lint prelude/cxx/cxx_library.bzl`.

scottcao (changes requested):

prelude/cxx/cxx_library.bzl
scottcao: Actually never mind, it's just one lint and it's easy enough for me to update it myself. But just FYI `compiler_type` is getting removed.

scottcao (approved):

Issue #1029, opened 2025-07-08
buck2 completion zsh crashes when built in debug mode from source


lf-:
Conclusion from this: there needs to be a CI check that runs `buck2 completion zsh` in debug mode, apparently.

Steps to reproduce:

```
$ cargo run --bin buck2 -- completion zsh
   Compiling buck2_bxl v0.1.0 (/Users/jade/co/buck2/app/buck2_bxl)
   Compiling buck2 v0.1.0 (/Users/jade/co/buck2/app/buck2)
    Finished `dev` profile [optimized + debuginfo] target(s) in 7.17s
     Running `target/debug/buck2 completion zsh`
#compdef buck2 buck
# Copyright (c) Meta Platforms, Inc. and affiliates.
#
# This source code is licensed under both the MIT license found in the
# LICENSE-MIT file in the root directory of this source tree and the Apache
# License, Version 2.0 found in the LICENSE-APACHE file in the root directory
# of this source tree.

# @generated by `target/debug/buck2 completion zsh`

# clap_complete generated content BEGINS

thread 'buck2-main' panicked at /Users/jade/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.37/src/builder/debug_asserts.rs:86:13:
Command chrome-trace: Argument names must be unique, but 'path' is in use by more than one argument or group
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/920d95eaf23d7eb6b415d09868e4f793024fa604/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/920d95eaf23d7eb6b415d09868e4f793024fa604/library/core/src/panicking.rs:75:14
   2: clap_builder::builder::debug_asserts::assert_app
             at /Users/jade/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.37/src/builder/debug_asserts.rs:86:13
   3: clap_builder::builder::command::Command::_build_self
             at /Users/jade/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.37/src/builder/command.rs:4404:13
   4: clap_builder::builder::command::Command::_build_recursive
             at /Users/jade/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.37/src/builder/command.rs:4307:9
   5: clap_builder::builder::command::Command::_build_recursive
[...]

PR #1058, opened 2025-08-04
Use unique clap argument IDs

bvinc:
Running `buck2 debug chrome-trace` in debug mode would panic as clap has a debug assert to check for duplicate argument IDs.  It is my understanding that duplicating argument IDs can cause bugs in release mode, even if they are under different subcommands.

### DIFF
--- a/app/buck2_cmd_debug_client/src/chrome_trace.rs
+++ b/app/buck2_cmd_debug_client/src/chrome_trace.rs
@@ -103,13 +103,13 @@ pub struct ChromeTraceCommand {
 
     /// The path to read the event log from.
     #[clap(
-        long,
+        long = "path",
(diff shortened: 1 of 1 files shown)


lf-:
this fixes #1029, please add a note in the commit message :)

scottcao:
I tried merging this internally but unfortunately it does look like we have existing usages using `--path`. That means migration internally, we need to use `soft_error` mechanism to detect all usages dynamically, migrate call sites to new flag, and then turn it into an error. If you don't mind putting up a PR for soft error, I can land it and do the migration and then land this PR on top.

bvinc:
Ah, it's been a long time since I've looked at this.  But my intention with this PR was not to change anything user facing.  Internally, clap has an "argument ID", and THAT is what I intended on changing.

I've updated to PR to include `long = "path"`, so that it can still be used like `--path`, but it also has an internal argument ID that is derived from the variable name, which is now unique.

PR #1059, opened 2025-08-05
prelude/go: Fix wrong starlark load paths for OSS

sluongng:
The issue was introduced in
Differential Revision: D79265796
Commit: ff865194a85b093d09b1e51fd0707522697100ee


sluongng:
cc: @podtserkovskiy @echistyakov

podtserkovskiy:
Thank you for fixing it

PR #1061, opened 2025-08-05
Reduce name collisions in vsgo-generated projects

zjturner:
This addresses two issues with non-uniqueness of solution item names.

1. `foo//:target` and `bar//:target`.  Previous work addressed this at the filesystem level by generating the projects into unique locations, but it didn't completely solve it, it needs to be disambiguated in the solution folder hierarchy as well.
2. `foo//bar:bar` and `foo//:bar`.  In this case you would have both a solution folder and a project with the same name, which would trigger non-uniqueness.

This PR implements two improvements:

* We always include the cell name in the generated solution folder hierarchy.
* Solution folders always are surrounded with parentheses, and project names are never surrounded with parentheses.  This actually fixes some confusing (but not otherwise problematic) behavior from before, where solution folders would sometimes have parentheses and sometimes not, which would create unusual sorting behavior and was hard for users to understand the reason for the parens.

### DIFF
--- a/prelude/ide_integrations/visual_studio/gen_sln.bxl
+++ b/prelude/ide_integrations/visual_studio/gen_sln.bxl
@@ -18,11 +18,20 @@ EndProject""".format(
         project_guid = project_guid,
     )
 
-def _gen_folder(folder_path, guid, project_names):
+def _gen_folder(folder_path, guid):
     project_name = basename(folder_path)
-    if project_name in project_names:
-        # "()" quote folder name to avoid prompt: The solution already contains an item named 'xxx'.
-        project_name = "(" + project_name + ")"
+
+    # Imagine you have `foo//bar:bar` and `foo//:bar`.  This means that underneath `foo`, you will have
+    # both a Solution Folder called `bar`, as well as a project called `bar`.  Visual Studio doesn't
+    # allow this, so we disambiguate folders from projects by surrounding folder names with parens.
+    # While we could try to be smart and only do this when we detect a collision, what happens then is
+    # that we would end up with solution folders, some with parens and some without parens, and it would
+    # break the sorting of child items when there is a long list of solution folders.  For example, you
+    # might end up with:
+    #    (bar)
+    #    (xyz)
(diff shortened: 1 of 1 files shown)


autozimu:
Looks great. Landing has initiated from internal Phabricator.

prelude/ide_integrations/visual_studio/gen_sln.bxl
```diff
@@ -35,18 +44,15 @@ def _gen_folder(folder_path, guid, project_names):
     )
 
 def _gen_project(target, vs_settings, project_names):
-    project_name = target.label.name
-    if project_names[project_name] > 1:
-        # Unique name to avoid prompt: The solution already contains an item named 'xxx'.
-        project_name = sanitize(target.label.package + "_" + target.label.name)
+    project_name = sanitize(target.label.name)
```
zjturner, line 47: @autozimu Removing this was intentional, but it made it back into the final committed version.  Did you revert this portion of my change?

prelude/ide_integrations/visual_studio/gen_sln.bxl
```diff
@@ -35,18 +44,15 @@ def _gen_folder(folder_path, guid, project_names):
...
 
     # List of project type GUIDs https://github.com/JamesW75/visual-studio-project-type-guid
     project_type_guid_cpp = "{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}"
 
     return _gen_entry(
         project_type_guid_cpp,
         project_name,
-        get_project_file_path(target.label, ".vcxproj"),
+        get_project_file_path(target.label, ".vcxproj").replace("/", "\\"),
```
zjturner, line 55: This part of my change was also reverted

scottcao:
Ah I believe auto-import only imported the first commit and I missed the second one. Can you make another PR for the second one? I will get it imported in

zjturner:
Sure, thanks!

zjturner:
https://github.com/facebook/buck2/pull/1066

PR #1062, opened 2025-08-06
Allow CAS TTL to be configurable

bvinc:
Expose the configuration named `cas_ttl_secs` to configure the assumed CAS TTL for bazel's remote execution API.


bvinc:
I think for this option in particular, people will probably want to specify a TTL in hours or days, and so allowing a suffix like "3h" or "30d" might be a good idea.  But I don't see any existing pattern for this in any configs.

Should this option just be called `cas_ttl` and accept or require a time suffix?

scottcao:
> I think for this option in particular, people will probably want to specify a TTL in hours or days, and so allowing a suffix like "3h" or "30d" might be a good idea. But I don't see any existing pattern for this in any configs.

> Should this option just be called cas_ttl and accept or require a time suffix?

My opinion is that this looks fine as is. A duration suffix can always be added later if there is demand for it, but otherwise asking people to do a little math before putting in a number here seems fine right now.

PR #1064, opened 2025-08-06
Add a json_file() argument type for bxl's command line parser

zjturner:
The idea here is to allow passing a path to a json file to a bxl arg.  the clap parser will load the json file and return its value as a json object, so the bxl just sees a dictionary.  This addresses issues with `cli_args.json()` and long json blobs that are too long for an operating system's command line.  It also makes invoking a bxl simpler, since it's generally not feasible to type out actual json content on a command line string.


autozimu:
@Nero5023

Nero5023:
It is landing now. Thanks for the PR

PR #1044, opened 2025-07-12
Allow `buck2 test` to build DefaultInfo and RunInfo providers

cbarrete:
This is a reboot of https://github.com/facebook/buck2/pull/702, which I had abandoned because the base had diverged too much.

Issue #873, opened 2025-03-12
Add support for excluding targets in `buck2 build`


cbarrete:
At work, we generate test targets with different sets of modifiers so that `buck2 test ...` builds/test a project with various configurations.
We tag the targets with non-default modifiers with an `extra_test` label, so that we can do `buck2 test --exclude extra_test ...` for faster iteration cycles.

As far as I can tell, there is no way to do the same thing for `buck2 build`. Is this intentional?

8Keep:
Use compatible_with or target_compatible_with to make them incompatible with the default configuration.

If you use /... target patterns in a build, incompatible targets will be skipped.

cbarrete:
Hey, thanks for the reply :)

Unfortunately this doesn't really have the same semantics, as the targets will _never_ be built, including e.g. in CI. The goal of `--exclude` is to selectively omit targets from the CLI, but to have them built otherwise.

TroyKomodo:
+1 on this its similar to bazels `tags = ["manual"]` where you can declare targets to be excluded from the build graph / test graph unless otherwise explicitly built by name or some other dependency depends on them. We have 3rd party deps that get built always even if they are not needed.

8Keep:
Idea is fine, would look at any pr implementing this

Issue #86, opened 2023-02-04
UI request: a shorthand for 'buck bxl' invocations (or: arbitrary 'buck foobar' commands)


thoughtpolice:
Here's a nifty one-liner I put behind the name `bxl`:

```bash
exec buck2 bxl "bxl//top.bxl:$1" -- "''${@:2}"
```

This requires you to:
- set `bxl = path/to/some/dir` in your root `.buckconfig` file `[repository]` stanza, and then
- export a bunch of top level `bxl()` definitions from `bxl//top.bxl` which can be invoked

Then you can just run the command `bxl hello --some_arg 123` in order to run an exported `bxl()` definition named `hello`. Pretty nice! And because the command invocation uses the `bxl//` cell to locate `top.bxl`, the `bxl` command can work in any repository that defines that cell, and the files can be located anywhere.

---

So my question is: could something like this be supported in a better way? The reason is pretty simple: it's just a lot easier to type and remember!

Perhaps the best prior art here I can think of is `git`, which allows `git foobar` to invoke the separate `git-foobar` binary, assuming it's in `$PATH` and has an executable bit set. We don't need to copy this exactly 1-to-1, but it's a generalized solution to this problem, and in fact it's arguable that being able to do a `buck2 foobar` subcommand is useful for the same reasons. So maybe that's a better place to start, and the `bxl` scripts could be a more specialized case of this.

ndmitchell:
It's great that Bxl is turning out to be so useful. Having `top.bxl` be hardcoded to being special seems unfortunate, and I'm not a huge fan of looking up things on the `$PATH`. But you could imagine adding entries to `.buckconfig`:

```
[commands]
bxlrun = "buck2 bxl bxl//top.bxl:$1 $*"
```

Or something of that form - so you can essentially introduce top-level command aliases to buck2. I could imagine defining `fasttargets` as an alias for `targets --streaming --keep-going --no-cache --inputs` (which is a command I do quite frequently).

thoughtpolice:
Yes, I think that something like that is the right way forward, but I didn't want to get too far ahead of myself!

[...]

PR #1065, opened 2025-08-07
test: allow `buck2.test_builds_targets` to build all RunInfo and DefaultInfos by default

thoughtpolice:


app/buck2_client_ctx/src/client_ctx.rs
```diff
@@ -294,6 +294,14 @@ impl<'a> ClientCommandContext<'a> {
             .log_download_method
             .clone())
     }
+
+    pub fn test_builds_targets(&self) -> buck2_error::Result<bool> {
+        Ok(self
+            .immediate_config
+            .daemon_startup_config()?
+            .test_builds_targets
+            .clone())
```
ahornby, line 303:
```suggestion
            .test_builds_targets)
```

8Keep:
@thoughtpolice I believe this is superceded by https://github.com/facebook/buck2/pull/1044 which was merged

Reopen if needed

thoughtpolice:
@8Keep That PR only adds support for building them in the command line, yes. This PR however adds support for doing that by default with a `.buckconfig` entry. I unfortunately can't reopen this myself, if you don't mind?

8Keep:
Sorry, it was a quick glance and I was mistaken. My opinion is that we should just switch over to this default internally, and then no buckconfig would be needed... Maybe we should have a buckconfig around for a while for transitioning purposes

thoughtpolice:
I've actually found at work that both behaviors can be useful; for example, we have expensive OCI image builds as well as many tests in a package, so just running `buck test ...` in a package will skip the image builds by default, which is good for most developers; while you can enable it on CI to keep it simpler and get everything in a single command. But in my personal monorepo, I always enable it (I have different convention to avoid that). So I feel like it could be nice for the end user to decide.

8Keep:
That seems like the intention would be to run only lightweight "things", and that set could include either tests or builds (or both). So one way or the other wouldn't really solve that universally.

Internally we have "heavyweight" labels to designate targets as being heavyweight. buck test has --exclude and --include to select targets to test. We could potentially make this just work with build too.

Or, you can always use a query like:

buck test $(buck query ...) with a labels query there to build/test whatever subset of targets you want.

Could also be combined with https://github.com/facebook/buck2/issues/86 for shorthand for ergonomics.

thoughtpolice:
Yeah, the main thing missing is that `build` does not have `--include` or `--exclude`. BTW, I also have a patch that lets you specify default target labels in `.buckconfig`, so e.g. CI developers can auto skip heavy targets.

8Keep:
@thoughtpolice Could you rebase this one + add a description with some context (link the other related PR)

cbarrete:
Relevant: https://github.com/facebook/buck2/issues/873

thoughtpolice:
Going to close this and refile with an updated description, etc.

PR #1072, opened 2025-08-11
fix certificate errors caused by tonic upgrade

bvinc:
tonic was upgraded from 0.10 to 0.12 and it broke webpki roots in the RE client.  In tonic 0.12 an additional method needs to be called to add webpki roots to your TLS config.


bvinc:
PKI certificates for remote execution have been broken since this commit: https://github.com/facebook/buck2/commit/754a939a944a49cf637b2c33868a9515cf572942

PR #1075, opened 2025-08-13
optimize calls to FindMissingBlobs

bvinc:
The cache now keeps track of in-flight requests for each blob.

- No longer make multiple requests about the same blob
- Sends full requests with 100 blobs whenever possible

In my tests of a large project, it reduced the number of blobs queried by about 20%, but it reduced the number of HTTP requests by about 90%.  The reduction in requests is largely due to better chunking of blobs into requests.  With this change, requests are much more optimal and contain as many blobs as possible, up to the hard coded limit of 100.

### DIFF
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -71,6 +71,7 @@ use regex::Regex;
 use tokio::fs::OpenOptions;
 use tokio::io::AsyncReadExt;
 use tokio::io::AsyncWriteExt;
+use tokio::sync::broadcast;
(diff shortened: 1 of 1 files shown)


bvinc:
I feel like the most controversial thing in this PR might be that if a FindMissingBlobs request fails, it reports an empty error message to all of the waiting tasks, and the waiting tasks report that the task they were waiting on failed, without any useful message.  I did this because the error type is not Clone, and I figure all the failed tasks should report something, and the real error message should still be logged by the one making the request.  But maybe this is too noisy and the waiting tasks should just silently fail?

PR #1076, opened 2025-08-13
`remote_execution`: Retry `BatchReadBlobs` requests

9999years:
We've seen a bunch of (transient) failures with our build cache when downloading artifacts. This adds retry logic for `BatchReadBlobs` requests similar to the `RemoteExecutionClient::new_retry` logic:

https://github.com/facebook/buck2/blob/83a21c0291118a94d1a5f173af17b3f3142940e2/app/buck2_execute/src/re/client.rs#L194-L217

In the future, we'll likely want more sophisticated tuning of backoff parameters (minimum & maximum delays, jitter, exponential backoff, etc.). I'm not able to test this at the moment as the failures are flaky.


sluongng:
It would be nice to extend this retry mechanism to all requests instead of just ReadBatchBlobs

jun-mercury:
We've been using this patch internally for a month, and it works well and reduced our occurrence of network transport errors to almost nothing.

PR #1080, opened 2025-08-15
python_toolchain: set and use preprocessor flags for extensions

zsol:
This PR:
* adds a new `extension_preprocessor_flags` field to `PythonToolchainInfo`
* starts using it from the `cxx_python_extension` rule
* wires the new field up from `remote_python_toolchain`
* moves the `python_cextension` targets from under `with_prelude/cpp` to `with_prelude/python`
* enables them so CI starts covering these
* removes the `main()` function from the extension cpp file

At the end of this all, `remote_python_toolchain` (and so the demo toolchains) is now capable of building Python C extensions without having Python installed on the host at all. 

Test plan:
```
buck2\examples\with_prelude  "python_toolchain: set a…" [8 +22-58]
❯ ..\..\target\release\buck2 run python/use_cextension:cext
Build ID: fe1005ad-8885-47c3-9d70-830a65addf98
Network: Up: 0B  Down: 0B
Command: run.
Time elapsed: 0.0s
BUILD SUCCEEDED - starting your binary
hello from cpython extension from cpp toolchain
```


zsol:
@itamaro wdyt?

itamaro (commented):
lgtm!

one naming nit -- I guess `extension_preprocessor_flags` is technically accurate, but perhaps it would be clearer to users to call this `extension_compiler_flags`? or even `cxx_extension_compiler_flags`, since this wouldn't make sense for, e.g., a rust python extension?

PR #1081, opened 2025-08-18
Allow load()ing .toml files from starlark

zsol:
This PR allows writing:
```python
load(":foo.toml", "value")
print(value)
```

The toml file is mapped to JSON, and then exposed the same way static JSON files are to starlark code. The single `value` name in a toml file will always be a dictionary (as TOML files are always tables).

TODO:
- [x] write e2e tests that are exercised in OSS CI
- [x] write docs

### DIFF
--- a/test.py
+++ b/test.py
@@ -391,6 +391,20 @@ def test(package_args: List[str]) -> None:
     print_running("cargo test --doc")
     run(["cargo", "test", "--doc", *extra_args, *package_args], timeout=timeout_sec)
 
+@contextmanager
+def chdir(target: Path) -> Generator[None, None, None]:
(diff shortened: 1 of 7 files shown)


thoughtpolice:
I've wanted this kind of feature literally a dozen times now, I think, and was recently thinking of (re?)proposing it; thank you so much for it!

zsol:
@JakobDegen wdyt?

JakobDegen:
Didn't see this

@zsol can you write up an API change proposal that I can copy-paste into Buck2 Dev? Doesn't need to be long, just enough to mention that this is a close analogue to the json case and then if you made any potentially controversial decisions mention those.

PR #1093, opened 2025-08-31
Add freebsd support in os_version()

svmhdvn:
This fixes the build on FreeBSD. There are still 4 remaining failing tests at this commit:

```
error: 4 targets failed:
    `-p buck2_client --lib`
    `-p buck2_forkserver --lib`
    `-p buck2_server --lib`
    `-p buck2_util --lib`
```


svmhdvn:
Ping on this, I'd like to package buck2 for FreeBSD and need this fix in.

PR #1101, opened 2025-09-23
Add configuration option to set the number of event logs retained on disk

reginaldl:
Introduce a new configuration parameter, `retained_event_logs`.
The parameter allows users to control the number of event logs retained on disk, replacing the previously hard-coded value of 12.
This change provides greater flexibility for users who need to retain more logs locally when Buck2 logs to disk.

This PR also fixes a bug where the wrong number of log files are being retained. This is due to the ExitResult logic which writes command_report subfolders in `buck-out/v2/log`. `remove_old_logs` counts all entries in this folder as opposed to log files only. We fix this bug by only counting files.

### DIFF
--- a/app/buck2_event_log/src/file_names.rs
+++ b/app/buck2_event_log/src/file_names.rs
@@ -46,7 +46,7 @@ pub(crate) fn get_logfile_name(
 
 pub(crate) async fn remove_old_logs(logdir: &AbsNormPath, retained_event_logs: usize) {
     if let Ok(logfiles) = get_files_in_log_dir(logdir) {
-        futures::stream::iter(logfiles.into_iter().rev().skip(retained_event_logs))
+        futures::stream::iter(logfiles.into_iter().rev().skip(retained_event_logs - 1))
             .then(|file| async move {
                 // The oldest logs might be open from another concurrent build, so suppress error.
                 tokio::fs::remove_file(file).await.ok()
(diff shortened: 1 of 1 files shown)


app/buck2_event_log/src/file_names.rs
```diff
@@ -44,11 +44,9 @@ pub(crate) fn get_logfile_name(
...
 }
 
-pub(crate) async fn remove_old_logs(logdir: &AbsNormPath) {
-    const N_LOGS_RETAINED: usize = 12;
-
+pub(crate) async fn remove_old_logs(logdir: &AbsNormPath, retained_event_logs: usize) {
     if let Ok(logfiles) = get_files_in_log_dir(logdir) {
-        futures::stream::iter(logfiles.into_iter().rev().skip(N_LOGS_RETAINED - 1))
+        futures::stream::iter(logfiles.into_iter().rev().skip(retained_event_logs))
             .then(|file| async move {
```
scottcao, line 50: Noticed that this went from N -1 to N. Is this correct?

scottcao: Maybe having a unit test for this would be helpful to verify that this is actually correct

reginaldl: Hi @scottcao! Sorry for the delay.
You are right, removing `- 1` was a mistake, we expect to create a new log file after calling `remove_old_logs`.
I am going to update this and add some tests.

I also found a bug here. This log folder may contain both files and folders. The folders are for command_reports, they are written by a different module. This other module should probably be fixed too because it only keeps writing and never cleans things up.
I think in this `remove_old_logs` we should only care about log files.

PR #1102, opened 2025-09-24
Add buck2 rules that implement CMake's configure_file() behavior.

zjturner:
This adds a set of rules that implement CMake `configure_file()` semantics, significantly easing the burden of integrating libraries that use CMake (or other configure-like systems) into buck builds.

Typically, such build systems will run a step before the build begins that determines values for all of the things it considers important.  Then it will output a header file (or otherwise modify the build in some way, for example by adding preprocessor defines) that contains preprocessor defines, typedefs etc that contain definitions based on the values that were determined earlier.

There are two basic categories of meta-rules implemented in this PR, and then one "real" rule that does the actual work.

**cctest rules** hardcode values about your toolchain that would otherwise be computed by something like autoconf, configure, or CMake itself.  For example, these systems may try to detect whether or not your toolchain has `ssize_t` defined, or the size of an integer.  These systems do this by compiling small little programs and examining the output to determine what happened.  This approach is flexible, but it is non-hermetic and necessarily requires a configure step.   Hermetic build systems -- by definition -- know everything about the build environment.  These rules do not have any outputs, and simply provide a way to encode the known values into build targets that can be used by the next category of rules: **substitution rules**.

**substitution rules** define mappings between variable names that might appear in the substitution file, and values that were previously encoded in the form of **cctest rules**.  To explain why this distinction is needed, consider two libraries, both of whom need to detect whether `ssize_t` exists on your system.  One library expect its configured header file to have a line of the form:

```
#define HAVE_SSIZE_T
```

while the other may expect to have a line of the form:

```
#define HAVE_BUILTIN_TYPE_SSIZE_T
```

Whether or not `ssize_t` exists is a universal truth, but the libraries expect to consume that information through different ways.  So we need one set of rules (**cctest rules**) to encode the universally known values, and another set of rules (**substitution rules**) to define the library-specific mappings.

Finally, we have the `cmake_configure_file` rule which accepts a list of substitutions and will perform the generation.

### DIFF
--- a/prelude/cxx/cctest.bzl
+++ b/prelude/cxx/cctest.bzl
@@ -1,0 +1,53 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+#
+# This source code is dual-licensed under either the MIT license found in the
+# LICENSE-MIT file in the root directory of this source tree or the Apache
+# License, Version 2.0 found in the LICENSE-APACHE file in the root directory
+# of this source tree. You may select, at your option, one of the
+# above-listed licenses.
+
+CcTestValueInfo = provider(
+    fields = {
+        "value": provider_field(typing.Any)
+    }
+)
+
+CcTestTypeSizeInfo = provider(
+    fields = {
+        "type": provider_field(str),
+        "size": provider_field(int)
+    }
+)
+
+def cctest_value_impl(ctx: AnalysisContext) -> list[Provider]:
+    return [
+        DefaultInfo(),
+        CcTestValueInfo(value = ctx.attrs.value)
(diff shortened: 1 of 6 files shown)


zjturner:
I ran into some limitations that I want to address before this goes in but I’ll leave it open for now.

specifically:  i want to change the cc test rules to be called “generic value” rules, move them out of cxx, add support for list value types, change the embedding rule yo use a dynamic action to read the content and return a GenericValueInfo, and move the concatenation / prefix / suffix logic and attributes into the list value rule so that it can work with non file values

PR #1103, opened 2025-09-25
Add support for bytestream compression

bvinc:
- If capabilities allow, use compression for bytestream API uploads and downloads.

- Uploading zero sized blobs now succeeds silently without uploading anything, when previously it would fail with an error.  This is allowed by the specification of the bazel remote execution API.

- If an upload returns with a committed_size of -1, which indicates the blob has already been uploaded, it is no longer an error.  Previously, it would have been reported as an error.

### DIFF
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -10,6 +10,8 @@
 
 use std::collections::HashMap;
 use std::env::VarError;
+use std::io;
+use std::io::Cursor;
 use std::num::NonZeroUsize;
 use std::pin::Pin;
 use std::sync::Arc;
@@ -18,6 +20,12 @@ use std::time::Duration;
 use std::time::Instant;
 
 use anyhow::Context;
+use async_compression::tokio::bufread::BrotliDecoder;
+use async_compression::tokio::bufread::BrotliEncoder;
+use async_compression::tokio::bufread::DeflateDecoder;
+use async_compression::tokio::bufread::DeflateEncoder;
+use async_compression::tokio::bufread::ZstdDecoder;
+use async_compression::tokio::bufread::ZstdEncoder;
 use buck2_re_configuration::Buck2OssReConfiguration;
 use buck2_re_configuration::HttpHeader;
 use dupe::Dupe;
@@ -69,9 +77,13 @@ use re_grpc_proto::google::rpc::Code;
 use re_grpc_proto::google::rpc::Status;
 use regex::Regex;
 use tokio::fs::OpenOptions;
+use tokio::io::AsyncBufRead;
+use tokio::io::AsyncRead;
 use tokio::io::AsyncReadExt;
 use tokio::io::AsyncWriteExt;
+use tokio::io::BufReader;
 use tokio::sync::broadcast;
+use tokio_util::io::StreamReader;
 use tonic::codegen::InterceptedService;
 use tonic::metadata;
 use tonic::metadata::MetadataKey;
@@ -222,6 +234,8 @@ pub struct RECapabilities {
     max_total_batch_size: usize,
     /// Does the remote server support execution.
     exec_enabled: bool,
+    /// Compressors supported by the "compressed-blobs" bytestream resources.
+    supported_compressors: Vec<Compressor>,
 }
 
 /// Contains runtime options for the remote execution client as set under `buck2_re_client`
@@ -248,6 +262,35 @@ impl InstanceName {
         match &self.0 {
             Some(instance_name) => format!("{instance_name}/"),
             None => "".to_owned(),
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum Compressor {
+    Zstd,
+    Deflate,
+    Brotli,
+}
+
+impl Compressor {
+    fn from_grpc(val: i32) -> Option<Self> {
+        // TODO: brotli cannot be supported until google proto definitions are updated
+        if val == compressor::Value::Zstd as i32 {
+            Some(Self::Zstd)
+        } else if val == compressor::Value::Deflate as i32 {
+            Some(Self::Deflate)
+        } else {
+            None
+        }
+    }
+
+    /// The compressor name used in compressed-blob resource paths
+    fn name(&self) -> &str {
+        match self {
+            Self::Zstd => "zstd",
+            Self::Deflate => "deflate",
+            Self::Brotli => "brotli",
         }
     }
 }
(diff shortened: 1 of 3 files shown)


bvinc:
The update just now was to fix clippy warnings.

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -345,6 +388,21 @@ impl REClientBuilder {
...
+            .supported_compressors
+            .contains(&Compressor::Zstd)
+        {
+            Some(Compressor::Zstd)
+        } else if capabilities
+            .supported_compressors
+            .contains(&Compressor::Deflate)
+        {
+            Some(Compressor::Deflate)
+        } else {
```
cbnilrem, line 404: It is unclear to me why there isn't a conditional for Brotli here. Please add explanatory comment if this is intentional.

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -252,6 +266,34 @@ impl InstanceName {
...
+    Brotli,
+}
+
+impl Compressor {
+    fn from_grpc(val: i32) -> Option<Self> {
+        if val == compressor::Value::Zstd as i32 {
+            Some(Self::Zstd)
+        } else if val == compressor::Value::Deflate as i32 {
+            Some(Self::Deflate)
+        } else {
```
cbnilrem, line 283: As elsewhere - it's unclear from context why Brotli isn't being called out as an option here. Please add a comment if this is intentional.

bvinc: Good catch.  The reason is that brotli is not supported in the google protocol definitions that are checked into this repo.  Once those are updated to their latest versions, I can change this code to support brotli.  I've added a comment for now.

remote_execution/oss/re_grpc/src/client.rs
bvinc: Good catch.  The reason is that brotli is not supported in the google protocol definitions that are checked into this repo.  Once those are updated to their latest versions, I can change this code to support brotli.  I've added a comment for now.

bvinc:
I've added a comment as to why brotli is not currently supported, which is because the google proto definitions currently checked in do not support brotli because they are out of date.  I didn't want to do too many different things in one PR.

So I have PR #1108 to update the google protocol definitions.  If #1108 gets in first, I'll update this one to have brotli support.

dtolnay:
@bvinc, since this PR, I am seeing materialized files get truncated to 128KiB when using BuildBuddy for remote execution. Can you tell whether this is Buck2 doing something wrong or BuildBuddy doing something wrong?

Repro: from an x86_64 or aarch64 Linux host:

1. Check out https://github.com/dtolnay/buck2-rustc-bootstrap
2. `./submodule_init.sh`
3. Create `.buckconfig.local` containing BuildBuddy buck2_re_client as shown in the readme; you will need a free API key from [app.buildbuddy.io](https://app.buildbuddy.io)
4. `buck2 build stage0:ci_llvm --show-simple-output | xargs -I{} ls -l {}/lib/libLLVM.so.20.1-rust-1.90.0-nightly`
    - Correct size: 178MiB
    - Actual size: 131072 bytes (128×1024)
5. `buck2 build stage1:rustc --show-simple-output | xargs ls -l`
    - Correct size: 93MiB
    - Actual size: 131072 bytes

cormacrelf:
@dtolnay Couple of things, may help, may not:

(0: I can reproduce the issue.)

1. I just ran the tests (`test_upload_compressed` + `test_download_compressed`) with 100MB of random data and also interspersed empty `ReadResponse`s instead of `vec![1; 1024]` and they still passed.

2. I can't find any constants in the buck source relating to 128KiB. But 131072 is the maximum block size for zstd blocks other than RLE-encoded. This is most likely related to the issue. Zstd is the only format supported by BuildBuddy. 

3. Only really sus thing I found in this PR so far is `let mut buf = vec![0, max_total_batch_size]`. The `max_total_batch_size` can be zero, which forces use of the bytestream API always, but then the chunked read buffer is zero sized, in which case no uploads would happen at all. If you fix this (`vec![0, max_total_batch_size.max(1)]`) then it fails `client::tests::test_upload_empty`, revealing a different bug: an empty file is not empty when run through a compressor! This is going to be tricky to solve (need some kind of `CountingReader`? Or a PR to async-compression), may just want to avoid zero sized uploads higher up, which we are probably already doing.

4. BuildBuddy code lives at https://github.com/buildbuddy-io/buildbuddy/blob/master/server/util/compression/compression.go. But they've had Zstd for years now. Someone would have noticed, you would think. Incidentally they seem to have done perf tests on best chunk sizes we could learn from.

5. This is a bug during compressed download. If you upload LLVM first, disable compression (by modifying client.rs) and download it again, it's fine. It also takes a lot longer: including daemon start, 13 sec vs 1m 37s.

6. If you do this, it downloads the full compressed file. This is either a bug with async-compression or we are holding it wrong.

   ```diff
                // Wrap the blob reader in a compression reader
                let reader: Pin<Box<dyn AsyncRead + Unpin + Send>> = match bystream_compressor {
   -                None => Pin::new(Box::new(blob_reader)),
   -                Some(Compressor::Zstd) => Pin::new(Box::new(ZstdDecoder::new(blob_reader))),
   +                None | Some(Compressor::Zstd) => Pin::new(Box::new(blob_reader)),
                    Some(Compressor::Deflate) => Pin::new(Box::new(DeflateDecoder::new(blob_reader))),
   ```


   ```sh
   du -sh (~/src/buck2/target/debug/buck2 build stage0:ci_llvm --show-simple-output)/bin/llvm-as
   ...
   Command: build.    Finished 1 cache (100% hit)                                                                 3.4s exec time cached (100%)
   Time elapsed: 21.1s
   BUILD SUCCEEDED
   2.1M    buck-out/v2/gen/rust/1c7f8cab15283ef5/stage0/__ci_llvm__/ci-llvm/bin/llvm-as
   ```

7. This test fails with 131072.

   ```
   use async_compression::tokio::bufread::ZstdDecoder;
   #[tokio::test]
   async fn break_zstd_directly() {
       let llvm_as = tokio::fs::File::open("/tmp/llvm-as.zstd").await.unwrap();
       let buffered = tokio::io::BufReader::new(llvm_as);
       let mut decoder = ZstdDecoder::new(buffered);
       let mut out = Vec::new();
       tokio::io::copy(&mut decoder, &mut out).await.unwrap();
       assert_eq!(out.len(), 4401672) 
   }
   ```

One more detail now that I found the problem, which is async-compression cannot read more than one zstd frame. I think most compressors use one frame only, so nobody has noticed before. I think BuildBuddy's zstd encoder may be finishing a frame each time it writes a block. This may be suboptimal in terms of compression ratio, but maybe this is faster and less memory intensive to do on the fly.

PR #1106, opened 2025-09-26
cargo_package: write documentation discouraging use

lf-:
This probably should be moved to another file name altogether ("reindeer-internal", "reindeer_third_party" or something like that) since most people read "cargo" and think it's what they want, but that will break a lot of people.


dtolnay (approved):
Thanks!

JakobDegen (approved):

PR #1108, opened 2025-09-26
Update the Google protocol definitions

bvinc:
Update the proto files to their latest versions.  This allows for some new methods and new compression type (brotli).  But annoyingly, it also causes some fields to become deprecated, even though we still want to set them.  This change only updates the protocol definitions and should not contain any behavior changes at all.


bvinc:
I've rebased this change since the first change had test failures that did not appear to be related to my change, hopefully the next time the tests run, they'll work.

PR #1110, opened 2025-09-29
Support for explicit Git object formats in .buckconfig

mmhat:
This PR adds a new optional config option for external cells retrieved via Git:
```
[external_cell_foo]
git_origin = ...
commit_hash = ...
object_format = sha1

[external_cell_bar]
git_origin = ...
commit_hash = ...
object_format = sha256
```
If specified, we pass `--object-format sha1` or `--object-format sha256` as arguments to the `git init` invocation.
In both cases we validate that the commit_hash is a valid digest for the respective algorithm. If no object_format is given, we check that the commit_hash is a valid SHA1 digest, but we do not pass the --object-format option when initializing the Git repository.

### DIFF
--- a/app/buck2_common/src/legacy_configs/cells.rs
+++ b/app/buck2_common/src/legacy_configs/cells.rs
@@ -9,6 +9,7 @@
  */
 
 use std::collections::HashSet;
+use std::str::FromStr;
 use std::sync::Arc;
 
 use allocative::Allocative;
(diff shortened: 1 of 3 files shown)


Nero5023:
Sorry for late reply. Thank you for doing it!
Could you do a rebase?

mmhat:
> Sorry for late reply. Thank you for doing it! Could you do a rebase?

@Nero5023 No problem; I rebased the commit :+1:

PR #1112, opened 2025-09-30
Fix deprecation of pkg_resources

tmi:
Replaces deprecated pkg_resources (which is scheduled for removal in 2025-11-30 !) with importlib.resources -- part of python stdlib since like python 3.9. No observable change expected (beyond disappearance of a bunch of warnings)

### DIFF
--- a/prelude/apple/tools/code_signing/app_id_test.py
+++ b/prelude/apple/tools/code_signing/app_id_test.py
@@ -6,10 +6,9 @@
 # of this source tree. You may select, at your option, one of the
 # above-listed licenses.
 
+import importlib.resources
 import plistlib
 import unittest
-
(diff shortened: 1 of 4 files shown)


prelude/apple/tools/code_signing/app_id_test.py
```diff
@@ -38,8 +37,9 @@ def test_entitlements_parsing(self):
             "test_resources/test3.plist",
         ]
 
+        package = importlib.resources.files(__packages__)
```
watwars: I think this should be `__package__` instead of `__packages__`

tmi: you are right, thx! Fixed

PR #1116, opened 2025-10-09
Add in auto install for Gentoo

td5038:
Adds in autoinstall for Gentoo users to be able to have that creature comfort. I'm a Gentoo user, so that's why I added it.


td5038:
This is VERY EXPLICITLY a creature comfort.

PR #1121, opened 2025-10-15
remove find_missing_blobs deduplication to fix bug

bvinc:
Previously, as part of an optimization to FindMissingBlobs calls, the find_missing_cache was changed so that the first task to ask about a blob could notify others of the result when it was finished.  This introduced a bug related to cancellation:  If the first task is canceled before it finishes, every other task that was waiting on the result fails with a "channel closed" error.

Since then, there has been a GetDigestsTtlDeduper introduced at a higher layer, outside of the remote execution client, that deduplicates these same calls.

This change completely gets rid of attempting to deduplicate FindMissingBlobs calls inside of the remote execution client.  At the moment, I think this is the best course of action, and to get deduplication of these calls, it should be preferred to enable deduplicate_get_digests_ttl_calls.

### DIFF
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -82,7 +82,6 @@ use tokio::io::AsyncRead;
 use tokio::io::AsyncReadExt;
 use tokio::io::AsyncWriteExt;
 use tokio::io::BufReader;
-use tokio::sync::broadcast;
 use tokio_util::io::StreamReader;
 use tonic::codegen::InterceptedService;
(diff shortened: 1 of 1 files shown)


scottcao:
Looks good.

I want to reference a comment from @JakobDegen from the original diff that introduced GetDigestsTtlDeduper.

> Fwiw the lack of buck-side dedupe has been a paint point for OSS since their RE clients don't do any of that - this particular diff will solve a problem for them, and also maybe we should just bite the bullet and do something like this across the board. Idk

If doing this across the board is something that would benefit you, we can certainly consider it.

PR #1129, opened 2025-10-21
fix suggestion to include full path

KlotzAndrew:
The current BUCK file suggestion usually returns an invalid path, this change returns a valid suggestion in the case of an invalid path to a BUCK file. 

When you have a missing BUCK file, buck tries to suggest what you meant. Instead of returning the full path, it returns the directory the suggested BUCK file is in, which can be confusing and means the suggestion cannot be copy+pasted

e.g. `root//cat` is not a valid location:
```
dir `root//animals/pets/cats` does not exist. Did you mean `root//cat`?
```

After, we now return the full path to the matched directory allowing copy+paste of the suggestion to fix:
```
dir `root//animals/pets/cats` does not exist. Did you mean `root//animals/pets/cat`?
```

PS: I have no idea if the tests pass, the README says they can only run inside meta so this is just a guess

### DIFF
--- a/app/buck2_common/src/file_ops/error.rs
+++ b/app/buck2_common/src/file_ops/error.rs
@@ -132,7 +132,10 @@ pub(super) fn extended_ignore_error<'a>(
                                 return Some(ReadDirError::DirectoryDoesNotExist {
                                     path: path.to_owned(),
                                     suggestion: DirectoryDoesNotExistSuggestion::Typo(
-                                        suggestion.to_owned(),
+                                        match parent.path().join_normalized(suggestion) {
+                                            Ok(p) => p.as_str().to_owned(),
+                                            Err(_) => suggestion.to_owned()
+                                        }
                                     ),
                                 });
                             }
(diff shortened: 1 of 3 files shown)


app/buck2_common/src/file_ops/error.rs
```diff
@@ -132,7 +132,10 @@ pub(super) fn extended_ignore_error<'a>(
                                 return Some(ReadDirError::DirectoryDoesNotExist {
                                     path: path.to_owned(),
                                     suggestion: DirectoryDoesNotExistSuggestion::Typo(
-                                        suggestion.to_owned(),
+                                        match parent.path().join_normalized(suggestion) {
+                                            Ok(p) => p.as_str().to_owned(),
+                                            Err(_) => suggestion.to_owned()
```
scottcao, line 137: Nit: it's kinda weird to return the original suggestion on error since the suggestion is not relevant. Can we get the following behavior instead? On error, return NoSuggestion as suggestion and fire a soft_error!() that joining normalized paths unexpected failed when recommending suggestions.

PR #1132, opened 2025-10-30
LSP gets type checking

### LINKED ISSUE #1126: `buck2 lsp` shows parse error diagnostics at line 1 column 1 rather than the error location
For this parse error (from CLI output):

```
Command failed:

Caused by:
    4: Error parsing: `prelude//haskell/compile.bzl`
    5: error: Parse error: unexpected symbol ',' here, expected one of "]"
           --> prelude/haskell/compile.bzl:1274:53
            |
       1274 |         package_deps: dict[str, dict[str, list[str]],  # `dict[modname, dict[pkgname, list[modname]]`
            |                                                     ^
            |
```

`buck2 lsp` shows a single diagnostic at line 1 column 1 with the text `` Error parsing: `prelude//haskell/compile.bzl` ``. The actual parse error message (`unexpected symbol ',' here, expected one of "]"`) is hidden, as is the error location (line 1274 column 53).

cormacrelf:
Basically this is "evaluate on save" in addition to the usual parsing.

- First, fix #1126 (for syntax errors), which is pretty nice. ~~We do sacrifice one line of buck2_error context, but it's worth it.~~
- Do not evaluate on every keystroke as starlark code may do a lot of work. Just on save.
- New trait method in LspContext to evaluate a module, default impl is noop.

Works for the raw starlark binary already:

image

~~Needs  some~~ Needed a lot of work with buck2_error to get this experience with buck2's lsp. I made a bespoke error converter and formatter for LSP only.

Before work on buck2_error span recovery and formatting:

Screenshot 2025-10-31 at 12 28 44 am

After:

Screenshot 2025-11-05 at 7 44 00 pm

--- 

### Roadmap

It is already very useful but here is a quick roadmap to make this even better, lowest hanging fruit first:

- Make `--unstable-typecheck` stable, whatever is holding it back. Helps a LOT, for the time being I recommend LSP users use `buck2 lsp --unstable-typecheck` as the LSP server.
- There is some barrier to type-checking provider code like `ctx.attrs.dep[ProviderName].blah`, the `.blah` is not checked probably because the .[] is too generic. (creating a provider manually does get type checked) That will get you much more coverage in real .bzl code. I quickly looked into this and it will require `#[starlark(ty_custom_function = ...)]` to be supported on methods.
- Possibly eval-on-keystroke by changing `LspContext::parse_file_with_contents` implementation to eval instead and emptying the eval_file_with_contents impl. Will need to be rigorous about not caching these evals, currently eval_build_file I think is cached. Which is fine for now as it's eval-on-save.
- If you can do that, you can get up-to-date type information from the type checker. It knows how to produce errors for `x: str = 5` so it does know what type things are or might be. Store this as you go and save it when freezing the module, ideally only in LSP mode. Then you get hover-for-type.
- Extend that to attribute completion. The holy grail.

### DIFF
--- a/app/buck2_common/src/legacy_configs/cells.rs
+++ b/app/buck2_common/src/legacy_configs/cells.rs
@@ -493,7 +493,10 @@ impl BuckConfigBasedCells {
         #[derive(buck2_error::Error, Debug)]
         #[buck2(tag = Input)]
         enum ExternalCellOriginParseError {
-            #[error("Unknown external cell origin `{0}`")]
+            // Disabled is handled outside this function
+            #[error(
+                "Unknown external cell origin `{0}`: valid values are 'bundled', 'git', 'disabled'"
+            )]
             Unknown(String),
             #[error("Missing buckconfig `{0}.{1}` for external cell configuration")]
             MissingConfiguration(String, String),


cormacrelf:
I got it to evaluate build files as build files, etc. and uhh this is amazing

image

cormacrelf:
Ok, this was quite a lot of stuff. But it works great.

There are a few things that I think will get annoying, but the feature is useful enough that someone else is bound to get annoyed enough to fix them too. So I think that's a good stopping point.

I invite you to try it and challenge you to live without it.

cormacrelf:
Reminder for me to check if print() in evaluated starlark code messes up the LSP stdio. I turned off printing in starlark_bin when LSP is evaluating but didn't check this for buck2. 

Edit: we are ok.

cormacrelf:
Ok, I was wondering why this didn't work on the prelude and as usual it's because the prelude is bundled so the eval part just evaluates the bundled sources, which does not incorporate any of your edits. This is not normally an issue in the wild. I don't see any way of detecting this as all the CellPath conversions etc work on these virtual project relative paths -- whether prelude is bundled or not, the file seems to live at `prelude/prelude.bzl`. I would like to have a way of warning the user about this so they don't think it's just broken. If anyone can think of one, go for it.

Edit: Never mind got it.

Screenshot 2025-11-05 at 8 56 36 pm

9999years:
This is awesome, thanks for putting this together!

cormacrelf:
You are welcome!

cormacrelf:
Bit of an update, I've been using this and improving types (see all the other PRs). Turns out running the actual typechecker is more useful than actually evaluating if you have to pick one, because the typechecker can return multiple errors. In that vein I'll probably take out the evaluation parts, unless I can get full-workspace cross-bzl-file evaluation errors to mix with typechecker errors. For now this branch is back to WIP.

At the end of the day you get this, typecheck on every keystroke and hover-for-type-info on bindings. It would be nice for the hover typer to run through dotted expressions by applying `.attr` to the `Ty` of the `base.of.the.expression`. If you're doing that you're getting mighty close to dot-attr completion.

starlark type checker interactive

cormacrelf:
I'm going to split this into a few more PRs:

- the buck2_error parts of this one, which fixes the parse errors not getting spans, and preps for cross-file errors
- eval-on-save, if people want that
- typecheck on every keystroke (was never part of this PR)
- workspace diagnostics, e.g. parse error in some other file prevents typechecking this one (was never part of this PR)

9999years:
@cormacrelf Splitting it up into separate PRs sounds great, hopefully that'll make it easier to get parts of this merged sooner.

PR #1134, opened 2025-10-30
Remove the execute bit from python cxx scripts

bvinc:
Remove the execute bit from python cxx scripts

These scripts are not being executed directly, so the execute bit is never really used.  The reason why it is beneficial to remove the execute bit is that when compiling remotely for Linux from Windows, the python scripts are sent without the execute bit being set, but when compiling remotely for Linux from Linux, the execution bit is being taken from the file on the file system.

The execution bit being different means that the input files will differ between the scenario where a Windows machine is building for Linux versus the scenario where a Linux machine is building for Linux.  This ruins the ability to cache actions between these two scenarios, as the action hash will be different.


bvinc:
Specifically, I really want to make `makefile_to_dep_file.py‎` non-executable, as it is the specific script that is causing us problems and does not even include a main function.  But I think the problem is more general than that, and it may make sense to go over every non-executable file used as inputs in buck builds.

I do not think any of these are ever executed directly.  Can anyone come up with any problems it would cause to remove the execution bits?

dtolnay (approved):
Thanks!

PR #1140, opened 2025-11-10
Merge pull request #1 from facebook/main

JaclynCodes:
buck2_eden: handle opqaue network errors without error code


JakobDegen:
This seems to not even be syntactically valid, and also I'm not sure the unit tests are actually adding anything

PR #1143, opened 2025-11-14
fix unused import error in windows build

ahornby:
Summary: Remove unused import that is breaking oss build on windows

Differential Revision: D87057605


JakobDegen (approved):

PR #1145, opened 2025-11-17
fix clippy forkserver break in windows github actions

ahornby:
Summary:
The oss build runs clippy on the workspace.  Unlike Cargo build dependencies Cargo workspace members can't have conditional inclusions by target os,  so forkserver is covered even on windows, resulting in failures when importing Unix only parts of stdlib.

Fix by making the top level forkserver lib.rs do nothing on windows.

Differential Revision: D87215575


JakobDegen (changes requested):

JakobDegen (approved):

PR #1148, opened 2025-11-19
Stricter and more complete typechecking

cormacrelf:
This fixes a number of issues in the typechecker. I think it's fair to say before this PR we were only really checking function call sites and return types. The improvements here make type annotations on regular bindings significantly more useful. In general, we behave a lot more like a Python typechecker, as I looked at what basedpyright does in these situations and made starlark do that.

1. `x: str = ""` checked the immediately assigned expression, but no other assignments to `x`. In fact `x` was fed to the binding solver anyway, which meant a binding could hold whatever you wanted it to. Now we obey the annotation for all assignments, including locals from function parameters. That's almost all of the churn in the prelude, reassigning to function parameters with a different type.
2. `x: str = ...; x: int = ...` was accepted and was equivalent to `int | str` because of point (1). Now this is disallowed, you can only annotate a binding once.
3. `x[0]: int = 5` was allowed and ignored. No longer.
4. Top level module bindings were not checked. Now they are. I wrote a recursive binding solver that checks the current scope before any child `def`s, so this manages to feed even fewer bindings to `solve_bindings` at once than before.
5. I forget if this worked before this PR, probably not because of point (1), but I added tests and made sure `x: str = ...; x, y = (1, 2)` was caught as a type error.

I fixed the prelude's type errors, using some separate LSP improvements beyond the ones I've submitted already.

As for reviewers, maybe @ndmitchell? A few old diff messages refer to you as knowing the most about this.

---

A couple of TODOs for me:
- [x] change the golden test not to have double decl
- [x] failing B2IFB test because assert_eq redefined

### DIFF
--- a/prelude/android/android_binary_native_library_rules.bzl
+++ b/prelude/android/android_binary_native_library_rules.bzl
@@ -232,13 +232,13 @@ def get_android_binary_native_library_info(
         if native_library_merge_non_asset_libs:
             mergemap_cmd.add(cmd_args("--merge-non-asset-libs"))
         native_library_merge_dir = ctx.actions.declare_output("merge_sequence_output")
-        native_library_merge_map = native_library_merge_dir.project("merge.map")
+        native_library_merge_map_file = native_library_merge_dir.project("merge.map")
         split_groups_map = native_library_merge_dir.project("split_groups.map")
         mergemap_cmd.add(cmd_args(native_library_merge_dir.as_output(), format = "--output={}"))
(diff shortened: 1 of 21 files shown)


Wilfred:
FWIW this looks great at a high level, but we don't have a lot of starlark people to review this PR, and it's a very big change. Could you split it up so we can work incrementally? For example, the toplevel bindings checks sounds like it could be separate PR.

Issue #1131, opened 2025-10-29
Rescan events from the notify crate are not being handled, causing cache inconsistencies


bvinc:
When not using watchman, which is the default, buck2 uses the `notify` crate to monitor file system events to keep its view of the file system consistent.  While watchman and the `notify` both sit on top of inotify on Linux, but they have different behavior and APIs.  Specifically, inotify can miss events when its internal queue overflows.  inotify tells you when this happens with a flag.

https://man7.org/linux/man-pages/man7/inotify.7.html

> Note that the event queue can overflow.  In this case, events are
       lost.  Robust applications should handle the possibility of lost
       events gracefully.  For example, it may be necessary to rebuild
       part or all of the application cache.  (One simple, but possibly
       expensive, approach is to close the inotify file descriptor, empty
       the cache, create a new inotify file descriptor, and then re-
       create watches and cache entries for the objects to be monitored.)

Watchman has logic to handle this:

https://facebook.github.io/watchman/docs/install#linux-inotify-limits

> When the kernel reports an overflow, watchman will assume that all the files have been modified and will re-crawl the directory tree as though it had just started watching the dir.

The `notify` crate, like the inotify API, simply reports to the caller when this happens.

https://docs.rs/notify/8.2.0/notify/event/enum.Flag.html#variant.Rescan

Buck2 does not handle this case, which is a bug, and it can cause buck2's caches to become out of sync with the file system when a large number of changes occur.

One possible fix is for buck2 to have a way to invalidate all of its file system caches when a rescan event is seen.

bvinc:
This was fixed by #1149

PR #1149, opened 2025-11-19
Drop all DICE nodes when FS events missed

bvinc:
Handle the case where the notify crate tells us that events were missed.  In this case, when sync is called at the beginning of a new build, the entire DICE graph gets dropped.  This should fix the bug where caches can become out of sync with the file system due to missed events.

### DIFF
--- a/app/buck2_file_watcher/src/notify.rs
+++ b/app/buck2_file_watcher/src/notify.rs
@@ -37,6 +37,7 @@ use notify::event::MetadataKind;
 use notify::event::ModifyKind;
 use notify::event::RemoveKind;
 use starlark_map::ordered_set::OrderedSet;
+use tracing::debug;
 use tracing::info;
 
 use crate::file_watcher::FileWatcher;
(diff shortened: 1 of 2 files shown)


bvinc:
I think the thing to consider here is the effect this change has on multiple builds happening concurrently.

bvinc:
This should fix #1131

app/buck2_file_watcher/src/notify.rs
```diff
@@ -268,7 +281,12 @@ impl NotifyFileWatcher {
         let mut guard = self.data.lock().unwrap();
         let old = mem::replace(&mut *guard, Ok(NotifyFileData::new()));
         let (stats, changes) = old?.sync();
```
JakobDegen, line 288: I think it'd be a bit cleaner to return an `Option<FileChangeTracker>` from this instead of shoehorning the data into that type

JakobDegen (changes requested):

PR #1152, opened 2025-11-27
buck2_error -> lsp diagnostics

### LINKED ISSUE #1126: `buck2 lsp` shows parse error diagnostics at line 1 column 1 rather than the error location

cormacrelf:
This fixes #1126 (no spans on parse errors in LSP) in a pretty comprehensive way. It is #1132, but cleaner and without the eval-on-save parts.

The simple way was to remove buck2_error wrapping of parse errors, leaving only starlark errors which preserved their spans. But this lost the error context for buck2 output, since buck2_error was providing that. And it did nothing for actual buck2_errors, like those from eval-on-save. So instead, I wrote an LSP formatter for arbitrary buck2_errors.

Features:

- Does not render spans for the current file in LSP, since that's redundant. But spans from elsewhere are rendered.
- If the error has multiple bits of starlark context in the current file, split into multiple errors
- Spans in StarlarkContext can be LSP-only so we don't pollute buck2_error output.
- CallStacks are preserved instead of stringified, which means we can find spans for LSP inside call stacks as well

This is good enough for eval-on-save, where errors come from all over the place and have starlark attached midway through, or twice, etc. I have another branch where this is adapted to provide workspace diagnostics to LSP, i.e. take you right to a parse error in another file.

### Why was this so complicated?

buck2_error is very stringy. Firstly, we were losing spans almost as soon as they entered buck2_error, only preserving them in some cases, and other times immediately rendering them to string. We were then losing any spans we managed to save, due to pre-rendering StarlarkContext as strings basically whenever we convert the buck2_error to/from anything else. So this PR preserves StarlarkContext everywhere, which was a lot of work to get right.

Note that you do not get a span attached directly when you `starlark::Error::from(buck2_error)`. It is hidden deep inside there in some buck2 context, which will resurface when you convert back to buck2_error.

There is a lot of change to buck2_error, but very few observable changes outside LSP. It changes only one golden test, with better merging of CallStacks.

### DIFF
--- a/app/buck2_error_tests/src/conversion.rs
+++ b/app/buck2_error_tests/src/conversion.rs
@@ -156,3 +156,128 @@ fn test_from_starlark_source_location_is_caller() {
         loc,
     );
 }
+
+#[test]
+fn test_starlark_fail() {
+    let mut a = Assert::new();
+
+    a.module(
+        "helper",
+        r#"
+def will_fail():
+    fail("root error")"#,
+    );
+
+    let e = a.fail(
+        r#"
+load('helper', 'will_fail')
+will_fail()"#,
+        "root error",
+    );
+
+    golden_test_template(
+        "src/golden/test_starlark_fail.golden",
+        trim_rust_backtrace(&format!("{:?}", buck2_error::Error::from(e))),
+    );
+}
+
+/// fail() goes through starlark -> buck2 -> starlark again, producing
+/// multiple starlark contexts that get concatenated.
+#[test]
+fn test_starlark_fail_through_multiple_evals() {
+    #[starlark_module]
+    fn module(builder: &mut GlobalsBuilder) {
+        fn re_eval_failure() -> starlark::Result<NoneType> {
+            let mut a = Assert::new();
+            a.module(
+                "inner",
+                r#"
+def will_fail():
+    fail("root error")"#,
+            );
+            let e = a.fail(
+                r#"
+load('inner', 'will_fail')
+will_fail()"#,
+                "root error",
+            );
+            let buck: buck2_error::Error = e.into();
+            Err(buck.context("between evals").into())
+        }
+    }
+
+    let mut a = Assert::new();
+    a.globals_add(module);
+
+    let e = a.fail(r#"re_eval_failure()"#, "between evals");
+
+    golden_test_template(
+        "src/golden/test_starlark_fail_through_multiple_evals.golden",
+        trim_rust_backtrace(&format!("{:?}", buck2_error::Error::from(e))),
+    );
+}
+
+#[test]
+fn test_span_label_from_starlark_eval() {
+    #[starlark_module]
+    fn module(builder: &mut GlobalsBuilder) {
+        fn native_failure() -> starlark::Result<NoneType> {
+            let e = buck2_error::buck2_error!(
+                buck2_error::ErrorTag::StarlarkError,
+                "native error root cause"
+            )
+            .context("error during native call");
+            Err(e.into())
+        }
+    }
+
+    let mut a = Assert::new();
+    a.globals_add(module);
+
+    a.module(
+        "helper",
+        r#"
+def call_native():
+    native_failure()"#,
+    );
+
+    let e = a.fail(
+        r#"
+load('helper', 'call_native')
+call_native()"#,
+        "error during native call",
+    );
+
+    golden_test_template(
+        "src/golden/test_span_label_from_starlark_eval.golden",
+        trim_rust_backtrace(&format!("{:?}", buck2_error::Error::from(e))),
+    );
+}
+
+/// Same native failure, but with an extra .context() added after
+/// the starlark evaluation, to verify it layers on top.
+#[test]
+fn test_span_label_from_starlark_eval_with_outer_context() {
+    #[starlark_module]
+    fn module(builder: &mut GlobalsBuilder) {
+        fn native_failure() -> starlark::Result<NoneType> {
+            let e = buck2_error::buck2_error!(
+                buck2_error::ErrorTag::StarlarkError,
+                "native error root cause"
+            )
+            .context("error during native call");
+            Err(e.into())
+        }
+    }
+
+    let mut a = Assert::new();
+    a.globals_add(module);
+
+    let e = a.fail(r#"native_failure()"#, "error during native call");
+
+    let buck = buck2_error::Error::from(e).context("added after starlark eval");
+    golden_test_template(
+        "src/golden/test_span_label_from_starlark_eval_with_outer_context.golden",
+        trim_rust_backtrace(&format!("{buck:?}")),
+    );
+}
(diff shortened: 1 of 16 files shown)


app/buck2_error/src/error.rs
```diff
@@ -52,9 +52,14 @@ pub struct Error(pub(crate) Arc<ErrorKind>);
...
 pub(crate) enum ErrorKind {
+    /// The originating error, before context is added
     Root(Box<ErrorRoot>),
-    /// For now we use untyped context to maximize compatibility with anyhow.
+    /// The output of [`e.context(...)`][Error::context]; recursively contains Error.
+    /// There is eventually a [Self::Root] at the bottom.
     WithContext(ContextValue, Error),
+    /// The output of [`e.mark_emitted()`][Error::mark_emitted]; recursively contains Error.
+    /// There is eventually a [Self::Root] at the bottom.
+    ///
```
ahornby, line 60:
```suggestion
```

app/buck2_error/src/context_value.rs
```diff
@@ -106,29 +113,67 @@ impl<T: TypedContext> From<T> for ContextValue {
...
+    /// We don't render the root error if `replaces_root_error` is set.
+    /// We render this in preference because it has a starlark
+    /// span and call stack.
+    ///
+    /// This is only to be set if error_msg contains a rendering of the root error.
+    ///
+    /// This field enables us to keep an actual StarlarkContext tha can be rendered
+    /// as an LSP span instead of literally replacing the root context with a
+    /// stringified span.
+    ///
```
ahornby:
```suggestion
```

app/buck2_error/src/lsp.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// error: fail: message here
+    ///   --> blah.bzl:10:2-15
+    ///    |
+    /// 10 | fail("message here")
+    ///    |
+    ///
+    /// Whereas if you render this in blah.bzl, you do not need to see
+    /// a printout of the code you are literally already looking at.
+    /// So we skip rendering this.
+    ///
```
ahornby, line 65:
```suggestion
```

app/buck2_server/src/lsp.rs
```diff
@@ -556,6 +563,57 @@ impl<'a> BuckLspContext<'a> {
     fn find_target(ast: &AstModule, target: TargetName) -> Option<Span> {
         ast.find_function_call_with_name(target.as_str())
     }
+
+    /// We have `AstModule::parse(project_relative_path.as_str(), ...)`
+    /// So all our codemaps, file spans, etc, are project relative paths.
+    ///
```
ahornby, line 572:
```suggestion
```

app/buck2_error/src/context_value.rs
```diff
@@ -106,22 +107,24 @@ impl<T: TypedContext> From<T> for ContextValue {
...
         if let Some(ctx) = other {
-            let trimmed = ctx
+            let frames = self
                 .call_stack
-                .trim_start_matches(starlark_syntax::call_stack::CALL_STACK_TRACEBACK_PREFIX);
-            let call_stack = format!("{}{}", self.call_stack, trimmed);
+                .frames
+                .iter()
+                .chain(ctx.call_stack.frames.iter())
+                .cloned()
```
JakobDegen: (nit)

```suggest
                .frames
                .iter()
                .cloned()
                .chain(ctx.call_stack.frames)
```

And maybe self by value and you can avoid the remaining clones?

app/buck2_error/src/root.rs
```diff
@@ -74,6 +74,7 @@ impl ErrorRoot {
         self.id
     }
 
+    #[allow(unused)]
```
JakobDegen, line 78:
```suggest
FIXME(cormacrelf,JakobDegen): Can't be right? 
```

That's a bit disconcerting

app/buck2_error/src/starlark_error.rs
```diff
@@ -83,16 +88,34 @@ fn inject_starlark_context(e: &crate::Error, starlark_context: StarlarkContext)
...
+                        // walk back up from where we got to
                         let mut inner_err = inner.clone();
                         for context in context_stack.into_iter().rev() {
                             inner_err = inner_err.context(context);
                         }
 
+                        // A string or other context_value is popped & replaced by a StarlarkContext
+                        // that describes it.
+                        //
+                        // XXX: we are losing the Typed nature of the context.
```
JakobDegen: Wow, this is a bit insane lol.

So I apologize a bit for making you fix bugs you didn't write here, but this code is just super hard to reason about and that makes me a bit nervous.

Instead of doing this, can we just push all of this bullshit into the formatting? So, I'd propose:

 1. Unimplement `Display` on `ContextValue` and `StarlarkContext`
 2. Remove the `error_msg` field from `StarlarkContext`
 3. Completely replace `inject_starlark_context` with just blindly pushing a `StarlarkContext` on top
 4. When creating a starlark error in the non-native case, just `inject_starlark_context(Error::new(description))`, so create a chain like `[starlark_msg, starlark_ctx]`
 5. When formatting, merge adjacent stuff together at that time so that you can render the starlark error however you like

I think you should be able to do this mostly without regressions, but I'd also be happy to take small rendering regressions in exchange for the simplicity

JakobDegen (changes requested):

app/buck2_error/src/root.rs
cormacrelf: I was a bit alarmed when I saw this too. I deleted the only actual usage, which was to construct a new ErrorRoot with the same metadata as a previous one.

app/buck2_error/src/starlark_error.rs
cormacrelf: Basically yes, I would prefer it that way.

cormacrelf: Done. No regressions in buck2_error_tests.

cormacrelf:
This is ready for review again. As a reminder, this is what you're getting at the end: parse errors show up in LSP attached to a span.

Screenshot 2026-04-18 at 11 27 39 pm

Be aware:

- I did everything you suggested and it worked pretty nice!
- First 7 commits tiny
- A few commits don't build completely until we get to 'Rewrite buck2_error formatting' because this was a lot of work rebasing and I've had enough now. Feel free to squash that whole batch 'Improve StarlarkContext::concat, less cloning` through `Rewrite buck2_error formatting`.
- Helper methods on Result/error may be a bit iffy. There was already a `pub fn context_for_starlark_backtrace` but it can't be used outside the crate as StarlarkContext is not public.

PR #1155, opened 2025-11-28
Temporarily disable zstd bytestream compression

dtolnay:
Summary:
Remote execution with BuildBuddy is still impacted by the bug described in <https://github.com/facebook/buck2/pull/1103#issuecomment-3413866377>.

Zstd compression can be re-enabled after `async-compression` has been fixed, potentially by <https://github.com/Nullus157/async-compression/pull/400>.

Differential Revision: D87992619

### DIFF
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -470,8 +470,17 @@ impl REClientBuilder {
             cache_cap
                 .supported_compressors
                 .iter()
-                .cloned()
-                .filter_map(Compressor::from_grpc)
+                .filter_map(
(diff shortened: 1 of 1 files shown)


cormacrelf:
Did you see? It works with `decoder.multiple_members(true);`. Just put that in the zstd decoder init, and probably also the gzip one, which will have the same issue. I should have summarised that as the outcome after my long investigation.

cormacrelf:
While you're there, may as well grep `brotli cannot be supported` and add brotli support, as the protobufs have been updated. Probably use `decoder.multiple_members(true);` for that as well.

dtolnay (commented):
Thank you!

Closing in favor of https://github.com/facebook/buck2/pull/1156.

PR #1160, opened 2025-11-30
Add missing newline in `buck2 expand-external-cells`

cbarrete:
The output was missing a newline, causing the shell prompt to be rendered right after the reminder.


dtolnay (approved):
Thanks!

PR #1161, opened 2025-11-30
Index string at chars boundaries

cbarrete:
This function would panic when the truncated message contained non-ASCII characters, and the truncation happened between within char boundaries.

I've not been able to test this because I can't seem to easily build Buck2 from source these days, but this should be trivial enough.

### DIFF
--- a/app/buck2_client_ctx/src/subscribers/superconsole.rs
+++ b/app/buck2_client_ctx/src/subscribers/superconsole.rs
@@ -1003,8 +1003,9 @@ fn truncate(contents: &str) -> Option<String> {
     if contents.len() > MAX_LENGTH + BUFFER {
         Some(format!(
             "{} ...<omitted>... {}",
-            &contents[0..MAX_LENGTH / 2],
-            &contents[contents.len() - MAX_LENGTH / 2..contents.len()]
+            &contents[0..contents.ceil_char_boundary(MAX_LENGTH / 2)],
+            &contents
+                [contents.floor_char_boundary(contents.len() - MAX_LENGTH / 2)..contents.len()]
         ))
     } else {
         None


app/buck2_client_ctx/src/subscribers/superconsole.rs
```diff
@@ -1001,10 +1001,20 @@ fn truncate(contents: &str) -> Option<String> {
     const MAX_LENGTH: usize = 200;
     const BUFFER: usize = " ...<omitted>... ".len();
     if contents.len() > MAX_LENGTH + BUFFER {
+        let first_slice_end = contents
+            .char_indices()
+            .map(|(i, _)| i)
+            .nth(MAX_LENGTH / 2)
+            .unwrap();
```
JakobDegen: Don't think this can be right. `contents.len()` is in bytes, you can have 200 bytes but <100 characters.

I think maybe you just want `.floor_char_boundary()`

JakobDegen (changes requested):

app/buck2_client_ctx/src/subscribers/superconsole.rs
cbarrete: > Don't think this can be right. contents.len() is in bytes, you can have 200 bytes but <100 characters.

Yeah, this is trading off one potential panic for another, less likely one.

> I think maybe you just want .floor_char_boundary()

I do! Thanks for pointing that out.
I've used ceil for the first part and floor for the second one so that it's less likely that it will split in the middle of a grapheme, but I'm happy to switch that if you prefer.

JakobDegen: > I've used ceil for the first part and floor for the second one so that it's less likely that it will split in the middle of a grapheme

It's not clear to me why that's the property this gets but I don't care :)

JakobDegen (approved):

PR #1167, opened 2025-12-07
Document materialization of artifacts referenced by string parameter …

### LINKED ISSUE #1166: String parameter macro artifacts are not materialized when cached or built remotely
Consider the following:

```starlark
genrule(
    name = "gen",
    out = "out.txt",
    cmd = "echo $(location :bin) > $OUT",
)

cxx_binary(
    name = "bin",
    srcs = ["main.cpp"],
)
```

Where the contents of `main.cpp` are irrelevant.

If RBE is enabled and `:bin` is built remotely or cached, it will not be materialized locally, so `out.txt` will point to a file that does not exist.

This can be simply reproduced by building `:gen` remotely, and it can be proved that this is an RBE-related materialization issue because `buck2 build --local-only --no-remote-cache :gen ` has the expected behavior: `bin` is materialized locally.

cbarrete:
…macros

Closes #1166.

### DIFF
--- a/docs/rule_authors/string_parameter_macros.md
+++ b/docs/rule_authors/string_parameter_macros.md
@@ -157,6 +157,35 @@ operation in the `buck audit` command.
 Expands to the named output file or directory of the given target, for rules
 that expose supplementary outputs.
 
+## Materialization of referenced artifacts
+
+Note that, unlike in systems such as [Nix](https://nixos.org/),
+artifacts that are referenced via string parameter macros are not
+necessarily materialized.
+
+This means that using string parameter macros to pass for example the
+path to a binary to a configuration file or a script, the binary may not
+be present on the filesystem at the path that is embedded in the
+configuration file/script.
+
+To ensure that the referenced artifacts are materialized, the rule being
+called must do one of the following:
+
+1. Call [`Artifact.with_associated_artifacts`](../api/build/Artifact/#artifactwith_associated_artifacts)
+   to add the input as an associated artifact to the output.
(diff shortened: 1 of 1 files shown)


cormacrelf:
I don't really see how this is specific to string parameter macros. I practically never use string parameter macros and I still need to know this. It's perhaps better placed on RunInfo because being a runtime dependency is the main reason you need something materialized, or in the docs on materialization generally. 

For RunInfo, you also have the ability to put hidden artifacts in RunInfo's cmd_args. I am not sure whether we materialize DefaultInfo (and other_outputs) when you run an associated RunInfo, but you could check and document what we do.

cbarrete:
> I don't really see how this is specific to string parameter macros. I practically never use string parameter macros and I still need to know this. It's perhaps better placed on RunInfo because being a runtime dependency is the main reason you need something materialized, or in the docs on materialization generally.

I guess this really depends on use cases. For example, the most common use case in my org is to actually expose shared libraries, not binaries, so `DefaultInfo` is relevant, not `RunInfo`. But more importantly, end users barely know/care about providers, they care about what they actually use, which is SPM. They will not even think of checking out the docs for providers, because it is not the level of abstraction that they operate at.

So while I don't mind _also_ adding a page on materialization (there doesn't seem to be one yet, only https://buck2.build/docs/users/advanced/deferred_materialization/), I think that it is important to have this here as well.

> For RunInfo, you also have the ability to put hidden artifacts in RunInfo's cmd_args. I am not sure whether we materialize DefaultInfo (and other_outputs) when you run an associated RunInfo, but you could check and document what we do.

`DefaultInfo` does not get built, much less materialized, when calling `buck2 run`.

PR #1168, opened 2025-12-07
Ensure test runnable generated by rust-project works with the default rust test harness

nick96:
I've been working on setting up a project with buck2 lately and ran into an issue with editor integration (specifically vscode with rust analyser).

The generated runnable doesn't work with buck2 because it hardcodes "buck". It also doesn't respect the `--buck2-command` flag and passed flags to the rust test binary that the default rust test harness does not understand (e.g. `--print-passing-details`). Now, the command is generated from the `Buck` struct so `--buck2-command is respected`. I've also tweaked the flags that get passed through to map to the default rust test harness, perhaps there should be a way of configuring these flags for custom test harnesses? The rational for each flag is:

- `--exact`: we only want to run the one test and have its full name. Since rust allows functions and modules to have the same name in the same namespace, it's possible that a test name is the prefix of another.
- `--no-capture`: I'm assuming `--print-passing-details` used to do something like this?

### DIFF
--- a/integrations/rust-project/src/buck.rs
+++ b/integrations/rust-project/src/buck.rs
@@ -212,19 +212,39 @@ pub(crate) fn to_project_json(
         check_cycles_in_crate_graph(&crates);
     }
 
+    let buck_test_command = buck.command([
+        "test".to_owned(),
+        CLIENT_METADATA_RUST_PROJECT.to_owned(),
+        "{label}".to_owned(),
+    ]);
+
+    let mut args: Vec<String> = buck_test_command
+        .get_args()
+        // It is safe to call unwrap on this because we've constructed all the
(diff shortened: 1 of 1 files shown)


cormacrelf:
This is good, I like the exact and no-capture. You're running one test so may as well show all output. Note that the previous code is for Meta's internal-only TPX test runner. We don't have that but they still use it and so this code should be different depending on cfg(fbcode_build), like it is in #767, which does a similar thing.

nick96:
Thanks for the feedback @cormacrelf! I've updated my PR based on it

Wilfred:
@nick96 would you mind running rustfmt on your changes?

nick96:
@Wilfred sorry for the delay. I've just pushed with formatting

PR #1169, opened 2025-12-08
re_grpc: wire up OSS RequestMetadata

sluongng:
In open source remote execution setup, request metdata let the server group
cache requests using a build (invocation) and build targets.


sluongng:
worth noting that there are still a few requests that don't come with metadata / identity attached.

sluongng:
Closing this stale branch for now. The current fork has request metadata folded into later action metadata/BES work, and this branch is badly out of date. We should resend this as a fresh smaller PR after the proto foundation/prerequisites are refreshed rather than keeping this stale stack open.

PR #1170, opened 2025-12-09
ReActionIdentity: make target and paths optional

sluongng:
Instead of making 'identity' optional in some of the request paths, make the fields inside ReActionIdentity optional instead.

This enables us to mandate 'identity' for all call RBE client consumer. We simplify the creation of the identity by introducing a minimal constructor to use in different places with insufficient data to build a full one.

This change is based on https://github.com/facebook/buck2/pull/1169


sluongng:
Closing this stale stacked follow-up for now. It depends on the old request-metadata branch shape, while the current fork has evolved. We should resend the optional identity cleanup as a fresh smaller PR after the metadata base is refreshed.

PR #1172, opened 2025-12-14
Ignore set_cfg_constructor() in external cells

LegNeato:
When a repository is used as an external cell by another repository, its PACKAGE/BUCK_TREE file's `set_cfg_constructor()` call would previously error because it's not the root cell.


cbarrete:
@JakobDegen  This would be very useful. I currently implement this client-side, but it would be great if it just worked out of the box

LegNeato:
Any hope of this getting merged?

PR #1184, opened 2025-12-20
fix: nix flake lock outdated

TroyKomodo:
The flake lock is using an older version of https://github.com/oxalica/rust-overlay

which at the time only supported rust version `2025-05-28`
https://github.com/oxalica/rust-overlay/blob/d68a69dc71bc19beb3479800392112c2f6218159/manifests/nightly/default.nix#L508

This updates the lock and then adds a documentation to future bumpers to also update the lock when they change the version again.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -48,11 +48,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1748399823,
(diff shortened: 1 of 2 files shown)


thoughtpolice:
This was updated recently by https://github.com/facebook/buck2/commit/1c708ee9b1cefbc31bcc05468f7017f5f17d1a0e -- sorry for this being ignored for so long!

PR #1187, opened 2025-12-27
rust-project: populate `sysroot_src`

cbarrete:
According to
https://rust-analyzer.github.io/book/non_cargo_based_projects.html, rust-analyzer should compute `sysroot_src` from `sysroot` if it is absent.

According to my testing, this is not the case, and rust-analyzer is broken if `sysroot_src` is not populated.

To fix this, we populate it from a new `--sysroot-src` flag, or the de facto standard `RUST_SRC_PATH`, which fixes the issue.

Ideally, rust-analyzer should be fixed but:
1. I haven't gotten to the bottom of that bug.
2. This commit remains correct in the meantime.
3. This commit allows setups where the binary sysroot and source sysroots are different (as is the case in Nixpks, where they come from different packages).

### DIFF
--- a/integrations/rust-project/src/cli/develop.rs
+++ b/integrations/rust-project/src/cli/develop.rs
@@ -62,6 +62,7 @@ impl Develop {
             stdout,
             prefer_rustup_managed_toolchain,
             sysroot,
+            sysroot_src,
             pretty,
             mode,
             check_cycles,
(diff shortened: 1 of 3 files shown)


cormacrelf:
This is covered by #767 in a bit more of a comprehensive way. (767 auto populates sysroot_src from sysroot, and also checks that that sysroot_src path exists, i.e. whether you have the `rust-src` component if using rustup.)

cbarrete:
I missed that, thanks!
Closing in favor of https://github.com/facebook/buck2/pull/767

cbarrete:
Reopening since https://github.com/facebook/buck2/pull/767 is seeing no activity/responses.

PR #1191, opened 2025-12-29
Improve documentation about build configuration

cbarrete:
This commit rearranges existing documentation and adds more explanation, including a comprehensive overview of the various parts (platforms, modifiers and transitions), and how they are relevant for end users.

In particular, a lot of information was moved from `rule_authors/` to `concepts/`, as build configuration is important for end users.

A lot of information was also deduplicated across pages.


8Keep:
@scottcao  @Nero5023 I like this change. Thoughts?

docs/concepts/configurations.md
```diff
@@ -3,14 +3,318 @@ id: configurations
...
-will compute the appropriate configuration for a target and display a version of
-that target's attributes with the configuration applied. The `uquery` command
+Build configurations are how Buck models building the same target in
+different ways. This can include (but is not limited to):
+
+- Target architecture
+- Target OS
+- Optimization level/build mode
+- Compiler type/version
+- Language version (e.g. C++ standard version or Rust edition)
```
cormacrelf: Rust's editions are not a good example. Editions are set on a per-crate basis, and the rust compiler also happily links different editions together, so most third party crates are still using old editions and that's fine, they all link to the same libcore/libstd and all interoperate. I would never add buck constraints for edition. The closest I would get to configuring an edition globally is setting a default edition in a macro. But you would never be able to switch between editions globally, changing it is a big task that requires touching the source code a lot.

I think C++ standard version does fit though, since you'll be linking a different C++ stdlib and nobody does multi-std builds.

docs/concepts/configurations.md
```diff
@@ -3,14 +3,318 @@ id: configurations
...
+  different effects.
+- Being used for constraining compatibility (e.g. "this target can only
+  be built in release mode" for a benchmark).
+- Being used to "transition" part of the build (e.g. "this target and
+  its dependencies are always built in release mode, regardless of the
+  dependent")
+
+## Configuration values
+
+Configurations can also include values taken from the buckconfig:
```
cormacrelf: If so, they cannot be used on the command line and have it *set* those buckconfigs. (I believe.) These config_settings can only be selected on / used to constrain.

docs/concepts/configurations.md
```diff
@@ -3,14 +3,318 @@ id: configurations
...
+        ":build_mode[debug]",
+        ":compiler[clang_21]",
+        ":asan[enabled]",
+    ],
+)
+```
+
+Note that the prelude defines some generic constraints, e.g. under
+`prelude//os:` and `prelude//cpu:`, which you might want to consider
+using for interoperability.
```
cormacrelf, line 81: A question I've had for a while is, if I have a new CPU type that's not in the prelude, can I extend that with a `constraint_value()` target outside the prelude? Seems like you might be able to with the old constraint_setting/constraint_value way, but I never tried it. The new unified API looks less amenable to that. Would be useful to document how to do this if it's possible.

docs/concepts/configurations.md
```diff
@@ -3,14 +3,318 @@ id: configurations
...
+        "DEFAULT": [],
+    }),
+)
+```
+
+If multiple conditions of the `select()` match, then the select will be resolved
+to the "most refined" of the conditions that match. A set of constraints (as in
+a `config_setting`) is said to "refine" another if it is a superset of that
+other's constraints. The "most refined" of a set is then the condition that
+refines all the others. If there is no "most refined" condition of the matching
```
cormacrelf: "If only one condition matches, then the select resolves to that condition, but if there are multiple matching conditions and none of them is the "most refined" condition, it is an error."

docs/concepts/configurations.md
```diff
@@ -3,14 +3,318 @@ id: configurations
...
+ones, it is an error.
+
+Note that `select()` is resolved during configuration. This happens
+after the evaluation of the BUCK file is completed, and so Starlark code
+run during BUCK file evaluation does not have access to the resolved
+value. This can make it difficult to have macros that do extensive
+modification or inspection of attributes (which should be done in rules
+instead). However, some functions
+([`select_map`](../api/build/#select_map) and
+[`select_test`](../api/build/#select_map)) allow performing
```
cormacrelf: Broken link to select_test

docs/concepts/configurations.md
```diff
@@ -3,14 +3,318 @@ id: configurations
...
+    target_compatible_with = select({
+        "//:cxx_standard[20]": ["prelude//:none"],
+        "DEFAULT": [],
+    })
+)
+```
+
+Target platform compatibility is transitive, all _dependents_ of an incompatible
+target are incompatible. In other words, a node is compatible if and only if the
+node itself and all of its transitive dependencies are compatible.
```
cormacrelf: Could be a bit clearer:

> Target compatibility requires all transitive dependencies to be compatible as well. In other words, a target is compatible if and only if the
target itself and all of its transitive dependencies are compatible. In the usual case of a dependency via `attrs.dep()`, a target's dependency will be configured and then checked for compatibility with the same configuration as the dependent target.

docs/concepts/configurations.md
```diff
@@ -3,14 +3,318 @@ id: configurations
...
+
+For example:
+
+```sh
+# Build this target with the default configuration.
+buck2 build :my_target
+# Build it with an entirely different configuration.
+buck2 build :my_target --target-platforms //my/other:platform
+# Build it with the default configuration, plus release mode.
+buck2 build :my_target?release
```
cormacrelf, line 259: Would also add `buck2 build :my_target -m release` as an example.

docs/concepts/configurations.md
```diff
@@ -3,14 +3,318 @@ id: configurations
...
+java_library(
+name = "foo",
+deps = [
+    "//libs:common",
+] + select({
+    "//constraints:x86": ["//libs:x86"],
+    "//constraints:mac-arm64": ["//libs:mac-arm64"],
+    "//constraints:windows-arm64": ["//libs:win-arm64"],
+    "DEFAULT": ["//libs:generic"],
+})
```
cormacrelf: Indentation

docs/concepts/configurations.md
```diff
@@ -3,14 +3,318 @@ id: configurations
...
+encoded into output paths. The configuration is currently represented as a hash
+of its values (a "hashed buck-out").
+
+## Target platform vs execution platform
+
+Buck distinguishes two kinds of targets: "regular" ones, and the ones
+used as build tools. The rationale is that it is common to want
+different build configurations for those. For example, it is typical to
+want build tools (e.g. a compiler) to be built/run in release mode, even
+when building debug targets.
```
cormacrelf: I don't really think it's useful to think of them as different kinds of _targets_. It is very common to have a target used in both modes, to compile stuff, and also at runtime.

My take on this:

> Buck requires you to specify both a target platform and an execution platform. The target platform is where your code will run. The execution platform is where the compilers and other build tools will run. Many different build tools have different names for these things ("host", etc), but those are Buck's names.

>
> The simplest execution platform setup is the one `buck2 init` uses. This setup gathers constraints from the host machine Buck is running on.
>
> ```ini
> [parser]
>   target_platform_detector_spec = target:root//...->prelude//platforms:default
> [build]
>   execution_platforms = prelude//platforms:default
> ```
>
> For many projects, this will suffice, and your target and execution platform will be the same (until you provide a config modifier, which applies only to the target platform). But the target/exec distinction forms the basis of all kinds of cross compilation and remote build execution. You can use these to express "I want to compile code that will eventually run on Windows, but all the build tools and compilers should run on my local linux computer". In that case the target platform is Windows, and the execution platform is Linux. You can imagine exotic situations in which the compiler itself has to be compiled first, on yet another execution platform.
>
> So, more complex setups are possible. You can:
> - Add more constraints and configure code differently when it will be executed in a build step (e.g. release mode, for faster builds of everything else)
> - Set up cross compilation, in conjunction with toolchains that will provide the right flags
> - Let buck automatically select from multiple execution platforms depending on what's being built (for example, most of the build can be done on Linux, but the linker might only run on Windows)
>
> The process for fully specifying your own execution platforms is:
>
> 1. Create a target that exposes an `ExecutionPlatformRegistrationInfo(platforms = [...])` provider. Each platform is an `ExecutionPlatformInfo`, which has its own `ConfigurationInfo` (a set of constraints describing it, e.g. it's an x86 server running linux). This `ConfigurationInfo` is used for exec platform resolution/compatibility, and also to configure software that will run there. So often you will tell it you want all build tools to be built themselves in release mode, so your builds are faster.
> 2. Configure the `.buckconfig` value `build.execution_platforms` to point to this target:
> 
>     ```
>     [build]
>     execution_platforms = platforms//:my_exec_platforms
>     ```
>
> In simple cases you will only have one execution platform, and the story ends there.
>
> In more complex cases, you may have multiple execution platforms. For example, you may have a remote build farm that has both Linux and Windows machines. When a build is requested for a particular configured target, Buck will iterate the platforms provided in the registration provider, and select the first platform whose configuration matches the execution constraints. Basically, some build tools only run on Linux, so if the tools need to be built, Buck will configure them to be built for Linux, and then when it comes time to run them, it will schedule them to run under the Linux execution platform. Other build tools (a cross-platform python script) could run anywhere and these will not influence the choice of exec platform for a given target. The `ExecutionPlatformInfo` provider that is ultimately chosen supplies key-value data that is sent to the remote build farm that can be used to comply with the request, like `"OSFamily": "linux"` or a given Docker image. You could have dozens of auto-generated execution platforms, or a few well-known platforms that are maintained and rotated as you migrate infrastructure over time. See \[TODO: relative link to <https://buck2.build/docs/rule_authors/configurations_by_example/#exec-deps>\] for more information on exec platform resolution.

docs/concepts/configurations.md
```diff
@@ -89,3 +393,16 @@ while the `nix` dependency is needed only for Linux. In `cquery` that
 distinction has been resolved; because the target has been configured for Linux,
 the `nix` dependency is present and indistinguishable from any other, while the
 `common-path` dependency is gone.
+
+## Execution groups
+
+Execution groups are a future feature that will allow a rule to perform
+execution platform resolution multiple times and then specify in which of the
+resolved platforms each action runs in.
```
cormacrelf, line 438: Add: "Traditionally, each target resolves a single execution platform."

cormacrelf (commented):
This is great work, very valuable docs. Wish this had been around 2 years ago. I have mostly small comments but one big one on exec platforms.

docs/concepts/configurations.md
cormacrelf: Actually, just put "If only one condition matches, then the select resolves to that condition." at the start of the para. The multiple matches case is rare and makes it seem like a difficult feature to wrap your head around, when it's mostly very easy.

docs/concepts/configurations.md
cbarrete: The thing I had in mind was using PACKAGE files to bump a subset of a repo to a new Rust edition.
I think that the use case is valid, but I've removed that reference to keep things non-controversial.

docs/concepts/configurations.md
cbarrete: As far as I am aware, that's not possible.

I did consider documenting both APIs, but I don't want to confuse the readers (there's already a lot in here) and I assume that Meta does not want to encourage using the new API.

The question is valid though, and I suggest opening an issue about that.

docs/concepts/configurations.md
cbarrete: That's right, although I think that the paragraph after the code block explains how to use this well enough?

I'm not sure what you're getting at, did you have an edit in mind?

docs/concepts/configurations.md
cbarrete: Woops, thanks!

docs/concepts/configurations.md
cbarrete: Yeah, I had just copied that over, but your suggestion reads much better. Thanks!

docs/concepts/configurations.md
cbarrete: Agreed/done/thanks

docs/concepts/configurations.md
cbarrete: Done

docs/concepts/configurations.md
cbarrete: > I don't really think it's useful to think of them as different kinds of targets. It is very common to have a target used in both modes, to compile stuff, and also at runtime.

You're right, thanks for catching that.

I have incorporated what you wrote, although I put most of it in the `rules_author/` file, as I think that it is not relevant for most end users, and don't want to confuse them.

Thanks, and let me know what you think!

cbarrete (commented):
Thanks for the review!

docs/concepts/configurations.md
cormacrelf: Turns out it works great in both cases.

```starlark
genrule(
    name = "printout",
    out = "out.txt",
    cmd = "echo 'cpu:" + select({
        "config//cpu:x86_64": "x86_64",
        "root//:napoleon": "napoleon",
        "DEFAULT": "other",
    }) + " abc:" + select({
        "root//:abc[abc]": "abc",
        "root//:abc[def]": "def",
        "root//:extra": "extra",
        "DEFAULT": "other",
    }) + "' > $OUT",
)

constraint_value(
    name = "napoleon",
    constraint_setting = "prelude//cpu/constraints:cpu",
)

constraint(name = "abc", values = ["abc", "def"], default = "abc")
constraint_value(name = "extra", constraint_setting = ":abc")

```


    buck2 build :printout --out=- -m root//:napoleon -m root//:extra
    BUILD SUCCEEDED
    cpu:napoleon abc:extra

docs/concepts/configurations.md
cormacrelf: Very well. I'll show you what I mean anyway.

> using PACKAGE files to bump a subset of a repo to a new Rust edition.

If you mean using set_package_value() and reading it with a macro wrapper for rust_library, absolutely I would do that, it's essentially the ideal API for setting editions. But that isn't a buck cfg/constraint.

If you mean with set_cfg_modifiers(), I would still never do that. That would then only apply if you built a package in that subtree directly, meaning some builds set the edition to 2018, some to 2024... It's impossible to write cross-edition-compatible rust code because new editions break old source code and they don't even think about breakage in the other direction so old editions almost certainly break new source code too. The edition is basically part of the source code for a given crate.

It would be normal and useful to have the rustc version (1.91.0, etc) be a constraint so you could check everything works on older compilers while using a newer compiler for everyday development.

docs/concepts/configurations.md
cormacrelf: It does have a good explanation of how to use it, but it comes straight after a section saying "here's config_setting, you can use it to set all its constituent constraints on the command line".

So

> `config_setting` can also include values taken from the buckconfig. These can ease a migration from a legacy buckconfig setting to a build constraint by allowing you to select() on known buckconfig values.
>
> ... This setting will be satisfied if ... etc
>
> You can only query buckconfig with this feature, you cannot write values. If used as a CLI modifier `-m root//:fastmode_enabled`, this does nothing.

docs/concepts/configurations.md
cbarrete: Oh interesting! The inconsistency in syntax is... unfortunate, though.

Anyway, I've documented it as it's not obvious.

docs/rule_authors/configurations.md
```diff
@@ -288,8 +247,73 @@ def target_compatible_with(target, cfg):
     return True
 ```
 
-## Execution groups
+## Toolchain deps
+
+In addition to `attrs.exec_dep()`, there are `attrs.toolchain_dep()`, which are
+similar but differ in an important way. These nodes don't select their execution
+platform, but instead inherit the execution platform of whatever target
+references them; hence, it must be recorded in the configured target label. In
```
cormacrelf: ~~; hence, it must be recorded in the configured target label~~ not quite clear what you're saying, but in any case that's buck's job, I don't need to think about putting stuff in target labels.

docs/concepts/configurations.md
cbarrete: I see what you mean, I agree.
I've incorporated this in the commit, let me know if you have more comments.

docs/concepts/configurations.md
cbarrete: Right, I agree with you.

I guess one could use incoming transitions instead, but `set_cfg_modifiers()` and a macro would probably the way to go.

Resolving this as there's no action needed here anymore.

cbarrete (commented):
@cormacrelf I've left some threads open in case you wanted to comment further. Can you please resolve them if you have no more concerns about them?

docs/rule_authors/configurations.md
```diff
@@ -4,237 +4,196 @@ title: Configurations
...
+compiler or other build tool would be an execution dep. This includes
+all exe [string parameter macro](./string_parameter_macros.md) deps (for
+example, `$(exe //:tool)`) and includes all `attrs.exec_dep()` deps.
+
+An exec dep differs in two ways from a normal dep:
+
+1. It will inherit the execution platform of its dependent instead of the target
+   platform
+1. A dependent's execution platform will be selected so that all exec deps are
+   target compatible with it.
```
cormacrelf: For # 1, maybe "Its target platform will be set to the resolved exec platform of the dependent. If it's going to run there, it needs to be built for that platform. Normal deps simply inherit the target platform."

For # 2, "It influences which exec platform is chosen for a dependent target. This is covered in exec platform resolution below."

Add after:

> Aside from the way they interact with dependents, exec deps are regular targets in the build graph. They may themselves be compiled using their own `exec_dep`s, and therefore may need to select their own exec platform based on their own exec deps. Each time a target somewhere in the build graph has `exec_dep`s, Buck will do another transition through exec platform resolution.
>
> You might not notice an incorrectly typed dependency edge if your only registered execution platform = your target platform = your host machine and you don't do much build configuration, but it matters once you start writing your own build tools and compiling for platforms other than your host machine. The typical error when you have misconfigured is "exec format error" on Linux, where Buck is trying to execute e.g. a Windows executable on a Linux machine.

docs/rule_authors/configurations.md
cormacrelf: Your statement at the end "execution platform resolution sees through them" is the most powerful intuition. You can go a step further and describe them as macros for groups of exec deps:

> In addition to `attrs.exec_dep()`, there is also `attrs.toolchain_dep()`. Toolchain deps must always point to an instance of a toolchain rule. A toolchain rule is much like a macro for adding one or more exec deps altogether, configuring them a little, and storing them in a convenient provider structure. This is useful because often a single programming language has a compiler, a linker, an assembler, a linter, a static analyzer and more, all in separate executables, many of which will call the others in normal operation. A toolchain dep allows you to add all of these at once.
>
> ~~In other words, `attrs.toolchain_dep()` is like a mix of ...~~
> The intuition that toolchains are like macros matches their two main properties:
> 
> 1. Toolchain deps have the same target platform as whatever uses them. So if you `select()` in the parameters to a toolchain rule, you match on how the *dependent* target was configured. Many toolchain rules allow you to set defaults or base flags for a given target platform in this way.
> 2. Toolchain deps are invisible to exec platform resolution of the dependent target, so that the exec deps of the toolchain act as if they are attached directly to the dependent target. They then participate in exec platform resolution for the dependent: Buck finds an exec platform for the dependent that is compatible with all the tools in the toolchain.
>
> This has many benefits:
> - It saves you from having to put the same few `attrs.exec_dep()`s on a bunch of different rules for the same programming language.
> - It "delays" the exec_dep transition and provides a point of configurability before this happens.
> - It bundles all configurability into one spot. The user can instantiate a toolchain at a known location (usually `toolchains//:languagename`) that includes exec_deps or paths to binaries and configures it all in one go.
>
> Toolchain rules are created by passing `is_toolchain_rule = True` to the `rule()` constructor.

docs/rule_authors/configurations.md
```diff
@@ -288,8 +247,73 @@ def target_compatible_with(target, cfg):
...
+
+```python
+target(
+    name = "A",
+    toolchain = attrs.toolchain_dep(default = ":B"),
+)
+target(
+    name = "B",
+    tool = attrs.exec_dep(default = ":C")
+)
```
cormacrelf: This is not a particularly possible example because the `target` rule must choose whether it's a toolchain rule or not. Make :B ` toolchain()` instead of `target()`.

cormacrelf:
For some reason Github won't let me resolve any threads so anything I didn't comment on further is lgtm

docs/rule_authors/configurations.md
cbarrete: Agreed, thanks. I've cut out a sentence out of 1. as I don't think it made the point any clearer.

docs/rule_authors/configurations.md
cbarrete: > hence, it must be recorded in the configured target label

No idea about that, I didn't write that, that was just moved around.
I agree that it didn't make sense, removed.

For the rest, we already have a toolchains page (2 actually, a user facing one, and one for rule authors), so I've deduplicated some of the information. Thanks for the suggestion!

cbarrete:
@scottcao @Nero5023 I think that this is ready, would you mind taking a look?

cbarrete:
@alexlian / @JakobDegen maybe..?

scottcao:
Sorry for the delay here.
This looks very good. I do need to make a few small edits. There are certain things that we're moving away from and some differences in our internal setup that I need to add a few callouts for. I will go ahead and do that internally and merge the PR when done (probably next week since I will be out on PTO later this week). 
Thanks again for putting this together!

cbarrete:
Sounds good, I'm looking forward to it!

PR #1196, opened 2026-01-07
fix(oss): update prelude to support OSS folly build with buck2

ben--:
This PR updates several parts of the prelude so that buck2 properly supports an OSS folly build.

The resulting version of buck2 is demonstrated working in [folly #2542](https://github.com/facebook/folly/pull/2542) which is a non-landable version of [folly #2541](https://github.com/facebook/folly/pull/2541) that shows the working builds.


ben--:
Not sure if the internal build has hung or if it is just taking its time.  But, if there are issues, I'm happy to work on them if they can be shared here.

PR #1213, opened 2026-01-22
use a connection pool for remote execution

bvinc:
Implement a connection pool for all remote APIs.  This can be configured using min_connections, max_connections, and max_concurrency_per_connection.  It roughly is a round robin algorithm, and the connection pool can grow when connections are maxed out with requests.  Once all connections are maxed out, the pool gives out channels that will provide back pressure when they are attempted to be used.


bvinc:
We've been using connection pool code for a few days.  There is one problem that we've noticed.  Previously, the number of actions that are executing remotely had an implicit limit.  Since all of the actions went over a single connection, if the HTTP2 server limits the number of concurrent streams to, say, 128, then this constituted an implicit limit on the number of jobs that could be executed remotely.  With this connection pool, there is no such limit.

I'll probably make another PR to implement this execution semaphore.

Ralith:
Multiple connections to the same remote host usually isn't optimal from a congestion control standpoint. Would it make more sense to raise the server's concurrent stream limit? If it accepts multiple connections, clearly it's willing to tolerate larger numbers of streams in practice.

bvinc:
@Ralith I disagree, or maybe I'm not sure what you mean by "from a congestion control standpoint".  BTW, in practice, most HTTP2 servers limit the number of concurrent streams per connection to around 100 or 128.

A single stream can be limited by latency and window sizes and packet loss.  Currently, each GRPC client (CAS, action cache, executor, and bytestream) has its own TCP connection.  So even multiple large file materializations all go through a single connection, where missing a single packet can stall everything.  If you really want to use all of your bandwidth, you need to use multiple streams.  This is especially true if you have high latency or any packet loss at all between you and your build servers.

I think using multiple streams is fairly standard advice.  For example, it is standard advice that if you want to download from s3 at full speeds, to use multiple TCP streams.  The `aws s3` command line tool downloads defaults to chunking downloads and uses multiple connections.

For another example, bazel uses a similar growing connection pool, and I've chosen similar defaults.

It's configurable though, so if you really wanted to use a single TCP connection instead of 4, with this PR, you can do that.

Ralith:
I thought your intent was to increase concurrency, but I see that I misunderstood. Mitigating TCP head-of-line blocking makes sense. Supporting HTTP/3 might be a more graceful solution in the long term.

> I disagree, or maybe I'm not sure what you mean by "from a congestion control standpoint"

Because each connection's congestion controller operates independently, without knowledge of the others, they can be slower to converge, and to re-converge after a disturbance. As you say, though, the tradeoff can still make sense.

remote_execution/oss/re_grpc/src/pool.rs
```diff
@@ -0,0 +1,459 @@
...
+                    address,
+                    self.config.min_connections,
+                    self.config.max_connections,
+                    self.config.max_concurrency_per_connection,
+                    self.channel_config.clone(),
+                )?;
+                pools.insert(address.to_owned(), Arc::new(host_pool));
+            }
+
+            Arc::clone(pools.get(address).unwrap())
```
scottcao: This is nit since it shouldn't happen, but we typically use expect instead of unwrap so that error message is better when something unexpectedly does go wrong.

remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -1060,6 +926,55 @@ impl REClient {
         self
     }
 
+    async fn acquire_bytestream_client(&self) -> anyhow::Result<ByteStreamClient<GrpcService>> {
+        let address = self.opts.cas_address.as_ref().context("No CAS address")?;
+        let channel = self.pool.get(address).await?;
+        let interceptor = InjectHeadersInterceptor::new(&self.opts.http_headers)?;
```
scottcao: The internal AI review mentioned this, which I thought was interesting:
> InjectHeadersInterceptor is created on every client acquisition, which means parsing headers repeatedly. Consider caching the interceptor or validating headers once during pool creation for better performance.

bvinc:
I updated 3 things:
* Fixed a very real bug where I was not setting `.max_decoding_message_size(...)` on the cas and bytestream client.  This could cause errors about message size being greater than 4MB.
* Made `ChannelPool::get use` the entry API and not call .unwrap()
* Cached the addresses and InjectHeadersInterceptor in the REClient, which gets rid of REClient holding on to the "opts"

I still think there needs to be a max_concurrent_executions, which I guess should be configured on the re_client, before this gets merged.  I'll try to make that PR today and link it here.

bvinc:
I made #1219 to create a configurable remote execution limit.  As I stated earlier, I think it's important to get #1219 in before this PR, because before this PR, there was an accidental execution limit when hitting the max number of concurrent HTTP2 streams.  The connection pool sidesteps this limit, so it would be good if some limit exists before using the connection pool.

bvinc:
I've rebased my changes.  So far, I've been using these changes successfully for a while now without any problems.

bvinc:
I realize that there is a correctness issue to simply adding a connection pool.  These connections could stay idle, get reset, and then be used later and cause a "broken pipe" issue.  I was never once able to reproduce this, as the remote execution client almost always gets destroyed and recreated between build commands.  Furthermore, using more connections gives greater exposure to network-related issues like routers resetting connections.  I have worked around this in 2 ways:
1.  http2 keep-alives are no longer configurable, but always enabled.
2. I've added request-level retries to everything except execute_with_progress.

It's unfortunate that the PR has gotten even bigger, but I think it is necessary for correctness, and retries are something that should be added to buck2 anyway.  If it is preferred, I can make a PR with only retries first, and try to land the connection pool afterwards.

Another change I made was to re-use the Endpoint struct, which is a performance improvement as we no longer need to read and configure TLS for each connection.

ndmitchell (approved):

PR #1219, opened 2026-01-28
add execution_concurrency_limit to RE client

bvinc:
Add a configurable execution_concurrency_limit to buck2_re_client that limits the number of remote executions that can happen concurrently.

I'm open to any sort of debate about what the default should be.  I randomly picked 400.  fbcode likes to use 4000.  Bazel likes to use 400.

It looks like the fbcode client probably internally already has a semaphore, but this would put a semaphore on the outside.  If you guys think it would be better for me to move the semaphore into the OSS client, let me know.

### DIFF
--- a/app/buck2_execute/src/re/client.rs
+++ b/app/buck2_execute/src/re/client.rs
@@ -479,6 +479,9 @@ struct RemoteExecutionClientImpl {
     /// How many simultaneous requests to RE
     #[allocative(skip)]
     cas_semaphore: Arc<Semaphore>,
+    /// How many simultaneous execute requests to RE
+    #[allocative(skip)]
+    exec_semaphore: Arc<Semaphore>,
     /// How many files we can be downloading concurrently.
     #[allocative(skip)]
     download_files_semapore: Arc<Semaphore>,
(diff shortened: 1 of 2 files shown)


cjhopman:
Is there something about grpc that makes 400 a better default than 4000? Or do we think that a lower limit here is more appropriate for oss users?

We use at least up to 10000 with buck2, so it generally has no problem with a higher limit.

bvinc:
To answer your question, no, I do not believe that there is any reason 400 is a better default for grpc than 4000.

The reason I came up with 400 is because our HTTP2 servers had a limit of 400 concurrent streams per connections, and so buck2 was accidentally limiting itself to 400 because all executions were going over a single http2 connection.  Other people's HTTP2 servers might have limits of 100 or 128.

As for what makes a good default, I really think it's a function of your remote execution cluster.  A big company with many executors probably wants to use a big number.

At my company we have a limited number of executors.  I had many users running with an unlimited number of remote executions, and it was causing queue times, which slowed builds for everyone.  A lower number allowed everyone instance access to remote executions.

Issue #1217, opened 2026-01-27
Debugging the action graph is really hard


lf-:
There's several questions I want to ask about the action graph that are currently really hard to answer.

## Why does $build work and $build' not work? or miss cache?

Often I am making some change and I want to know what it did to the build graph. Something like `nix-diff` would be helpful. I think that target determinator does something like this, kind of, but it's not really the same at all.

A major use case for this is "why did this build not hit cache?": I want to be able to get the action digest before and after for each action in the build of some target and diff those trees, ideally aligned with the buck2 level graphs. It is acceptable that this is a file export from each one that is then passed to another command.

## What args did the compiler get?

`aquery` doesn't necessarily materialize arg files, so you have to, I guess, use `buck log what-ran`, except if it hit cache and didn't actually run.

It's hard to directly ask for the args of the exact action you want, which is partially because actions aren't nameable. It's also not really possible to request that said action be built in isolation with a single command line that I know of.

## What Starlark or build command produced this aberrant output?

A problem I have pretty often is that I have some path like so: `buck-out/v2/gen/root/200212f73efcd57d/__app_settings__/haskell_stubs_shared.argsfile`, and I would like to know without extensive source code inspection where it came from in Starlark *or at least* what *action args* produced it.

I would also like to know what its dependencies are. An example type of problem necessitating this was that our `haskell_link` actions had a bug in them where they were missing some `cmd_args` dependencies, and we had to stare at the Starlark code in question to figure out where it came from and what deps were listed there.

Imaginary syntax:
```
$ buck aquery buck-out/v2/gen/root/200212f73efcd57d/__app_settings__/haskell_stubs_shared.argsfile
```

## What deps did my action have recorded for it?

It's very easy to e.g. hit #1151 or similar and not be able to debug it easily. There is theroretically a way of exporting a graphviz graph, but it's not interactive (and seems to be mutually exclusive with showing the action details with `-A`), so it's not especially helpful IMO.

Even the verbose `-A` output doesn't mention the dependencies of each action in the list, so even if I did do a lot of ctrl-F, the info is just not there. I also can't look up more details after looking at the graph visually somehow, because actions aren't nameable.

There is `--output-format html` but there is some unclear-to-me Meta reason that's not released: #973.

The problem that I think exists here is that actions aren't nameable (not entirely true! their output paths are named! but you can't look anything up by output path), so it's not possible to first find the problem action and *then* examine it more closely. If this were Nix, I would be able to *very* easily do this:

```
[...]

Issue #1027, opened 2025-07-07
OSS workflow: not possible to run `//tests/core/help` so docs PRs need FB internal fixups


lf-:
This is because `tests` is in `project.ignore`.

There's golden tests in there which can't be run outside FB, so they are ~guaranteed to fail for external contributions.

- https://github.com/facebook/buck2/pull/1013
- https://github.com/facebook/buck2/pull/1025

At least one proximate cause is https://github.com/facebook/buck2/issues/1026, which is that the pytest macros being used by the buck2 test suite are not available externally.

PR #1227, opened 2026-02-10
bxl: get output artifacts for an Action

lf-:
- Rename action to bxl.Action so it is findable in docs (I don't know if that was what was preventing docs generation, but it's worth a try!)
- Add a method to get the output artifacts for an action in bxl. This allows selectively rebuilding the action by ensuring those outputs.

I added a test, but I have no way to run those besides www.metacareers.com: https://github.com/facebook/buck2/issues/1027

This is part of making action graph debugging slightly easier, which I want to prototype by writing some rather nasty BXLs: https://github.com/facebook/buck2/issues/1217

Unfortunately I *think* there's no way to get the dependencies after the fact, since I *just* have a RegisteredAction. I believe that I can probably do something very bad with owner() then rerunning the analysis to get deps though.

### DIFF
--- a/app/buck2_bxl/src/bxl/starlark_defs/nodes/action.rs
+++ b/app/buck2_bxl/src/bxl/starlark_defs/nodes/action.rs
@@ -12,9 +12,11 @@ use std::convert::Infallible;
 use std::sync::Arc;
 
 use allocative::Allocative;
+use buck2_artifact::artifact::artifact_type::Artifact;
 use buck2_build_api::actions::RegisteredAction;
 use buck2_build_api::actions::query::ActionQueryNode;
 use buck2_build_api::actions::query::OwnedActionAttr;
+use buck2_build_api::interpreter::rule_defs::artifact::starlark_artifact::StarlarkArtifact;
 use buck2_core::deferred::base_deferred_key::BaseDeferredKey;
 use buck2_error::buck2_error;
 use buck2_interpreter::types::target_label::StarlarkConfiguredTargetLabel;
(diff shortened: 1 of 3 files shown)


app/buck2_bxl/src/bxl/starlark_defs/nodes/action.rs
```diff
@@ -90,6 +92,17 @@ fn action_methods(builder: &mut MethodsBuilder) {
...
     }
+
+    /// Gets the artifacts built by this action.
+    fn outputs<'v>(this: StarlarkAction) -> starlark::Result<Vec<StarlarkArtifact>> {
+        Ok(this
+            .0
+            .action()
+            .outputs()
+            .iter()
+            .map(|a| StarlarkArtifact::new(Artifact::from(a.clone())))
```
Nero5023: Here we can use `a.dupe()`.

tests/core/bxl/test_audit_data/audit.bxl
```diff
@@ -15,6 +15,7 @@ def _audit_output_action_exists_impl(ctx):
     action = ctx.audit().output(buck_out)
 
     asserts.equals(action.owner(), target.label)
+    asserts.equals(actions.outputs()[0].basename, "with_output.txt")
```
Nero5023: Here is a typo, should be `action` not `actions`

Nero5023:
Hey, thank you for the PR!

Overall lgtm! Only twos nit I commented above, could you resolve those and do a rebase?

lf-:
@Nero5023 I've fixed and rebased, sorry about the delay. Would you be able to take a look at the other two PRs in this series also?

PR #1234, opened 2026-02-16
fix: allow `str` in `_lintify` signature

JonasKruckenberg:
The current `_lintify` signature only allows `ResolvedStringWithMacros` with afaik you cannot actually pass to it. This means a custom Rust toolchain cannot actually set `allow_lints`, `deny_lint`, etc.

this changes fixes this by just changing the signature to `str | ResolvedStringWithMacros` similar to other functions in the prelude.


dtolnay (approved):
Thanks!

PR #1244, opened 2026-02-18
fix: warn when --include/--exclude are used suspiciously

### LINKED ISSUE #1243: `buck test --include foo //...` confusingly silently skips all tests
## Description

When using `buck test` with the `--include` flag, the position of the flag relative to the target pattern affects whether tests are matched.

This is probably intentional behaviour but it is *extremely* confusing and merits a warning at minimum (and quite possibly revision).

## Setup

Create a directory `//foobar` in this repo:

BUCK:
```starlark
sh_test(
    name = "test_with_label",
    test = "test.sh",
    labels = ["some_test_label"],
)

sh_test(
    name = "test_without_label",
    test = "test.sh",
)
```

test.sh:

```
#!/bin/bash
# Simple passing test
exit 0
```

## Reproduction

Repro'd on: a72a88b6c897bfd3d97b621240a93718d56f92f4

### Weird: Flag before target pattern (matches no tests)

```
cargo run -p buck2 -- test --include some_test_label //foobar/...
Tests finished: Pass 0. Fail 0. Fatal 0. Skip 0. Omit 0. Infra Failure 0. Build failure 0
NO TESTS RAN
```

### Working: Flag after target pattern

```
cargo run -p buck2 -- test //foobar/... --include some_test_label
✓ Pass: gh_facebook_buck2//foobar:test_with_label (0.1s)
Tests finished: Pass 1. Fail 0. Fatal 0. Skip 0. Omit 0. Infra Failure 0. Build failure 0
```

## Verification

I didn't goof up the labels AFAIK:

```
cargo run -p buck2 -- uquery //foobar:test_with_label --output-attribute labels
{
  "gh_facebook_buck2//foobar:test_with_label": {
    "labels": [
      "some_test_label"
    ]
  }
}
```

## Expected Behavior

Either:
- Both commands should produce the same result
- The one that silently does the surprising thing of skipping all tests should print a warning

lf-:
This is a workaround to not break users, but still make the UX okay. I suspect that inside fb, there are lots of people that do: buck test //... --include myproject myotherproject

I chose not to error out on this case, since I suspect scripts might do `["buck", "test", *targets, "--include", *labels]`, which is valid for no targets and shouldn't be an edge case.

Fixes: https://github.com/facebook/buck2/issues/1243

Test plan with reproducer in //foobar:
```
$ target/debug/buck2 test --exclude meow //foobar/...
No target patterns specified, but --include/--exclude flags are set.
This is likely a mistake: put targets first, then --include/--exclude, since include/exclude consume all remaining args
hint: The following requested labels look like target patterns: '//foobar/...'
hint: Try putting them before --include/--exclude.
hint: For example: buck2 test //foo --include mylabel

Build ID: 925dd280-7479-4ddb-82d8-2f0773b40412
Command: test.
Time elapsed: 0.0s
Tests finished: Pass 0. Fail 0. Fatal 0. Skip 0. Omit 0. Infra Failure 0. Build failure 0
NO TESTS RAN

$ target/debug/buck2 test //foobar/... --exclude meow
✓ Pass: gh_facebook_buck2//foobar:test_without_label (0.0s)
---- STDOUT ----

---- STDERR ----

✓ Pass: gh_facebook_buck2//foobar:test_with_label (0.0s)
---- STDOUT ----

---- STDERR ----

Build ID: a979dac1-d6b0-4cab-ae66-98020371b729
Loading targets.   Remaining     0/9                                                                        80 targets declared
Analyzing targets. Remaining     0/16                                                                       1 actions, 1 artifacts declared
Command: test.
Time elapsed: 0.3s
Tests finished: Pass 2. Fail 0. Fatal 0. Skip 0. Omit 0. Infra Failure 0. Build failure 0
```

### DIFF
--- a/app/buck2_client/src/commands/test.rs
+++ b/app/buck2_client/src/commands/test.rs
@@ -8,6 +8,8 @@
  * above-listed licenses.
  */
 
+use std::fmt::Write;
+
 use async_trait::async_trait;
 use buck2_cli_proto::CounterWithExamples;
 use buck2_cli_proto::TestRequest;
@@ -75,6 +77,29 @@ fn print_error_counter(
     }
     Ok(())
 }
+
+/// Check if we should warn about potentially misplaced --include/--exclude flags.
+/// Returns Some(suspicious_labels) if warning should be shown, where suspicious_labels
+/// are label values that look like they might be target patterns (contain '/' or ':').
+fn should_warn_about_flag_position(
+    patterns: &[String],
+    include: &[String],
+    exclude: &[String],
+) -> Option<Vec<String>> {
+    if patterns.is_empty() && (!include.is_empty() || !exclude.is_empty()) {
+        let suspicious_labels: Vec<String> = include
+            .iter()
+            .chain(exclude.iter())
+            .filter(|label| label.contains('/') || label.contains(':'))
+            .cloned()
+            .collect();
+
+        Some(suspicious_labels)
+    } else {
+        None
+    }
+}
+
 #[derive(Debug, clap::Parser)]
 #[clap(name = "test", about = "Build and test the specified targets")]
 pub struct TestCommand {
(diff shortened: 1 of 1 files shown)


lf-:
We could get 90% of this turned into a hard error if we forbade colons and slashes in labels. There's still some target patterns which don't have those, iirc, and those can't be hard errors, but the ones that *are* considered "suspicious" here can be an error.

Filing a discussion ticket https://github.com/facebook/buck2/issues/1245

scottcao (approved):

PR #1250, opened 2026-02-23
bxl aquery: allow passing in bxl.Action

lf-:
This allows for fine grained action graph debugging by letting bxl ask
about individual Actions' dependencies.

I chose this API because it's the smallest API surface change I could
think of to improve this.

One thing which Claude pointed out is that you can't have heterogeneous
lists of Actions and aquery nodes. This is sorta unfortunate.

Allows users to implement "What deps does my action have recorded for
it" from https://github.com/facebook/buck2/issues/1217 themselves.

Note that I tried running the tests I could using buck bxl but
I have not gone to https://metacareers.com/ to be able to run the python
side.

Example usage:
Use `ctx.audit.output()` to obtain an Action object and then ask about
its dependencies.

Stacked on:
- https://github.com/facebook/buck2/pull/1227

### DIFF
--- a/app/buck2_build_api/src/query/bxl.rs
+++ b/app/buck2_build_api/src/query/bxl.rs
@@ -12,6 +12,7 @@ use std::future::Future;
 use std::pin::Pin;
 
 use async_trait::async_trait;
+use buck2_artifact::actions::key::ActionKey;
 use buck2_core::cells::CellResolver;
 use buck2_core::cells::name::CellName;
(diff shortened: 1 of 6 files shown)


8Keep:
@lf- could you rebase

lf-:
done!

8Keep:
Doesn't quite build on master - probably due to some changes with pointers/copying recently for mem improvements, should be an easy fix though

PR #1251, opened 2026-02-23
bxl: allow users to make atarget sets

lf-:
Previously this was just not exposed. Note that there may be reason to yak shave the name, idk.

Stacked on:
- https://github.com/facebook/buck2/pull/1227
- https://github.com/facebook/buck2/pull/1250


8Keep:
The two prs underneath are merged, so this one could be rebased whenever you're ready.

lf-:
@9999years can you? i um.. medical leave. thanks!

PR #1275, opened 2026-03-29
add UEFI OS constraint value

JonasKruckenberg:
This change adds a `//os/constraints:uefi` value (and `//os:uefi` alias) to support building UEFI applications using buck2. ofc UEFI is only an OS in the wider sense (and it is kinda like weird windows), but it is how other toolchains (e.g. Rust and LLVM) treat UEFI.


theoparis:
Is there a reason this was closed? I was looking to use this for my own osdev project.

PR #1278, opened 2026-04-01
Remove compilation_database.bxl include path hack

cbarrete:
This was a bad hack, which was broken in the presence of header namespaces (in which case `-I .` would not put the actual headers in the include path, causing random jumps to `buck-out` anyway).

The correct fix is to use header maps, which integrate correctly with Clang-based tooling, and provide accurate jumping to the source tree. This commit also adds some documentation about this.

### DIFF
--- a/docs/users/how_tos/compilation_database.md
+++ b/docs/users/how_tos/compilation_database.md
@@ -3,7 +3,13 @@ id: compilation_database
 title: Compilation databases
 ---
 
-# Generating compilation databases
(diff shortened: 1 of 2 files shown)


cbarrete:
@cjhopman, do you think you could get someone internally to take a look at this?
The CI failures are unrelated.
Thanks!

cbarrete:
@Nero5023 maybe?

scottcao (approved):

PR #1281, opened 2026-04-02
Make HashedConfigurationPlatform equality consistent with Hash

bvinc:
HashedConfigurationPlatform had a manual Hash impl using only output_hash, but derived PartialEq/Eq/Ord that compared all fields including is_exec_platform (via ConfigurationPlatform). This inconsistency caused the static interner to create separate entries for the same platform used in exec vs non-exec context.

Fix: implement PartialEq/Eq/Ord/PartialOrd manually using only output_hash, consistent with the existing Hash and StrongHash impls.

### DIFF
--- a/app/buck2_core/src/configuration/data.rs
+++ b/app/buck2_core/src/configuration/data.rs
@@ -422,16 +422,7 @@ impl ConfigurationDataData {
     }
 }
 
-#[derive(
-    Debug,
-    Eq,
-    PartialEq,
-    Ord,
-    PartialOrd,
-    Allocative,
-    derive_more::Display,
-    Pagable
-)]
+#[derive(Debug, Allocative, derive_more::Display, Pagable)]
 #[display("{}", full_name)]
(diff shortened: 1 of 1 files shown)


bvinc:
I believe this is a regression that was introduced when is_exec_platform was added to `ConfigurationPlatform` that occurs with the default behavior of hash_cfg_with_exec_platform==false.

The downstream effects of this that I saw were pretty convoluted.  We had a BXL script, that when executed, would execute the same AnonTarget twice, at the same time, with the same output directory, and they would clobber each other.  Apparently this same action, was getting two different DICE indices, because they were "unequal" because they differed in is_exec_platform, even though they were equal in every other way, including their ActionDigest.

JakobDegen (changes requested):

JakobDegen:
This is fixed on master via another approach (`is_exec_platform` is gone), so closing. Reopen if you're still having problems

PR #1283, opened 2026-04-03
starlark: Export the `RecordTypeMatcher` for users in the public API

### LINKED ISSUE #120: disambiguating record types defined in different files

PhilipMetzger:
This fixes https://github.com/facebook/starlark-rust/issues/120. Since it allows users now to compare records from different files.

### DIFF
--- a/starlark-rust/starlark/src/values/types/record.rs
+++ b/starlark-rust/starlark/src/values/types/record.rs
@@ -49,4 +49,5 @@ pub(crate) mod record_type;
 pub(crate) mod ty_record_type;
 
 pub use crate::values::record::instance::Record;
+pub use crate::values::record::matcher::RecordTypeMatcher;
 pub use crate::values::record::record_type::RecordType;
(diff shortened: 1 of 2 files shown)


PhilipMetzger:
cc @ndmitchell for some visibility. I also am writing a patch for https://github.com/facebook/starlark-rust/issues/144 so it'd be nice if these could also be included in the new starlark-rs release.

ndmitchell (approved):

ndmitchell:
This has CI failures because other pieces are still private.

PhilipMetzger:
> This has CI failures because other pieces are still private.

Should be fixed now.

PhilipMetzger:
ping on this for another import.

PhilipMetzger:
ping again

scottcao (approved):

PR #1286, opened 2026-04-09
Track symlinks between local files via DICE dependencies

cormacrelf:
### Problem

Here's a test case that I was quite alarmed Buck fails at. The use case is reading files at bzl eval time that don't have a .toml suffix, as TOML, for https://github.com/cormacrelf/elk which I have updated to depend on this entire thing at the heart of its strategy only to realise it doesn't work very well at all. (I would be happy with a way to tell load() to load as TOML if we can't get symlinks to work.)

```sh
BUCK
blah.lock
blah.lock.toml -> blah.lock   [a symlink]
```

```toml
# blah.lock
value = "original"
```

```starlark
# BUCK
load(":blah.lock.toml", "value")

genrule(
    name = "test",
    out = "out.txt",
    cmd = "echo {} > $OUT".format(value),
)
```

Now, buck *does not pick up changes to blah.lock*.

```sh
buck2 build --out=- :test
{value: original}

echo 'value = "changed"' > blah.lock

buck2 build --out=- :test
{value: original}

# wat?
# now you have to buck2 clean to get it to pick up changes...
```

### Solution

Support invalidating through symlinks.

I am not very familiar with the fs operations here, but it's working alright. This is still pretty likely not to be quite correct and I invite some attention on the implementation.

In general, directory symlinks are already handled quite well by `fn read_path_metadata` in `buck2_common/src/io/fs.rs`. The fs metadata we get in the dice compute() function is really well designed, so whoever wrote that, great work. E.g. we get the first directory symlink found when navigating from the project root, with the rest as a forward relative path. This maps very nicely onto the recursive dice dependencies; we just resolve that first dir symlink and keep going. This was all very easy to actually implement, and the real code change is tiny.

The tests are more involved.

Things to think about

- Cross cell symlinks -> the cell is included in all these dice keys. No problem.
- External symlinks -> buck won't be receiving any watchman updates for external symlinks, so we basically don't change behaviour at all for these, DICE treats them as it did all symlinks before, having no knowledge they are anything but local files/dirs. As before, symlinks get resolved where FileOps is ultimately implemented: `fn read_path_metadata` in `buck2_common/src/io/fs.rs`.
- **Recursive symlinks** should cause buck2 errors. **Done**. This one was extra tricky. Claude has obviously read the Bazel codebase, it tried to rely on CellPath sorting a/b before a-b, which it does not. I added a test for catching recursion in a contorted case like this. We can actually make ForwardRelativePath's PartialOrd/Ord impls do that, it would be convenient in this exact spot but blast radius is big enough already.

Testing

- This has unit tests for which the harness had to be significantly upgraded to behave more like the real world IO ops. It is neat that we have something pretty closely approximating what a filesystem will do, but that is obviously quite dependent on the test harness being exactly right.
- We basically need integrations test versions of all the tests I added. I can't run those, so maybe someone at Meta can translate em.
- For what it's worth, if you run this branch, it does invalidate through symlinks as designed and does catch symlink recursion etc.

### DIFF
--- a/app/buck2_common/src/file_ops/dice.rs
+++ b/app/buck2_common/src/file_ops/dice.rs
@@ -38,6 +38,7 @@ use crate::file_ops::delegate::get_delegated_file_ops;
 use crate::file_ops::error::FileReadError;
 use crate::file_ops::error::extended_ignore_error;
 use crate::file_ops::metadata::RawPathMetadata;
+use crate::file_ops::metadata::RawSymlink;
 use crate::file_ops::metadata::ReadDirOutput;
 use crate::ignores::file_ignores::FileIgnoreResult;
 use crate::io::ReadDirError;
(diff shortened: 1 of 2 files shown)


cormacrelf:
I spotted another test case to add, writing down as a TODO

`a/b/c.txt` where `a/b -> e` and `e -> /nix/store/...`. Retargeting `a/b` elsewhere should be picked up. I believe this test will pass already...


Ah never mind. That would only test the test harness.

cormacrelf:
I guess this doesn't deal with uploads to remote execution. I'll look into that.

JakobDegen:
Looking at this now. You're right that there's a problem here, and that these bugs need to be fixed, but two important details:
 1. At least initially, but quite possibly also permanently, the direction of this bug fix is not going to be towards "handling" this, but rather towards banning symlinks in most cases. While it would technically be imaginable to support symlinks for things such as `load`s, the semantics of that are quite unclear and actually very difficult to get right, and even if we do that then people will start wanting it on BUCK files which is yet far more unclear
 2. Doing this only for `read_file` isn't enough, we have to do it for all the other IO operations too

I actually am working on this right now and expect to have the symlink-banning in place somewhat shortly (weeks). So I'm going to close this PR for now. If you want symlink support we can possibly discuss that elsewhere, but also "read a non-toml file as toml" seems like a reasonable ask too.

cormacrelf:
Fair enough! I do hope you don't fully ban external symlinks, there are some pretty legit use cases for making Nix work automatically on RE that Jade has been exploring (see discord).

JakobDegen:
@cormacrelf external symlinks won't be banned, and neither will symlinks as source artifacts - we're just going to ban symlinks in the paths of BUCK/bzl files and cases similar to that

PR #1287, opened 2026-04-09
Support `load("uv.lock?as=toml", "value")` syntax

cormacrelf:
Alternative to #1286.

> The use case is reading files at bzl eval time that don't have a .toml suffix, as TOML, for https://github.com/cormacrelf/elk which I have updated to depend on this entire thing at the heart of its strategy only to realise it doesn't work very well at all. (I would be happy with a way to tell load() to load as TOML if we can't get symlinks to work.)

Here's that alternative. The example becomes

```starlark
# BUCK
load(":blah.lock?as=toml", "value")

genrule(
    name = "test",
    out = "out.txt",
    cmd = "echo {} > $OUT".format(value),
)

# blah.lock
value = "original"
```

and basically this works. This code I *do* understand quite well, and it's simple and there are no X factors as to how to get it to work.

### DIFF
--- a/app/buck2_core/src/bzl.rs
+++ b/app/buck2_core/src/bzl.rs
@@ -31,6 +31,17 @@ enum ImportPathError {
     Suffix(CellPath),
 }
 
+/// The format a loadable file should be parsed as, when it can't be inferred
+/// from the file extension. Set via an `?as=` hint in a `load()` string, e.g.
+/// `load(":uv.lock?as=toml", "value")`.
+#[derive(
+    Clone, Copy, Hash, StrongHash, Eq, PartialEq, Debug, Allocative, Pagable
+)]
+pub enum ImportFormat {
+    Json,
+    Toml,
+}
+
 /// Path of a `.bzl` file.
 #[derive(Clone, Hash, StrongHash, Eq, PartialEq, Debug, Allocative, Pagable)]
 pub struct ImportPath {
@@ -40,6 +51,12 @@ pub struct ImportPath {
     /// The cell of the top-level build module that this is being loaded
     /// (perhaps transitively) into.
     build_file_cell: BuildFileCell,
+    /// If set, the file is parsed as this format regardless of its extension,
+    /// as requested by an `?as=` load hint. `None` means infer from the
+    /// extension. This is part of the path's identity: it rides along wherever
+    /// the `ImportPath` goes (including across the `buck2_node` boundary), so
+    /// the module type never has to be re-guessed from the extension.
+    format: Option<ImportFormat>,
 }
 
 impl ImportPath {
@@ -70,7 +87,41 @@ impl ImportPath {
         Ok(Self {
             path,
             build_file_cell,
+            format: None,
         })
+    }
+
+    /// Create an ImportPath for a file with an explicit format override (e.g. `?as=toml`).
+    /// Skips the file extension validation since the format is determined by the hint,
+    /// not the extension.
+    ///
+    /// Do not include the ?as=toml hint as part of the path.
+    pub fn new_with_explicit_format(
+        path: CellPath,
+        build_file_cell: BuildFileCell,
+        format: ImportFormat,
+    ) -> buck2_error::Result<Self> {
+        if path.parent().is_none() {
+            return Err(ImportPathError::Invalid(path).into());
+        }
+
+        if path.path().as_str().contains('?') {
+            return Err(ImportPathError::Invalid(path).into());
+        }
+
+        Ok(Self {
+            path,
+            build_file_cell,
+            format: Some(format),
+        })
+    }
+
+    pub fn new_same_cell_with_explicit_format(
+        path: CellPath,
+        format: ImportFormat,
+    ) -> buck2_error::Result<Self> {
+        let build_file_cell = BuildFileCell::new(path.cell());
+        Self::new_with_explicit_format(path, build_file_cell, format)
     }
 
     /// LSP creates imports for non-bzl files.
@@ -89,6 +140,7 @@ impl ImportPath {
         Ok(Self {
             path,
             build_file_cell,
+            format: None,
         })
     }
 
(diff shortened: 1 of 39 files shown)


8Keep:
I know how useful this would be, but wonder if anyone can help come up with some safeguards or limitations for this. Hopefully starting a conversation where we can figure out what to do.

This has all kinds of implications on target determination, as an example we have our out of the box version of a target determinator released here: https://github.com/facebookincubator/buck2-change-detector/

You lose a lot of the guarantees that you get with buck: any source file can now be loaded by a buck/bzl file across the whole repo and there's no way of knowing that without generating a full targets graph, parsing it, and finding out the dependency graph of bzl loads, which is very slow for a big heavy repo. Previously, you can skip a lot of computation by assuming the Package Boundary rules of buck are followed, follow certain target determination rules if you touch a BUCK or PACKAGE file, and the whole graph is not needed unless you touch a ".bzl" file.

cc: @lf- who linked me this pr, if you have any thoughts

lf-:
Oh no, I see where you're coming from. I do suspect that the existing toml and json feature causes problems for this as well though. Maybe we could say it's some allowlist of extensions from buckconfig that btd reads?

8Keep:
The json and toml loads do have that same negative effect. I'm pretty sure this was briefly discussed before that merged but I didn't really get involved.. maybe I should have, but maybe now is a good time to try to come up with something that could help protect against the proliferation. I'm drawing a blank right now but maybe something with package boundaries?

This pr does expand it from just json and toml file extensions (technically limited, like .bzl files, it's a known set) to practically unlimited. So maybe we want some buckconfig that declares the allowed mappings.

I'm on too many different thought tracks today to think too hard about it right now. I'm hoping we can come up with something simple to address this TD problem though.

lf-:
I think it would possibly be fairly reasonable to restrict this feature to only within the current package, yeah; although, that idea does have a trouble which is that there's not really any obvious answer to how one would export a value to other packages aside from maybe creating fancy target types so you can reference it as a target (but then you can't really call functions on it to further process it). It is irritating to be able to reference data files anywhere in the repo because the package boundaries help humans too.

bazel has bzl visibility. Maybe you could have something like "must be either defined in the current package or referenced through a bzl in another package which re-exports it"?

This feature is also just generally one where it's alarmingly close to a monadic build action feature, but is just short in that it's merely reading a file (which normally you're not allowed to do in analysis!).

idk. idle thoughts; similarly brain frazzled.

cormacrelf:
I have had a look at BTD and I don't quite see the problem. BTD's targets input includes the parsed import graph. I can't find a way to use BTD without having it call `buck2 targets --imports` at some point. I could not find anything where it looks at .bzl files / BUCK files manually or matches on file extensions etc (aside from ignoring changes to prelude.bzl). BTD appears to be pretty purely `fn(import graph, diff) => target list`. So as long as we include `?as=toml` imports in the buck2 targets output, we're good. We don't have to worry about "all .json files cause big impact" either, because only .json files that get imported in .bzl are included in the graph output.

I added an e2e test to this PR and made sure that the `buck2 targets --imports` produces uv.lock as a leaf node, like a .bzl with no imports. And then I fixed this, because it wasn't being properly emitted.

The other half can be tested for reassurance in BTD with e.g.

```
    #[test]
    fn test_rerun_unimported_lock_does_not_schedule() {
        // An *unrelated* lockfile changes — one that nothing loads. The
        // loading package must NOT be scheduled.
        let base = Targets::new(vec![
            TargetsEntry::Import(BuckImport {
                file: CellPath::new("foo//app/BUCK"),
                imports: Box::new([CellPath::new("foo//lockdir/uv.lock")]),
                package: Some(Package::new("foo//app")),
            }),
            TargetsEntry::Import(BuckImport {
                file: CellPath::new("foo//lockdir/uv.lock"),
                imports: Box::new([]),
                package: None,
            }),
        ]);
        let cells = CellInfo::testing();
        let changes =
            Changes::testing(&[Status::Modified(CellPath::new("foo//lockdir/other.lock"))]);

        let rerun = rerun(&cells, &base, &changes).unwrap().unwrap();
        assert!(!rerun.contains_key(&Package::new("foo//app")));
    }
```

Basically once you get down to BTD it does not look at the file path much at all. It is relying on `buck2 targets --imports` and the shape of the import graph that buck generates, so I don't think we need to change BTD at all. Certainly there could be optimisations in BTD I didn't find. I guess maybe you're referring to optimisations outside of BTD where you decide whether to run BTD at all etc?

8Keep:
So yes you're right on that external BTD impl, however we have a more complicated BTD internally with many fast paths where we don't need to load the full target graph.

However I think I've just realized I think I can make an adjustment there to make this case fast with a trie for all the bzl/toml/jsons loaded for the whole repo and we might not have to worry about this.

cormacrelf:
So this is probably mergeable since you could just not use ?as=toml inside Meta until this has a fast path, right? I can buckconfig-flag it with default OSS-true Meta-false if you want me to. You are probably the person best placed to make a buckconfig dictating what file extensions can be ?as=toml'd in buck in order to ensure the BTD tries remain exhaustive lists of eligible files.

8Keep:
Yes, I can try to merge this tomorrow if you can add in a buckconfig to let users choose which mappings they want to allow

(for now, we could think about removing the buckconfig gating in the future to make it more available)

cormacrelf:
I have added `buck2.load_as_allowlist = uv.lock, *.blah` with globset. Default is none. I also decided to disallow loads from other packages. I had a quick look at https://bazel.build/concepts/visibility#load-visibility and figured, if we want to do that, there will never be any way to reduce visibility of a file for load-as if we allow it from everywhere now. Thanks for linking that @lf-. So basically you can re-export from a .bzl file if you want to use in another package/cell, and we will be able to control visibility from .bzl some day, and there will be no need for a hacky workaround to control as=toml visibility.

The same logic applies to direct JSON and TOML loads so I restricted to package-internal loads for those as well. That is a breaking change but the error message has the easy fix embedded in it.

I also implemented load visibility. It touches many of the same code paths as this PR so I'll hold off submitting for now but it was very simple and will be very nice.

PR #1288, opened 2026-04-09
Make soft error act the same in OSS

bvinc:
I'm going to propose that the soft_error macro should act the same for meta's internal build versus the OSS build.  Particularly, there have been times where benign uncommon situations are a quiet soft error for meta, and a panic for the OSS community.  I think the best situation for the OSS community is identical behavior as the meta build.

### DIFF
--- a/app/buck2_core/tests/soft_error.rs
+++ b/app/buck2_core/tests/soft_error.rs
@@ -15,7 +15,6 @@ use std::sync::Once;
 use buck2_core::error::StructuredErrorOptions;
 use buck2_core::error::initialize;
 use buck2_core::error::reset_soft_error_counters;
-use buck2_core::is_open_source;
 use buck2_core::soft_error;
 use buck2_error::buck2_error;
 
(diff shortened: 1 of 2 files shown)


8Keep:
The difference is that technically, all soft errors we want to eventually turn into a hard error. Soft error is a way to measure it and burn down the usage to zero, though there are probably some soft errors that aren't being worked on internally.

I would prefer if we could keep the soft error logic as a hard error in open source, but instead have a buckconfig or something to allow users to disable certain soft errors from being a hard error, so that:

a) users can know that something they are doing is a bad practice that we want to remove
b) but they have a way to opt out of that soft error breaking them, until it becomes a hard error, if ever.

The goal would be to allow open source users to gradually migrate while upgrading buck2 versions, unlike today where you have to fix your stuff immediately when a soft error is introduced. The approach of making them all invisible to open source though would probably not be ideal, to avoid us just breaking open source when we flip soft errors to hard errors in the same way.

Would you be willing to make that happen?

PR #1295, opened 2026-04-15
Propagate linker_type correctly in the `system_cxx_toolchain`

Nehliin:
Previously the system cxx toolchain would hard code the linker type based off the os. This works fine in most cases but breaks windows gnu builds for example. The `linker_type` is already correctly provided by the `CxxToolsInfo` type but ignored. This pr uses the linker_type from there instead and adds a select to the path clang tools to separate darwin from gnu targets.


Nehliin:
sorry for the random ping but is there a reason for CI not importing this pr? Have I messed something with the signing up? @Nero5023

JakobDegen (approved):

8Keep:
I fixed the formatting a bit for internal formatting and am trying to land this, thanks for the contribution!

Issue #876, opened 2025-03-13
rustdoc HTML roots issues


alexlian:
### Discussed in https://github.com/facebook/buck2/discussions/705

<div type='discussions-op-text'>

<sup>Originally posted by **cormacrelf** June 28, 2024</sup>
This is very minor, but maybe a useful report in relation to the recent commit https://github.com/facebook/buck2/commit/7cc841545add97265d61ceefd0aee8a5f93edee8

> Delete support for --extern-html-root-url 
>
> Use `rustc_flags += ["-Zcrate-attr=doc(html_root_url = {})".format(...)]` in a macro instead.
>
> If we reinstate an `html_root_url_prefix` attribute on the RustToolchainInfo in the future, we might want to consider doing it in a way that supports different prefixes per cell, or by package prefix.
>
> ```
> html_root_url_prefix = {
>     "fbcode": "/intern/rustdoc",
> }
> ```

I recently got a convoluted rustdoc build going. I did in fact mess with `html_root_url_prefix` on RustToolchainInfo, and did pretty much what is described there.

My experience was that it was not close to being flexible enough. Linking to https://docs.rs for third party crates (from `reindeer`) *with the correct version number* is the litmus test for anything like this. Was not feasible with a dict of package prefixes. The issues were:

1. the part of the code that was writing the `--extern-html-root-url` args has access to a RustDependency (or whatever) structure, whose name can be different from the original `rust_library`'s. The name often comes from an `alias` rule. You need to look through aliases to find the crate name that you should generate a link from, otherwise you may be linking to nowhere. This is possible but very hacky. (My code examined one of the .rlib artefacts and got its owner's label.)
2. I ended up writing a miniature string templater for the full URL to put `${label.name}` etc in the right spots, because what the prelude was appending after the `html_root_url_prefix` was not suitable.
3. And for docs.rs, which required *parsing* the label, I basically ended up writing out the full package name wrapped in some easily replaced text, and literally using `sed` on all the HTML to turn that into a `docs.rs` URL. It was bad and slow.

Overall, the single dict of strings on the toolchain approach was terrible. Would not recommend bringing it back. The toolchain was simply the wrong place for it. In the next attempt I will use `-Zcrate-attr` as recommended, and obviously that will be very simple to apply to `reindeer` generated BUCK files. You can just do the `$name-$version` parsing in the macro and put a docs.rs URL in there.</div>

PR #1297, opened 2026-04-17
Rustdoc merging with RFC3662

### LINKED ISSUE #1097: Rustdoc builds should be mergeable using RFC3662
I currently struggle to maintain a big set of hacks to merge rustdoc builds across a crate graph by parsing search index javascript files. It breaks on nearly every rust release. According to the text of [RFC3662](https://rust-lang.github.io/rfcs/3662-mergeable-rustdoc-cross-crate-info.html), Meta does exactly the same thing and struggles just as much.

That RFC, which dramatically improves this situation, was [implemented](https://github.com/rust-lang/rust/pull/129337) more than a year ago now. Has there been any progress switching doc builds to use the unstable `rustdoc --merge=...` flag? There is no mention of this in the prelude, so I guess no. Is it likely you'll accept a PR that implements this or are the required changes too sweeping and you're planning to do it yourselves?

I note that:

-  A merged rustdoc build needs a target set as input. You could have a `rust_doc_bundle` target but it's probably not that useful.
- Therefore the final merge step should be BXL. It would take a target set, request a doc build for each with `--merge=none`, then specify an action to `--merge=finalize` with all the targets' outputs.
- An important feature is the treatment of crates not in the requested target set. I handle this by adding a flag `"-Zcrate-attr=doc(html_root_url = \"{}\")".format(docs_rs_url)` on third-party dependencies via a macro. Might be nice if this were not something Buck users would have to spend 6 hours figuring out on their own. Perhaps an attr on rust_library that reindeer populates automatically.

cormacrelf:
Fixes https://github.com/facebook/buck2/issues/1097. Basically you can now:

- Generate merged rustdoc for all your first-party crates, or any subset of them, via bxl
- With a search index of all of them
- With all srcs and cross-crate trait impls available
- With perfect caching of the single-crate doc builds
- With the option to link to docs.rs for 3rd party crates
- With all rustdoc_flags respected
- With forwards compat for the very recent change from `rustdoc --emit=invocation-specific` to `--emit=html-non-static-files`

Also deals somewhat with the problems in https://github.com/facebook/buck2/issues/876 by making `-Zcrate-attr` easier to use with a simple attr on `rust_library(rustdoc_html_root_url= ...)`. If this is merged, we should update reindeer to generate this automatically.

Also makes rustdoc themes easier to use, by generating stub theme files so we don't force a rebuild of all crate docs when making tiny tweaks to themes. The only thing rustdoc actually looks at until the final merge step is the filename. This applies to regular [doc] subtargets, but they get the real theme files.

All this is exercised in the example project.


cormacrelf:
Maybe @dtolnay would like to review?

There are some comments about how rustdoc resolves external links to docs.rs for third party crates, to the effect of "rustdoc is pretty annoying here" that might be appropriate to PR against rustdoc itself. If its priority order for getting a URL while documenting a downstream crate were not so weird, we could use the --extern-html-root-url override (or however you spell it) and not have to use `-Zcrate-attr=doc(...)`.

The hacky way of adding `-Zcrate-attr` works pretty well for now and does deal with incorrect-but-baked-into-source `doc(html_root_url)` metadata that appears on a fair few third party crates. I understand meta just uses a macro for such purposes but making an attribute on rust_library overall makes this feature way easier to use and allows us to exclude third party crates from doc builds at all in the BXL. If we fixed rustdoc and used flags on downstream deps instead, we would still need the upstream crate to know its URL to put it in a provider. And finally, if we get reindeer generating docs.rs urls from the version info, it will work out of the box.

Maybe the BXL could also do with some work, you are better at BXL than I am.

Only other thing I can think to add is a wrapper script for the BXL that hosts it on localhost or opens an appropriate index.html in the browser.

PR #1303, opened 2026-04-26
Race condition fix

### LINKED ISSUE #1302: Race condition in `NamedSemaphores::get` breaks `OnePerToken` exclusivity
## Summary

`host_sharing::NamedSemaphores::get` is not safe to call concurrently for the
same name. A check-then-act sequence on the underlying `DashMap` allows two
callers to each install their own `SharedSemaphore` for the same key, after
which the "only one in-flight per token" invariant that
`HostSharingRequirements::OnePerToken` and `OnePerTokens` rely on is silently
broken. This is a correctness bug, reachable from the normal
`HostSharingBroker::acquire` code path.

## Where

`host_sharing/src/named_semaphores.rs`, function `NamedSemaphores::get`.

The function looks up `name` in a `DashMap<String, SharedSemaphore>`; if
absent, it constructs a new `SharedSemaphore` and inserts it. The lookup and
the insert are two separate operations, with no lock held across them.

## The race

Two threads, `A` and `B`, call `NamedSemaphores::get("foo")` at roughly the
same time, with no entry yet present for `"foo"`:

1. `A` performs the lookup and observes that no entry exists.
2. `B` performs the lookup and also observes that no entry exists.
3. `A` constructs a fresh `SharedSemaphore`, call it `Sem_A`, and inserts
   `("foo", Sem_A)` into the map. `A` returns a clone of `Sem_A`.
4. `B` constructs a fresh `SharedSemaphore`, call it `Sem_B`, and inserts
   `("foo", Sem_B)` into the map, overwriting `Sem_A`. `B` returns a clone of
   `Sem_B`.

After step 4:

- `A` holds a clone of `Sem_A`.
- `B` holds a clone of `Sem_B`.
- The map only knows about `Sem_B`.

`Sem_A` and `Sem_B` are independent objects, each with one permit. `A` and
`B` can both successfully call `acquire(1)` at the same time. Any later
caller looks up `"foo"` in the map, gets `Sem_B`, and observes its permit as
available whenever `B` is not actively holding it — even when `A` is. The
contract "at most one in-flight `acquire` per name" is violated.

## Why it matters

`NamedSemaphores` is the only mechanism behind
`HostSharingRequirements::OnePerToken(name, _)` and
`HostSharingRequirements::OnePerTokens(names, _)`. These variants are how
Buck2 actions express constraints such as "this test must not run
concurrently with another holder of the same exclusivity token", typically
because the action depends on a non-shareable resource (a fixed TCP port, a
hardware device, a shared external fixture).

When the race fires, two such actions can run at the same time. The visible
failure surfaces upstream — a flaky test, an "address already in use" error,
a corrupted fixture — far from the actual cause, which makes it hard to
diagnose in the field.

The race is most likely on a freshly constructed `NamedSemaphores` under
bursty contention. That is exactly the situation at the start of a large
`buck2 test` run on a busy CI host.

## Reproduction

The current tests in this module are all single-threaded and cannot observe
the race. The following test races N threads through a `Barrier` on the same
name and asserts that all returned clones share one underlying semaphore.
It fails on the current implementation and passes deterministically on the
proposed fix.

```rust
#[test]
fn test_get_is_atomic_under_contention() {
    use std::sync::Arc;
    use std::sync::Barrier;

    const N: usize = 32;
    for _ in 0..32 {
        let named = Arc::new(NamedSemaphores::new());
        let barrier = Arc::new(Barrier::new(N));

        let handles: Vec<_> = (0..N)
            .map(|_| {
                let named = Arc::clone(&named);
                let barrier = Arc::clone(&barrier);
                std::thread::spawn(move || {
                    barrier.wait();
                    named.get("contended")
                })
            })
            .collect();

        let sems: Vec<_> = handles.into_iter().map(|h| h.join().unwrap()).collect();

        let _held = sems[0]
            .try_acquire(1)
            .expect("first acquire must succeed");
        for s in &sems[1..] {
            assert!(
                s.try_acquire(1).is_none(),
                "all clones must share the same underlying semaphore",
            );
        }
    }
}
```

Notes:

- The read-only fast path is preserved, so the common case (entry already
  present) still avoids the write-shard lock.
- No public API change. Behavior under contention becomes correct rather
  than racy.

## Severity

Correctness, latent. The race is timing-dependent and silent, so it is
unlikely to have caused an obvious failure during development, but it is
reachable from production code paths on every multi-threaded Buck2 invocation
that uses `OnePerToken` or `OnePerTokens`.

## Related (separate issue)

`acquire_from_permits_and_identifiers` in `host_sharing/src/host_sharing.rs`
will self-deadlock if a caller passes an `OnePerTokens` containing duplicate
names, because `SortedVec` does not deduplicate. This is a related but
distinct bug and should be tracked separately.

dkranchii:
## Summary

Fixes a TOCTOU race in `host_sharing::NamedSemaphores::get` that breaks the `OnePerToken` / `OnePerTokens` exclusivity guarantee used by `HostSharingBroker`.

The previous implementation did a non-atomic `get` followed by `insert` on the underlying `DashMap`. Two concurrent callers for the same name could each observe `None`, each construct a fresh `SharedSemaphore`, and each `insert` it — the second `insert` silently overwrites the first. The two callers then hold different semaphores under the same name, and the "only one in-flight `acquire` per name" contract is violated. Visible failures surface upstream as flaky tests, "address already in use", or corrupted shared fixtures.

Closes #1302

## The race

Two threads `A` and `B` call `NamedSemaphores::get("foo")` with no entry yet:

1. `A` looks up `"foo"` → `None`.
2. `B` looks up `"foo"` → `None`.
3. `A` constructs `Sem_A` and inserts `("foo", Sem_A)`. Returns `Sem_A`.
4. `B` constructs `Sem_B` and inserts `("foo", Sem_B)` (overwriting). Returns `Sem_B`.

`A` holds `Sem_A`; the map holds `Sem_B`. Both can `acquire(1)` at the same time — they are independent objects, each with one permit. Any later caller looks up `"foo"`, gets `Sem_B`, and observes its permit as available even while `A` is "holding the lock".

## Fix

Use `DashMap::entry(...).or_insert_with(...)` so the create-if-missing step is atomic per shard. The closure runs at most once per name across all threads, so all callers observe the same underlying `SharedSemaphore`.

The read-only fast path (`if let Some(...) = self.buckets.get(...)`) is preserved so the common case — entry already present — still avoids the write-shard lock. No public API change.

## Test plan

- [ ] `cargo test -p host_sharing` passes locally, including the new `test_get_is_atomic_under_contention`.
- [ ] (Optional verification) Temporarily reverting just the body of `get` to the previous `get`-then-`insert` form makes `test_get_is_atomic_under_contention` fail within a few iterations, confirming the test catches the original bug.

The new regression test races 32 threads through a `Barrier` on the same name and asserts that all returned clones share one underlying semaphore. It is deterministic on the fixed code and probabilistic on the racy code; the outer 32× repetition makes it reliable on a typical multi-core dev machine while keeping wall-clock cost in the tens of milliseconds.

## Related (separate, not addressed here)

`acquire_from_permits_and_identifiers` in `host_sharing/src/host_sharing.rs` will self-deadlock if a caller passes an `OnePerTokens` with duplicate names, because `SortedVec` does not deduplicate. This is a related but distinct bug; happy to file a follow-up issue and PR if useful.
host_sharing: fix TOCTOU race in NamedSemaphores::get


cormacrelf:
Looks right to me.

dkranchii:
@scottcao updated the pr based on your comment "Can we get comments to be more concise?". thanks!!

dkranchii:
@scottcao  can you please deploy the changes. thanks.

PR #1304, opened 2026-04-27
Add incremental linking support

reginaldl:
- This adds support for incremental linking to the prelude cxx rules. Incremental linking only works for MSVC. It's enabled via the `link_preference` attribute to `cxx_binary`. For compilers other than MSVC, this attribute is ignored. When `link_preference=="incremental"`, the action sets `local_only=True` as well as `allow_cache_upload=False`, and `no_outputs_cleanup=True`.

- Also fix `prebuilt_cxx_library` to propagate `ResourceInfo` from deps. `cxx_library `propagates `ResourceInfo` from its dependencies, allowing resource providers to flow through the dependency graph. `prebuilt_cxx_library` does not, which means any `ResourceInfo` attached to its deps is silently dropped. Here, we add `ResourceInfo`` propagation to `prebuilt_cxx_library_impl` to match the behavior of `cxx_library`.


akrieger:
There should be two parts to this I think. The toolchain should have an attribute for incremental linking behavior, and binaries should have an attribute for requesting or overriding it. The toolchain level attribute can be an enum of disabled/supported/enabled, which allows for configuring defaults on the platform based on your use case. Eg. host platforms can always have it be 'disabled' for cacheability, and your toolchain implementation can configure the default behavior through your choice of means.

For targets, the attribute should serve to override the default behavior of the toolchain. If the toolchain has the default enabled, a target can opt out. If supported, a target can opt in. If the toolchain does not support incremental linking then the target level attribute is ignored.

zjturner:
Updated the incremental linking part.  LMK if this is what you had in mind.

prelude/cxx/link.bzl
```diff
@@ -173,6 +175,13 @@ def cxx_link_into(
         gc_sections_data = None
 
     shared_library_interface = ctx.actions.declare_output(output.short_path + ".tbd", has_content_based_path = False) if opts.produce_shared_library_interface else None
+
+    if is_incremental_link:
+        ilk_filename = paths.replace_extension(output.short_path, ".ilk")
```
akrieger, line 180: nb. for other readers of this PR: `/ILK:` flag can override this but we don't have to support it here because it's a toolchain implementation detail and not a target level detail like `exe_name` is.

akrieger (approved):
LGTM

PR #1307, opened 2026-05-01
fix: stop using broken buck-resources crate from crates.io

### LINKED ISSUE #1226: Packages `buck-resources` and `sorted_vector_map` have BUCK files inside them, thus breaking `reindeer` targets
Sample error:
```
BUILD FAILED
Error evaluating build file: `gh_facebook_buck2_shims_meta//third-party/rust:BUCK.reindeer`

Caused by:
    0: Traceback (most recent call last):
         * shim/third-party/rust/BUCK.reindeer:21497, in <module>
             cargo.rust_library(
         * prelude/rust/cargo_package.bzl:237, in _cargo_rust_library
             native.rust_library(name = name, **kwargs)
         * prelude/native.bzl:490, in _rust_library_macro_stub
             rust_library(**kwargs)
         * prelude/rust/rust_library.bzl:1136, in wrapper
             rust_library(**kwargs)
         * prelude/rust/rust_common.bzl:14, in rust_common_impl
             rust_rule(_workspaces = workspaces, **kwargs)
       
       error: Error coercing attribute `srcs` of `gh_facebook_buck2_shims_meta//third-party/rust:sorted_vector_map-0.2.0`
         --> prelude/rust/rust_common.bzl:14:9
          |
       14 |         rust_rule(_workspaces = workspaces, **kwargs)
          |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          |
    1: Error coercing attribute `srcs` of type `attrs.list(attrs.source(), default=[])`
    2: Error coercing ["vendor/sorted_vector_map-0.2.0/src/lib.rs", "vendor/sorted_vector_map-0.2.0/src/map.rs", "vendor/sorted_vector_map-0.2.0/src/set.rs"]
    3: Error coercing "vendor/sorted_vector_map-0.2.0/src/lib.rs"
    4: Coercing `vendor/sorted_vector_map-0.2.0/src/lib.rs` as a source
    5: Source file `vendor/sorted_vector_map-0.2.0/src/lib.rs` does not exist as a member of package `gh_facebook_buck2_shims_meta//third-party/rust`.
```

Prior to running this I ran `./bootstrap/reindeer --third-party-dir shim/third-party/rust vendor && ./bootstrap/reindeer --third-party-dir shim/third-party/rust buckify`. This can be fixed by deleting the BUCK files, but I think this probably should actually be done by *not* fetching these crates from the internet to begin with, or maybe the version of Reindeer is too old or something?

lf-:
Instead: use the one in-tree! The problem with the one on crates.io is that it got published with a BUCK file in the source tarball, so it breaks Buck when you `reindeer vendor` it as the sources are no longer visible. The crates.io publication should certainly be fixed, but there's no reason to use the crates.io version in-repo to begin with.

Fixes: https://github.com/facebook/buck2/issues/1226

### DIFF
--- a/shed/completion_verify/BUCK
+++ b/shed/completion_verify/BUCK
@@ -18,7 +18,7 @@ rust_binary(
         },
     }),
     deps = [
-        "fbsource//third-party/rust:buck-resources",
+        "fbcode//buck2/integrations/resources/rust:buck_resources",
         "fbsource//third-party/rust:clap",
         "fbsource//third-party/rust:ptyprocess",
         "fbsource//third-party/rust:tempfile",
--- a/shim/third-party/rust/Cargo.toml
+++ b/shim/third-party/rust/Cargo.toml
@@ -48,7 +48,6 @@ bit-vec = "0.9"
 bitflags = "2.11.1"
 blake3 = { version = "=1.8.2", features = ["mmap", "rayon", "traits-preview"] } # Check overlay before updating
 bstr = { version = "1.10.0", features = ["serde", "std", "unicode"] }
-buck-resources = "1"
 bumpalo = { version = "3.20.2", features = ["allocator_api", "collections"] }
 byteorder = "1.5"
 bytemuck = { version = "1.25", features = ["const_zeroed", "derive", "min_const_generics", "must_cast", "nightly_portable_simd", "nightly_stdsimd"] }


lf-:
cc @scottcao, this issue was in my way while working on Buck OSS.

dtolnay (commented):
With recent reindeer vendor improvements (https://github.com/facebookincubator/reindeer/commit/861a0cd46acd9e166b5cb82034397124bf52e441, https://github.com/facebookincubator/reindeer/commit/7bf06f162f0d32e98093264591eb00bc16d3bc48) as well as https://github.com/facebook/buck2/commit/79168dc138c4150b4ad749c5132a8e871f5608a7, the BUCK file in crates.io `buck-resources` should no longer be an issue. Please rebuild reindeer from the git repo, rerun reindeer vendor and reindeer buckify, and it should result in that BUCK file getting deleted.

The `bootstrap/reindeer` file in this repo should get updated in the next couple days.

lf-:
That does sound like a great improvement, though I'm still kind of curious why one would use the out of tree one from crates rather than reference the one in-tree. Is there a particular reason for using the one on crates?

dtolnay:
We can change shed/completion_verify/BUCK to use the in-tree buck-resources but that does not help your issue because we cannot easily drop the dependency from shim/third-party/rust/Cargo.toml. That directory is synced from the monorepo into several other git repos, some of which need to use buck-resources and shouldn't need to contain the source.

PR #1308, opened 2026-05-01
buckconfig/watchmanconfig: ignore jj directory & cargo

lf-:
Neither of the configurations were actually ignoring these things; it's necessary to duplicate exclusions in both to stop the internal (fallback) file watcher from emitting junk into DICE, even though the excludes in it don't affect the internal watcher itself (it still sets inotify on the directories unconditionally).


8Keep (approved):

scottcao (approved):

PR #1310, opened 2026-05-02
builtin file watcher: log when graph is dropped

lf-:
The file watcher can lose events due to event storms from the kernel or such; previously this was never actually reported. This reports it.

Test plan: yolo. I don't really have a way to repro an event storm.


cormacrelf:
I'm not sure but you can probably just suspend buckd via `kill -STOP` or `kill -TSTP`. Then you spam it by just touching a file `max_queued_events + 1` times. When you resume `kill -CONT`, it should trigger queue overflow.

/proc/sys/fs/inotify/max_queued_events is where new inotify instances get their queue size from during init.

Issue #1309, opened 2026-05-01
FS race conditions in content-based paths


lf-:
I'm aware that some race conditions were fixed in 1716d84cfc8c46664c4e8b15656136c276b6633b "Make fs_util::remove_all idempotent under concurrent removes".

I'm running with `buck2 2026-04-30-c6b541e34a9e9448d14a22777412bfba5b01191a` which should include that, but it's still broken with the same sort of errors:

```
Action failed: gh_facebook_buck2_shims_meta//third-party/rust:zmij-1-build-script-run (write __ld_shim.sh)
`write_file` setting executable `/Users/jade/co/buck2/buck-out/v2/art/gh_facebook_buck2_shims_meta/third-party/rust/__zmij-1-build-script-run__/6ef2005251eb6c51/__ld_shim.sh`: metadata(/Users/jade/co/buck2/buck-out/v2/art/gh_facebook_buck2_shims_meta/third-party/rust/__zmij-1-build-script-run__/6ef2005251eb6c51/__ld_shim.sh): No such file or directory (os error 2)
```

I have this repro'ing pretty consistently when running clean builds of buck2 itself on macOS 15.7.5 with SentinelOne making my filesystem performance extra bad; I suspect that this race can also be encouraged to reproduce with rr chaos mode or other injection of random filesystem delays.

I would note that this filesystem code absolutely should be redesigned (writing files and making them executable as two separate steps operating on paths [is an unnecessarily race-prone approach](https://blog.sebastianwick.net/posts/how-hard-is-it-to-open-a-file/)), but it's not the cause of the sadness observed, it merely increases the rate of the sadness by giving more opportunities for it to happen.

Seemingly the cause of the observed sadness is that content-addressed paths are getting written by multiple separate actions, and those multiple actions aren't considering the path already written as appropriate, so they are calling cleanup() on it.

https://github.com/facebook/buck2/blob/22cc3b777bff8679aa935706e70c09d4064d6751/app/buck2_execute_impl/src/materializers/immediate.rs#L63-L64

The bugfix mentioned above doesn't actually solve the problem: it makes it so that remove() can race with itself safely, but that's not what is occurring here: `set_executable()` is racing with `remove()`, and more broadly, *other actions* consuming the path are racing with it being removed. As far as I can tell, the problem is that it is being *removed at all*.

[...]

PR #1312, opened 2026-05-04
Document content-based paths

cormacrelf:
There was no documentation for this that the public had access to. So I wrote some.

I'm not 100% sure about some of the stuff at the bottom.

[Rendered](https://github.com/cormacrelf/buck2/blob/document-cbp/docs/rule_authors/content_based_paths.md)

### DIFF
--- a/docs/rule_authors/content_based_paths.md
+++ b/docs/rule_authors/content_based_paths.md
@@ -1,0 +1,263 @@
+---
+id: content_based_paths
+title: Content-Based Paths
+---
+
+Content-based paths are a mechanism that enables Buck2 to deduplicate work
+across different configurations of a target.
+
+:::info
+  Content-based paths are not yet the default for all actions, but the
+  prelude is moving all possible actions to use content-based paths
+  (2026-05).
+:::
+
+## Actions and configurations
+
+A target may appear in multiple
+[configurations](../concepts/configurations.md). The purpose of
+different configurations is to have slightly different actions run,
+e.g. `rustc -O2` in release mode. The focus of this page is actions
+which *do not* change like that across configurations, to some degree.
+For example:
+
+- [`http_archive`](../../prelude/decls/http_archive/) and other rules
+  that download or decompress static data are
+  usually completely independent of configuration.
+- Interpreted languages like TypeScript don't usually depend very much
+  on configurations. You invoke `tsc`/`esbuild` just the same whether
+  you are targeting Linux or Windows, ARM or x86_64, ASAN or no. These
+  build actions might depend on just one dimension of the configuration,
+  e.g. debug/release.
+
+Historically, Buck2 included the hash of the configuration in the
+output path of every output [artifact](../concepts/glossary.md#artifact), so
+that artifacts produced under different configurations land in different
+locations:
+
+```warning
+buck-out/v2/gen/<configuration-hash>/<cell>/__<target>__/<output>
+```
+
+The output paths of an action are part of the command line and therefore
+part of the hash of an action, which is the cache key. This means that
+even if a target's actions *do not* do anything special
+depending on the configuration, like `-O2`, Buck2 would
+still execute (and cache) those actions separately — once per
+configuration.
+
+## Content-based paths and action keys
+
+With content-based paths, we eliminate configuration hashes from the
+paths. Instead, an action sees input paths with a content-based hash of the
+file as part of the path, and output paths with a placeholder where the
+hash goes. After the action has been executed, the output files are
+moved to a content-based path, with the hash set to the first 8 bytes of
+the digest.
+
+Here are two content-based path renderings:
+
+1. `buck-out/v2/art/cell/output_artifacts/__target__/libfoo.o` (**placeholder**)
+2. `buck-out/v2/art/cell/97752af0dd8a8d17/__target__/libfoo.o` (**real content-based path**)
+
+In a given `actions.run()`, all inputs have a **real** content-based path, and
+all outputs have **placeholder** paths. You won't see the same artifact
+rendered both ways in a given action; an output artifact
+(`out.as_output()`) must use a placeholder because we cannot know the
+file's hash until we run the action.
+
+When all paths are content-based or placeholders, the same action will
+be deduplicated across configurations, because there is no reference to
+the configuration hash in the action key. Note that
+the cell name and package/target names are still in the path, so this does
+not deduplicate across targets, only across configurations of one
+target.
+
+Once an artifact is bound, we hash the file and use a content-based
+path when we materialize it or pass it to other actions. In some cases
+this may be optimised if we know the file contents or digest in advance,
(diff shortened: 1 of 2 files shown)


docs/rule_authors/content_based_paths.md
```diff
@@ -0,0 +1,250 @@
...
+With content-based paths, we eliminate configuration hashes from the
+paths. Instead, an action sees input paths with a content-based hash of the
+file as part of the path, and output paths with a placeholder where the 
+hash goes. After the action has been executed, the output files are 
+moved to a content-based path, with the hash set to the first 8 bytes of 
+the digest.
+
+Here are two content-based path renderings:
+
+1. `buck-out/v2/art/cell/output_artifacts/__target__/libfoo.a` (**placeholder**)
```
lf-: Woah that's a surprising choice in the design; it means that actions of the same name but different configurations have to be *serialized* on each other. This is probably an intentional tradeoff, but it's kind of surprising to me.

Nix does this placeholder thing by generating a *random* (or, in this case, perhaps configuration based) hash and replacing *that*.

docs/rule_authors/content_based_paths.md
```diff
@@ -0,0 +1,250 @@
...
+    Note right of B: Inputs resolved with REAL content hash<br/>(already known from upstream artifacts)<br/>Output resolved with placeholder<br/>"output_artifacts"
+    B->>B: Build Action<br/>output_file="buck-out/v2/art/cell/output_artifacts/__target__/libfoo.a"<br/>args use placeholder path too
+    B->>AC: Lookup(Action)
+    alt Cache MISS
+        AC-->>B: not found
+        B->>RE: Execute action
+        Note left of RE: Executor writes libfoo.a to path<br/>containing "output_artifacts"
+        RE-->>B: Action Result<br/>{name:"buck-out/v2/art/cell/output_artifacts/__target__/libfoo.a", digest: D}
+        B->>AC: Store(Action → Action Result)
+        Note right of AC: Value stored in cache still<br/>carries the PLACEHOLDER path —<br/>no real content hash yet baked in.
```
lf-: This is surprising to me. The value stored in cache carries the placeholder path. How does that work/what does that mean? I might have the wrong mental model of how CAS works, but I was thinking that:

1. the *contents* of `buck-out/v2/gen/...../` go into CAS artifact abcd
2. the action producing that artifact gets registered in action cache pointing to abcd

And that since the *contents* are what goes into CAS, the concept of "carrying the placeholder path" shouldn't really make any sense; the only things going to the cache server containing a placeholder would be actions [and probably not even those! they might receive a hash?]. Is that right? Or am I mistaken about what's uploaded to CAS and how?

Also this might mean that isolation dirs cause complete cache misses; is that true? If so that would be wild.

docs/rule_authors/content_based_paths.md
```diff
@@ -0,0 +1,250 @@
...
+2. `buck-out/v2/art/cell/97752af0dd8a8d17/__target__/libfoo.a` (**real content-based path**)
+
+In a given action, all inputs have a **real** content-based path, and
+all outputs have **placeholder** paths. You won't see the same artifact
+rendered both ways in a given action; an output artifact
+(`out.as_output()`) must use a placeholder because we cannot know the
+file's hash until we run the action.
+
+Once an artifact is bound, we may hash the file and use a content-based
+path when we materialize it or pass it to other actions.
```
lf-: An artifact may be bound without executing it; for `write()` you *do* know the final path (#1309) but not for `run()`. So this sentence is sort of unclear.

lf- (commented):
Generally LGTM, thank you for writing this, it's clarified the model a bit for me. I'm still foggy on how exactly this works with distributed build caching.

Maybe it needs a diagram showing the build graph and one step of resolving, along with exactly what's sent to CAS.

docs/rule_authors/content_based_paths.md
cormacrelf: They are! When you run them locally. At least I read a comment to that effect.

docs/rule_authors/content_based_paths.md
cormacrelf: You are not mistaken, I am referring to the paths in the action result in AC. I will clarify.

And yes I believe isolation dirs do bust the caches. RE does see the `v2`. It's pretty clear on https://buck2.build/docs/concepts/isolation_dir/ with all the warnings about disk and resource usage.

cormacrelf: The layout of buck-out in RE is identical to local. That's how we can have hybrid RE that races both.

docs/rule_authors/content_based_paths.md
cormacrelf: If you used a random hash you would still have to put it in the action key and that's obviously bad. I think Nix is different because it has control over its executor a bit more and can inject path logic. RE protocol cannot.

cormacrelf: Also worth noting that a different kind of local executor would not require serializing the executions. Separate workdirs where we lay buck-out out again avoids conflicting writes to output files. (Along with preventing undeclared dependencies from getting into a build, which would be the main purpose.)

docs/rule_authors/content_based_paths.md
cormacrelf: Clarified

docs/rule_authors/content_based_paths.md
cormacrelf: New diagram. Much improved I think, thanks for the feedback.

IanChilds:
Thanks for doing this, and sorry for the delay! I'll just make a few minor changes and get it approved and then I'll import it

scottcao (approved):

docs/rule_authors/content_based_paths.md
```diff
@@ -0,0 +1,333 @@
...
+title: Content-Based Paths
+---
+
+Content-based paths are a mechanism that enables Buck2 to deduplicate work
+across different configurations of a target.
+
+:::info
+  Content-based paths are not yet the default for all actions, but the
+  prelude is moving all possible actions to use content-based paths
+  (2026-05).
```
winterqt, line 12: Does it perhaps make sense to mention (maybe not here, but elsewhere on this page) scenarios where content-based paths don't work and/or are unideal? (Especially given the project-wide setting.)

cormacrelf: Yes. I will add. Basically you want to use them unless you have actions that won't work. I think I'll link https://github.com/facebook/buck2/pull/1331 because it's quite illustrative of the circumstances where things fall down:

- multiple outputs written in one action
- one of the outputs references other outputs by path, which during execution is a placeholder path

And it solves it so it's a good example of that too.

IanChilds: I'm going to land what I had in the version Scott Cao accepted on Friday. Feel free to open another pull request if you want to include any changes that you made since then.

PR #1318, opened 2026-05-12
fix: ExternalRunnerTestInfo.type matches constructor

lf-:
I found a really funny thing in the ExternalRunnerTestInfo docs: the `type` field is shown as `.test_type`, not `.type` (!).

This is because it *is* `.test_type` as a field and `type` in the constructor. This is super confusing, let's unify the field (that I doubt anyone accesses) with the constructor so the docs make more sense.

### DIFF
--- a/app/buck2_build_api/src/interpreter/rule_defs/provider/builtin/external_runner_test_info.rs
+++ b/app/buck2_build_api/src/interpreter/rule_defs/provider/builtin/external_runner_test_info.rs
@@ -72,7 +72,7 @@ use crate::interpreter::rule_defs::resolved_macro::ResolvedStringWithMacros;
 #[repr(C)]
 pub struct ExternalRunnerTestInfoGen<V: ValueLifetimeless> {
     /// A Starlark value representing the type of this test.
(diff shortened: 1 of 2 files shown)


cormacrelf:
I have definitely experienced this issue before. :+1:

JakobDegen:
> (that I doubt anyone accesses)

Heh, grep in fbsource gives me 23 results.

But also, `.type` is a starlark builtin for getting the name, we definitely don't want to overload that. Can we go the other way instead? Add `test_type` as a thing we accept in the constructor, I'll volunteer to (dispath an AI to) codemod our constructors, and then we go from there? (I've also got a diff going that makes all the existing parameters named-only, that should help with that)

lf-:
> But also, `.type` is a starlark builtin for getting the name, we definitely don't want to overload that.

news to me! that's unfortunate, and I agree this would be a flawed approach then. 

> Can we go the other way instead? Add `test_type` as a thing we accept in the constructor, I'll volunteer to (dispath an AI to) codemod our constructors, and then we go from there?

Sounds great to me!

PR #1326, opened 2026-05-22
starlark: Support unpacking `Set`'s

### LINKED ISSUE #144: How to unpack a `set`?
I am using Starlark (and `starlark-rust`) as a configuration language, and would love to be able to use sets.
Unfortunately, I cannot extract any values from them - `SetData` is [`pub (crate)`](https://github.com/facebook/starlark-rust/blob/522cd8ff4ef52e0f70e3b7ed2ab4a959f62d3331/starlark/src/values/types/set/value.rs#L64).

Is there any harm in exposing it? Would you accept a PR?

PhilipMetzger:
It didn't support it yet and it is useful for embedder's. I mostly followed stepancheg's advice in the bug.

This fixes https://github.com/facebook/starlark-rust/issues/144

cc @ndmitchell for visibility.


PhilipMetzger:
ping on this so it hopefully lands.

PR #1328, opened 2026-05-23
Support incoming transitions for C++ rules

### LINKED ISSUE #1186: Is there a plan to add support for incoming transitions to more prelude rules?
Basically what the title says. In particular, rules like `cxx_binary` would be particularly useful.

I'd be happy to raise a PR, but because this requires removing support for old school transitions (`cfg` and incoming transitions being mutually exclusive), I'm not sure that it would be accepted.

cbarrete:
Closes #1186.

### DIFF
--- a/prelude/decls/cxx_rules.bzl
+++ b/prelude/decls/cxx_rules.bzl
@@ -180,7 +180,8 @@ cxx_binary = prelude_rule(
         | buck.contacts_arg()
         | _cxx_binary_and_test_attrs()
     ),
-    cfg = constraint_overrides.transition,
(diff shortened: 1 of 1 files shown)


scottcao:
Unfortunately, there are some internal usages that block the use of new incoming transition on cpp rules at the moment. However, I think this is a good change overall. Can you make this change OSS-only?

scottcao (changes requested):

cbarrete:
Thank you for taking a look.

I'm not familiar with the tooling you have to split internal/OSS code, so let me know if this is not done the right way.

scottcao (approved):

PR #1329, opened 2026-05-27
Option to watch fs events for subdirs separately from root dir

bvinc:
This creates a new project setting "watch-included-root-dirs-only" which watches subdirectories at the root level separately. This allows real ignoring of the buck-out folder, which makes the possibility of missing fs change events much less likely.

This is the most important on mac, where fsevent has no knobs to turn, and no way to really ignore the buck-out folder.  Without this change, large builds will easily and consistently overflow the internal fsevent buffers and cause file system events to be missed.  After this change, almost no events happen, as changes outside of the buck-out folder are rare.

Unfortunately, this requires manual non-recursive polling of the root directory occasionally.  I chose to put this in calls to sync2, which usually happen once at the beginning of a build.

### DIFF
--- a/app/buck2_file_watcher/src/file_watcher.rs
+++ b/app/buck2_file_watcher/src/file_watcher.rs
@@ -129,10 +129,26 @@ impl dyn FileWatcher {
                 WatchmanFileWatcher::new(project_root.root(), root_config, cells, ignore_specs)
                     .buck_error_context("Creating watchman file watcher")?,
             )),
-            "notify" => Ok(Arc::new(
-                NotifyFileWatcher::new(project_root, cells, ignore_specs)
+            "notify" => {
+                let watch_included_root_dirs_only = root_config
+                    .parse::<bool>(BuckconfigKeyRef {
+                        section: "project",
+                        property: "watch-included-root-dirs-only",
+                    })
+                    .buck_error_context(
+                        "Failed to parse project.watch-included-root-dirs-only config",
+                    )?
+                    .unwrap_or(false);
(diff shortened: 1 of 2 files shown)


app/buck2_file_watcher/src/file_watcher.rs
```diff
@@ -129,10 +129,26 @@ impl dyn FileWatcher {
                 WatchmanFileWatcher::new(project_root.root(), root_config, cells, ignore_specs)
                     .buck_error_context("Creating watchman file watcher")?,
             )),
-            "notify" => Ok(Arc::new(
-                NotifyFileWatcher::new(project_root, cells, ignore_specs)
+            "notify" => {
+                let watch_included_root_dirs_only = root_config
+                    .parse::<bool>(BuckconfigKeyRef {
+                        section: "project",
+                        property: "watch-included-root-dirs-only",
```
cormacrelf, line 136: Snake case for buckconfig keys isn't it? Also maybe bike shed it a bit more

cormacrelf: I really wish you could grep BuckconfigKeyRef::new and get a big list of all of them to see where you might put a new one and what you might name it. But it ain't so.

daandemeyer:
@bvinc I've pushed an alternative version of this PR as the second commit to https://github.com/facebook/buck2/pull/1351 which makes use of https://github.com/notify-rs/notify/pull/950 to drasticaly simplify the changes on buck's end. Would be great if you could take a look and maybe try it out.

PR #1331, opened 2026-05-29
fix(rust): resolve build-script `OUT_DIR` paths to content-based path

JonasKruckenberg:
Rust build scripts that emit a path inside OUT_DIR via `cargo:rustc-env` or `cargo:rustc-link-search` (eg cranelift) intermittently failed to compile because the content-addressed OUT_DIR will be materialized at `<content-hash>/OUT_DIR` while the `cargo:rustc-env`  value will continue to use the placeholder `output_artifacts/OUT_DIR` path.

This change adds logic to `buildscript_run.py` and `rustc_action.py` to rewrite paths that live inside the `OUT_DIR` to be prefixed with a sentinel `${__BUILDSCRIPT_OUT_DIR__}` that we then resolve later when the content hash is known.

### DIFF
--- a/prelude/rust/tools/buildscript_run.py
+++ b/prelude/rust/tools/buildscript_run.py
@@ -16,11 +16,14 @@ import re
 import subprocess
 import sys
 from pathlib import Path
-from typing import Any, IO, NamedTuple, Optional
-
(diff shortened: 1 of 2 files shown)


dtolnay (approved):
Thanks!

PR #1334, opened 2026-06-03
Skip -ffile-prefix-map for assembler targets

reginaldl:
Skip `-ffile-prefix-map` flags for assembler targets (NASM, MASM, gas) which don't understand this compiler option

### DIFF
--- a/prelude/cxx/compile.bzl
+++ b/prelude/cxx/compile.bzl
@@ -1963,8 +1963,9 @@ def _mk_argsfiles(
             # The precompile_args field overrides these.
             return
 
-        if _get_category(ext) == "asm_compile":
-            # Assemblers (NASM, MASM, gas) don't understand -ffile-prefix-map.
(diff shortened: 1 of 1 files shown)


dtolnay:
The `_get_category(ext) != "asm_compile"` on this line would now always be True:

https://github.com/facebook/buck2/blob/61da76c0d33484f39d98c801a94c65f9bcf41f67/prelude/cxx/compile.bzl#L1993

reginaldl:
> The `_get_category(ext) != "asm_compile"` on this line would now always be True:
> 
> https://github.com/facebook/buck2/blob/61da76c0d33484f39d98c801a94c65f9bcf41f67/prelude/cxx/compile.bzl#L1993

I updated to check the compiler type instead. Gcc/clang would compile .s/S and support `-ffile-map-prefix` so my initial condition was too broad.

dtolnay (approved):
Thanks!

PR #1339, opened 2026-06-10
buck2_external_cells: Don't inherit Git environment variables

9999years:
If you run `buck2` in a `pre-commit` hook, then the inherited `GIT_DIR` and similar can cause `buck2` to run destructive commands like `git reset --hard` in the wrong repository!


JakobDegen (approved):

PR #1346, opened 2026-06-26
buck2_server: Implement Git revision data

9999years:
What it says on the tin, closes a long-standing TODO and helps with our telemetry.

TODO: Should we just reuse the `hg_revision` field for Git revisions? They _are_ also 40-character hashes...


scottcao (approved):

PR #1351, opened 2026-07-03
file_watcher: filter buck-out events before relativizing in the notify watcher

daandemeyer:
Two changes to the notify file watcher.

The first fixes a bug where the watcher relativized every event path before doing anything else. Relativizing rejects components buck's path types can't represent, like a literal backslash (systemd's escaped unit names are one source of these in build outputs) or non-UTF-8 bytes, and the error poisons the watcher: every command after that fails with "Error relativizing: ... is not relative to project root" until the daemon is killed. It now checks the buck-out prefix on the raw path first, relativizes leniently for the ignore check, and for anything still unrepresentable invalidates the nearest representable parent directory instead of erroring, like the watchman and edenfs watchers already do.

The second stops the watcher watching buck-out and ignored directories at all. It watched the whole project root recursively and filtered those events out only after delivery. This switches to a notify fork with a watch_filtered() API that skips them at registration time, so on Linux they never get inotify watches and produce no events. That drops the main source of wakeups during builds, avoids running into fs.inotify.max_user_watches on large buck-out trees, and speeds up daemon startup. FSEvents and Windows can't watch selectively, so there the events are dropped inside notify before they reach us.

The second commit needs watch_filtered(), which isn't in a released notify yet, so it's pulled in with a patch.crates-io git dependency for now. The first commit doesn't depend on the fork.


app/buck2_file_watcher/src/notify.rs
```diff
@@ -85,20 +85,26 @@ impl NotifyFileData {
...
             // We ignore the buck-out prefix, as those are uninteresting events caused by us.
             // We also ignore other buck-out directories, as if you have two isolation dirs running at once, they are not interesting.
             // We do this in the notify-watcher, rather than a generic layer, as watchman users should configure
             // to ignore buck-out, to reduce the number of events, rather than hiding them later.
-            if path.starts_with(InvocationPaths::buck_out_dir_prefix()) {
-                // We don't want to event add them as ignored events, since they are super common
-                // and very boring
-                continue;
+            //
+            // Checked on the raw path *before* relativizing: build outputs can transiently
```
keszybz: "Checked on" → "Check" ?

app/buck2_file_watcher/src/notify.rs
```diff
@@ -85,20 +85,26 @@ impl NotifyFileData {
...
+            // contain names buck's path types reject (e.g. a literal backslash), and
+            // erroring on them would poison the watcher over events we discard anyway.
+            if let Ok(rel) = path.strip_prefix(root.root().as_path()) {
+                if rel.starts_with(InvocationPaths::buck_out_dir_prefix().as_str()) {
+                    // We don't want to event add them as ignored events, since they are super common
+                    // and very boring
```
keszybz: This sentence is strange…

daandemeyer:
I'm aware that this doesn't mean I can store paths with backslashes in buck-out, but it does give me a chance to fix them up before buck notices them.

PR #1353, opened 2026-07-03
Stop doing a tset projection walk in write_json validation

csamak:
This seems unnecessary given validation in TransitiveSet::new.

In one of our use cases, this was adding a 10s cost to dynamic analysis.


scottcao (approved):

PR #1354, opened 2026-07-03
github: fall back to standard runners on forks

daandemeyer:
The linux-build-and-test, windows-build-and-test and
windows-build-examples jobs run on Meta's larger runners (4-core-ubuntu
and windows-8-core, added in #545 and #812 to speed up CI). Runner
labels are repository (owner) specific, so on forks these jobs sit
queued forever waiting for a runner that does not exist.

Select the runner label based on the repository owner so that forks
fall back to the standard GitHub-hosted runners and CI still runs
there, just slower. The larger runners currently use the same images as
ubuntu-latest and windows-latest, so forks build with the same OS and
toolchains as upstream.


dtolnay (approved):
Thanks!

PR #1356, opened 2026-07-03
github: add a workflow to manually create a timestamped release

daandemeyer:
Add a workflow_dispatch workflow that publishes a release tagged with the
current UTC date and time (to the second), including dotslash files. The
build steps shared with the existing push-triggered workflow are extracted
into a reusable workflow.

This is especially useful on forks to manually create releases from
unmerged changes, without having to wait for changes to land upstream.

### DIFF
--- a/.github/dotslash-config.json
+++ b/.github/dotslash-config.json
@@ -64,6 +64,38 @@
             "path": "rust-project-x86_64-unknown-linux-musl"
           }
         }
+      },
+      "starlark_fmt": {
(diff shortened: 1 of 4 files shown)


thoughtpolice:
Yeah, I have a similar workflow here: https://github.com/thoughtpolice/buck2/blob/main/.github/workflows/binaries.yml this would be really nice to merge!

8Keep (approved):

PR #1357, opened 2026-07-04
superconsole: styling tweaks and (hopefully?) improvements

davidbarsky:
Hi folks! I recently starting using Buck2 again (I didn't realize how much I missed it until I developed a moderately complicated build!). Anyways, I had some spare cycles and polished the superconsole TUI to be more consistent and bring out the latent tabular design. Screenshot of this in Ghostty below:

Screenshot 2026-07-03 at 6 29 44 PM

Here are the concrete changes:
1. The horizontal timed list separator is now drawn as a continuous line, using a box-drawing character (`─`, U+2500) instead of the hyphen.
2. Slow actions are colored by by active-execution stage, not total elapsed time. The red/yellow escalation is now keyed off on time executing within a stage, rather than the entire action. I'm not currently using remote caching _or_ execution, so a lot of actions would become red _way_ too often. As without remote execution/caching, most tasks are simply waiting in a queue, the (valuable!) yellow/red tinting lost all meaning due to its frequency and overuse.
3. Terminal resize events are now handled: subscribers get a `ConsoleInteraction` (consisting of toggle, key, resize), which is now used to immediately redraw the superconsole.
4. Restyled progress header. Few things:
    1. the phase lines now are now a table with a bold header row. Per-phase columns are moved into aligned columns.
    2. the network and session info are anchored to the top and remain there the entire build.
5. Superconsole spans are deduped and modeled using a `EventDisplay { label, detail,
category }` instead of a pre-flatted string like 1`"{label} -- {detail}"`. The rough before/after is below. Here are some notable changes:
    1. the target label is user's main terminal text color, the category/identifier are in surrounded in dimmed brackets, derived from the main color. The elapsed time and non-action are also dimmed, they are separated using a dimmed `·` separator.
    2. stage names are padded to a fixed width so their times align on a column basis.
    3. Truncation is per-cell using ellipses. The time column will *never* go away, but as the terminal window gets smaller, certain bits of data will be ellipsed. First, the subaction column, then the detail, then the target.
   4. Age-based warning colors only apply to the action, not the full row.
6. The action category is deduplicated against the target label. Actions that repeat the target's label now just show the category, identifiers that differ stay. This means an action that'd previously render like `pkg:foo [copy foo]` would now render as `pkg:foo [copy]`.

Before:
```
pkg:target -- action (category identifier) [prepare 5.0s]   10.0s
```

After:
```
pkg:target [category identifier] prepare         5.0s       10.0s
```

I don't recall how importing a PR into a phrabricator diff works, but these commits all should build independently.

### DIFF
--- a/app/buck2_client_ctx/src/console_interaction_stream.rs
+++ b/app/buck2_client_ctx/src/console_interaction_stream.rs
@@ -240,7 +240,7 @@ impl SuperConsoleToggle {
             SuperConsoleToggle::Dice => "DICE",
             SuperConsoleToggle::DebugEvents => "debug events",
             SuperConsoleToggle::TwoLinesMode => "two lines mode",
-            SuperConsoleToggle::DetailedRE => "detailed RE",
+            SuperConsoleToggle::DetailedRE => "network details",
             SuperConsoleToggle::Io => "I/O counters",
             SuperConsoleToggle::TargetConfigurations => "target configurations",
             SuperConsoleToggle::ExpandedProgress => "expanded progress",
--- a/app/buck2_client_ctx/src/console_interaction_stream.rs
+++ b/app/buck2_client_ctx/src/console_interaction_stream.rs
@@ -8,13 +8,19 @@
  * above-listed licenses.
  */
 
+use crossterm::event::Event;
+use crossterm::event::EventStream;
+use crossterm::event::KeyCode;
+use crossterm::event::KeyEventKind;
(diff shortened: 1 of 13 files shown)


davidbarsky:
Rebased this against `main`, so hopefully CI will pass again!

davidbarsky:
whoops, rustdoc failures. fixing.

scottcao (approved):

PR #1359, opened 2026-07-04
fix: Enable more crates to build on windows

gilescope:
Rust crates whose `build.rs` compiles C via `cc-rs` (wasmtime, secp256k1-sys, zstd-sys, …) fail on Windows with a misleading `cc-rs: Could not copy or create a hard-link to the generated lib file`. Two Windows-only prelude bugs cause it:

1. **`rust/tools/from_any_dir.py`** used `os.execl`, which on Windows is a non-blocking `_P_OVERLAY` spawn — the `$CC`/`$AR` shim returns 0 while `cl.bat`/`lib.bat` run detached, so cc-rs sees success with no output. Use `os.spawnv(os.P_WAIT)` on `nt`.

2. **`utils/cmd_script.bzl`** `.bat` wrappers invoke a repo-root-relative `run_msvc_tool.py` with no `cd`, so they break when `from_any_dir` chdir's to the buildscript cwd (`can't open file` → exit 2, which also makes cc-rs's family detection fall back to GNU). Resolve the payload via `%~dp0`, like `command_alias.bzl`. No `cd`, so caller-supplied `%*` still resolve against the caller's cwd.

Bug 1 masked Bug 2; both are needed. POSIX is unaffected (there `$CC` is a real compiler, not a `.bat` wrapper). Verified on `windows-latest`: a wasmtime build script that failed before now builds.

Testing proof: windows x86 build here: https://github.com/gilescope/buck2-fixups/pull/65

(AI assisted by Claude 4.8)

### DIFF
--- a/prelude/rust/tools/from_any_dir.py
+++ b/prelude/rust/tools/from_any_dir.py
@@ -51,6 +51,19 @@ def main():
     cc = list(filter(lambda arg: arg != "-Wl,-lomp", cc))
 
     os.chdir(args.cwd)
+    if os.name == "nt":
+        # os.exec* on Windows is a non-blocking _P_OVERLAY spawn: the parent
+        # process exits 0 immediately while the child (e.g. cl.bat / lib.bat)
+        # runs detached. A cc-rs buildscript then sees the compiler/archiver
+        # "succeed" with no object/archive produced yet, and fails at the
+        # hard-link of the never-created lib. Spawn and WAIT for the real exit
+        # code instead. (os.spawnv uses the same CRT launch that os.execl does,
+        # so it runs the .bat the same way -- just synchronously.)
+        try:
+            sys.exit(os.spawnv(os.P_WAIT, cc[0], cc))
+        except Exception:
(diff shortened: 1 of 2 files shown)


dtolnay (changes requested):
The from_any_dir.py fix looks good. Thanks!

The cmd_script.bzl fix currently breaks a large number of build scripts internally. For example https://github.com/rust-lang/rust/blob/1.96.1/library/compiler-builtins/compiler-builtins/build.rs fails like this:

```console
Action failed: fbsource//xplat/rust/toolchain/sysroot/1.96.1:compiler_builtins-0.1-build-script-run (buildscript)
The system cannot find the path specified.
Command failed with exit code 1
Command: ['..\\..\\..\\..\\..\\..\\..\\..\\60cf61f8efc6fe9a\\xplat\\rust\\toolchain\\sysroot\\1.96.1\\__compiler_builtins-0.1-build-script-run__\\__rustc_shim.bat', '--version']
```

I think you might only want to do this for some subset of the callers of `cmd_script`.

PR #1360, opened 2026-07-04
bxl: expose allbuildfiles() and rbuildfiles()

daandemeyer:
allbuildfiles() and rbuildfiles() are supported by the uquery and
cquery commands but were not exposed to BXL. Add them to ctx.uquery(),
ctx.cquery() and ctx.lazy.uquery(), delegating to the same
implementations used by the query commands. The lazy cquery context
only supports eval(), so it is not modified.


JakobDegen (approved):

8Keep (approved):

PR #1362, opened 2026-07-05
starlark: typecheck against the implicit globals from the prelude

daandemeyer:
The typecheck oracle only contained the native globals, so any name the
interpreter injects into the module scope was reported as not found:
the prelude's public symbols, the members of the prelude's `native`
struct in BUCK files, and the root implicit import. This made it
impossible to typecheck BUCK files or .bzl files that use prelude
symbols without loading them.

Extend the oracle globals with those symbols, taking the values from
the evaluated prelude module. The typechecker derives types from the
actual frozen values, so rule calls are checked against their
attribute-based signatures, including select() unions and
required/optional parameters.

That alone is not enough for BUCK files: the typechecker only ran the
binding solver on top-level `def` statements, and the lint fill pass
skips calls, so top-level rule calls were still unchecked. Run the
solver on top-level expression, `if` and `for` statements as well. The
fill pass does not check those, so nothing is reported twice;
assignments and loads stay with the fill pass.

Calls in top-level assignments are still unchecked, since the fill
pass evaluates those and returns Any for calls.

### DIFF
--- a/app/buck2_cmd_starlark_server/src/typecheck.rs
+++ b/app/buck2_cmd_starlark_server/src/typecheck.rs
@@ -67,9 +67,10 @@ impl Cache<'_> {
         match self.oracle.get(&(cell, path_type)) {
             Some(g) => Ok(g.dupe()),
             None => {
-                let globals = Environment::new(cell, path_type, &mut self.dice.clone())
-                    .await?
-                    .globals;
+                let env = Environment::new(cell, path_type, &mut self.dice.clone()).await?;
+                let globals = env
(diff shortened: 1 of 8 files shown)


cormacrelf:
I'll have a look the globals injection, it feels like I've written this code before in one of my many typechecking PRs but it's been a while

For the top level type checking we already have this which is more complete https://github.com/facebook/buck2/pull/1289

daandemeyer:
Ah oops I missed your pr, my bad, Feel free to merge the missing bits of this pr into yours and then I'll be happy to close this one in favor of yours.

PR #1364, opened 2026-07-07
fs: allow backslashes in paths via buckconfig

daandemeyer:
File names and relative paths reject backslashes on every platform. On
Windows a backslash is a path separator, but on other platforms it is a
valid path character that occurs in practice. Building OS images, for
example, can produce systemd-escaped names such as
`system-systemd\x2dfoo.slice`, which Buck2 currently rejects.

Keep rejecting backslashes by default, but add
`[buck2] allow_backslashes_in_paths = true` to permit them on platforms
where they are not path separators. Treat the setting as daemon startup
configuration so changing it restarts buckd and one daemon cannot mix
path invariants. Windows continues to reject literal backslashes.

Initialize the low-level path policy from the startup configuration in
both the client and daemon. Same-value initialization is idempotent for
`--no-buckd`, while conflicting values are rejected. This also preserves
the original concrete error type of `from_system_path`, since validation
no longer performs a fallible environment lookup.

Cover policy initialization, platform behavior, and all affected path
validators in buck2_fs tests. Cover buckconfig parsing in buck2_common,
and document the opt-in in the buckconfig reference.

Signed-off-by: Daan De Meyer <daan@amutable.com>

### DIFF
--- a/app/buck2_client_ctx/src/immediate_config.rs
+++ b/app/buck2_client_ctx/src/immediate_config.rs
@@ -32,6 +32,7 @@ use buck2_fs::fs_util;
 use buck2_fs::paths::abs_norm_path::AbsNormPath;
 use buck2_fs::paths::abs_norm_path::AbsNormPathBuf;
 use buck2_fs::paths::abs_path::AbsPath;
+use buck2_fs::paths::init_allow_backslashes_in_paths;
 use buck2_fs::working_dir::AbsWorkingDir;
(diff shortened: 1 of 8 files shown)


lf-:
I feel vaguely like this should be a buckconfig setting instead; the ergonomics of passing in an env-var are not ideal.

daandemeyer:
@lf- Reworked to a buck config option

PR #1367, opened 2026-07-08
prelude: let `default_allow_cache_upload` govern `rust_library` uploads

davidbarsky:
I hit this in a setup where I'm using buck2 *without* [RE but with remote action/artifact caching](https://github.com/buchgr/bazel-remote). Given that `rust_binary` was typically uploaded, but `rust_libraries` were not, this smells like a bug.

If I were using RE, a `rust_library` would be automatically uploaded, but... I'm not. This should be a no-op if used with RE.


dtolnay (approved):
Thanks!

PR #1370, opened 2026-07-13
buck2_events: Add OpenTelemetry `EventSink`

9999years:
This adds a new `OtelEventSink` which sends `InvocationRecord` wide events to an OpenTelemetry collector, if configured.

This respects the following environment variables:
- `$OTEL_EXPORTER_OTLP_*`
- `$OTEL_SDK_DISABLED`
- `$OTEL_SERVICE_NAME`
- `$OTEL_RESOURCE_ATTRIBUTES`
- `$TRACEPARENT`
- `$TRACESTATE`

We've been running this patch internally and it works well.

Most of the diff is in `app/buck2_events/src/sink/otel_record.rs` which is an essentially mechanical translation from `InvocationRecord`s to a set of Otel attributes.


9999years:
Rebased.

PR #1371, opened 2026-07-13
re_grpc: request zstd-compressed `BatchReadBlobs` responses

9999years:
There are many small `BatchReadBlobs` requests, and they're large enough that compression would help. When the server advertises zstd support in its capabilities, advertise zstd as an acceptable response compressor and decompress any zstd-compressed blobs the server returns.

NB: It would be nice if we could only compress blobs above a certain size, but we don't know the size when we make the request. We continue to advertise IDENTITY an acceptable response compressor, so we may need to patch our cache server to choose an appropriate size threshold above which to compress data (Bazel chooses 100 bytes, AIUI).

In a moderately sized build with ~81,000 BatchReadBlobs requests, this patch shows significant benefits:
- Request latency dropped:
  - p100: 16,653ms to 3,468ms (4.8x speedup)
  - p90: 953ms to 593ms
  - p75: 529ms to 312ms
  - p50: 235ms to 161ms
  - p25: 110ms to 96ms
  - p0: 67ms to 68ms
- Cumulative blob size over the network dropped from 5,807MiB to 1,626MiB (3.57x compression ratio)
- Blob size distribution over the wire improved:
  - p100: 3.7M to 741.8K
  - p90: 188.4K to 58.0K
  - p75: 103.2K to 33.9K
  - p50: 12.9K to 3.7K
  - p25: 32B to 45B (regression)
  - p0: 15B to 2B (regression)

### DIFF
--- a/remote_execution/oss/re_grpc/src/client.rs
+++ b/remote_execution/oss/re_grpc/src/client.rs
@@ -161,7 +161,8 @@ pub struct RECapabilities {
     /// Largest size of a message before being uploaded using bytestream service.
     /// 0 indicates no limit beyond constraint of underlying transport (which is unknown).
     max_total_batch_size: usize,
-    /// Compressors supported by the "compressed-blobs" bytestream resources.
+    /// Compressors supported by the "compressed-blobs" bytestream resources. Also used to decide
+    /// whether we can request zstd-compressed `BatchReadBlobs` responses.
     supported_compressors: Vec<Compressor>,
 }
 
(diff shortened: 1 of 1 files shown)


remote_execution/oss/re_grpc/src/client.rs
```diff
@@ -175,6 +176,8 @@ pub struct RERuntimeOpts {
     cas_ttl_secs: i64,
     /// Maximum number of digests per `FindMissingBlobs` RPC.
     find_missing_blobs_batch_size: usize,
+    /// Whether to request zstd-compressed `BatchReadBlobs` responses (server supports zstd).
+    batch_read_zstd: bool,
```
9999years, line 180: We could have a set of supported compressors here instead, but we'd have to do a slightly more invasive refactoring. Let me know if you'd like me to do that.

9999years:
Rebased.

PR #1372, opened 2026-07-14
rust-project: add; use target kinds to rust-analyzer's `provider.bzl`

davidbarsky:
ERSC’s `rust_test` is a decorator around prelude's `rust_test`; delegating to prelude.rust_test, but replacing `RunInfo` and `ExternalRunnerTestInfo` so that we can run tests-per-process (we wrote our own, dumb, external test runner).

Today, rust-project does a hard match on the rule name, but this set of changes wires the lib/bin/test through the RustAnalyzerInfo provider, exposing it to rust-project. This allows rust-project to actually _succeed_ within ERSC's repo. It is, in my view, composes more elegantly, and I wish I had noticed this whoopsie when I was still at Meta, working on rust-project.


capickett:
Pulling it in to address lints on phabricator side.

One thing to add imo: in resolve deps we explicitly check for rust_library|binary|test rules. We can add an extra query for targets with this new attribute. In your case your rule is exactly named `rust_test`, but with this new attribute, you don't have to keep the name exactly the same.

Wilfred:
Thanks for the PR @davidbarsky :)

FWIW we ended up adding 5938780820603f4ade912250932333116e9a72b0 to ensure this was forwards/backwards compatible, you might want that commit too depending on how you're deploying rust-project.

PR #1373, opened 2026-07-15
Fix clang-specific cc/cxx invocations in cargo_buildscript

hdgarrood:
After updating to a version of buck2 that included these changes (using bundled prelude), we started seeing the following when building the `openssl-sys` crate via a Reindeer-generated BUCK file:

```
Header expansion error:
Error { kind: ToolExecError, message: "command did not execute successfully (status code exit status: 1): LC_ALL="C" "…/__openssl-sys-0.9-build-script-main-run__/74052bece190009d/__cc_shim.sh" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "-I" "../../../../../../toolchains/b42aeba648b8c415/__openssl_includes__/out.link-dev/include" "-Wall" "-Wextra" "-E" "build/expando.c"" }
Failed to find OpenSSL development headers.
```

The real error (which, unfortunately, is not shown) is that `--ld-path` is a clang-specific flag - or at least, that it is not recognized by gcc. It seems a little unlikely to me that buildscripts need to access the linker in general so I'm not sure we necessarily need to be passing `--ld-path` at all? But this change should at least expand the set of cargo build scripts that can successfully run with gcc toolchains, without affecting clang.

I've tested this in the context of Mercury's monorepo and I'm confident it works there, but I'm struggling a bit to set up a reproducer/test outside of the context of that monorepo. Can you help me figure out whether this should be accompanied by a test, and if so what kind of test?

### DIFF
--- a/prelude/rust/cargo_buildscript.bzl
+++ b/prelude/rust/cargo_buildscript.bzl
@@ -304,6 +304,8 @@ def _cargo_buildscript_impl(ctx: AnalysisContext) -> list[Provider]:
     )
     deps_link = deps_tset.project_as_args("default")
     sanitizer_flags = ["-fno-sanitize=all"]
+    cc_is_clang = cxx_toolchain_info.c_compiler_info.compiler_type.startswith("clang")
+    cxx_is_clang = cxx_toolchain_info.cxx_compiler_info.compiler_type.startswith("clang")
     env["LD"] = _make_cc_shim(
         ctx = ctx,
         name = "__ld_shim",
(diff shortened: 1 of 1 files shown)


lf- (commented):
FYI the reason a build script would be calling a c compiler is because it's compiling C, which they do sometimes, with the `cxx` crate.

dtolnay:
@lf- I think the uncertainty was about why it would run a linker, not why it would run a C compiler. Buildscripts that compile C would almost never link because Cargo is what runs the linker. The build script just informs Cargo of the compiled object location using `rustc-link-search` https://github.com/rust-lang/cc-rs/blob/cc-v1.2.67/src/lib.rs#L1630-L1644.

Build scripts that run a linker are usually doing some `configure`-like thing, for example to extract a library version number from a `#define` in a C header by compiling, linking, running, and parsing the output of some C code. One example is the `gmp-mpfr-sys` crate.

dtolnay (approved):
Thanks!

PR #1379, opened 2026-07-21
fix failing analysis in docs/buck2_lab

### LINKED ISSUE #1147: Multiple targets failing on analysis
Hi,

I just ran `buck2 targets -A --json //...` on the repo and it failed right away on multiple errors, here's the first one:

```
Error parsing gh_facebook_buck2//docs/buck2_lab/greeter_lib
Error parsing: `gh_facebook_buck2//docs/buck2_lab/greeter_lib:BUCK`

Caused by:
    error: Parse error: unexpected new indentation block here, expected one of "\n", "(", "+", "-", "...", "DEDENT", "FLOAT", "FSTRING", "IDENTIFIER", "INTEGER", "STRING", "[", "break", "continue", "def", "for", "if", "lambda", "load", "not", "pass", "return", "{" or "~"
      --> docs/buck2_lab/greeter_lib/BUCK:16:1
       |
    16 |     rust_test( # @oss-enable
       | ^^^^
       |
```

I think it might be worth adding a github action to run builds or at least analysis on PRs.

lf-:
This is most likely caused by Meta's code formatter.

Fixes: https://github.com/facebook/buck2/issues/1147

### DIFF
--- a/docs/buck2_lab/greeter_lib/BUCK
+++ b/docs/buck2_lab/greeter_lib/BUCK
@@ -1,8 +1,7 @@
-# @nolint 
+# @nolint
 
 # @oss-disable[end= ]: load("@fbsource//tools/build_defs:rust_library.bzl", "rust_library")
 # @oss-disable[end= ]: load("@fbsource//tools/build_defs:rust_unittest.bzl", "rust_unittest")
-
 
 oncall("build_infra")
 
@@ -15,8 +14,10 @@ rust_library(
     ],
 )
 
-# @oss-disable: rust_unittest(
-    rust_test( # @oss-enable
+# @oss-disable: _target = rust_unittest
+_target = rust_test  # @oss-enable
+
+_target(
     name = "test",
     srcs = ["tests/test.rs"],
     deps = [


8Keep (approved):

8Keep:
After this, does the whole project parse successfully?

lf-:
Alas no:

```
Error evaluating expression:
    //...
    ^---^


Caused by:
    0: From load at shed/completion_verify/packages/BUCK:1
    1: File not found: gh_facebook_buck2_shims_meta//buck2/shed/rpm_download/packages.bzl
```

I think this probably needs a way to disable the test on non-meta systems.

8Keep:
If you run buck targets with --keep-going, maybe --jsonl to give you a list of all the parse errors in the repo, that'd be a good task to put up, to give the full picture.

lf-:
> If you run buck targets with --keep-going, maybe --jsonl to give you a list of all the parse errors in the repo, that'd be a good task to put up, to give the full picture.

Only one I still see is that rpm download thing. Which we could also throw an agent at reimpl'ing for all it matters! I just worry that it depends on fb's package list to translate to an actual concrete URL, which, well. um... nobody else has one of those except Antlir whose exported package lists are broken (.. and me.. i guess.. but my random Oracle S3 experiment is Not production infrastructure https://github.com/lf-/antlir/tree/fedora44).

PR #1383, opened 2026-07-21
pagable: Fix stale DataKey documentation

davidbarsky:
Stale rustdoc that wasn't caught by Phabricator, I think. Should allow https://github.com/facebook/buck2/pull/1357 to pass.


davidbarsky:
Closing since @8Keep is fixing internally.

PR #1384, opened 2026-07-21
Rename rust-toolchain to rust-toolchain.toml

daandemeyer:
rustup supports both, but the extensionless variant is not recognized by some tools (e.g. mise), so switch to the recommended .toml name and update all references to it.


dtolnay (commented):
Thanks!

dtolnay (approved):

PR #1393, opened 2026-07-22
shim/rust: fix a few crate fixups for OSS buck2-on-buck2

thoughtpolice:
When trying to bootstrap buck2-on-buck2, we need a few changes to the default fixups. These don't fix everything perfectly yet but they're more correct anyway.

- `lalrpop` is no longer used
- `ctor` (on macOS IIRC) needs to have `cargo_env` set
- `ring` on msvc windows was passing an invalid build flag
- `zstd` crate enabled multithreading, but the fixup didn't apply `-DZSTDMT` as required
- `crossbeam` has seen various updates and these were just rewritten.

Note that while there are 4 commits I think it's probably fine to just land this stack as one commit through Phabricator, which is probably the default behavior still.


dtolnay (approved):
Thanks!

PR #1396, opened 2026-07-22
blake3 fixups: fix unsafe overbroad enabling of fancy instructions

lf-:
Previously to this commit, these instructions were inadvertently enabled in the portable (must run on all systems) sections of the library.

This PR also adds support for target_compatible_with in shim/third-party/rust, eliminating `compatible_with` bugs with the previous definitions.

### DIFF
--- a/shim/third-party/rust/fixups/blake3/fixups.toml
+++ b/shim/third-party/rust/fixups/blake3/fixups.toml
@@ -13,7 +13,10 @@ buildscript.run = false
 ['cfg(target_arch = "x86_64")']
 cfgs = ["blake3_sse2_ffi", "blake3_sse41_ffi", "blake3_avx2_ffi", "blake3_avx512_ffi"]
 
-# , any(target_env = "fbcode", target_env = "gnu")
+# NOTE: extra vector instructions must NOT be enabled on this library. blake3_dispatch.c does the
+# runtime CPUID check and blake3_portable.c is the baseline fallback, so both of them execute on
+# machines with no fancy instructions at all. Telling the compiler the target supports AVX-512 lets
+# it auto-vectorize those two TUs with AVX-512 so they crash at runtime on incompatible systems.
 [['cfg(all(target_arch = "x86_64", any(target_os = "linux", target_os = "macos")))'.cxx_library]]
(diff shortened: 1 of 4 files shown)


shim/third-party/rust/fixups/blake3/fixups.toml
```diff
@@ -22,18 +25,31 @@ srcs = [
...
     "c/blake3_sse41_x86-64_unix.S",
-    "c/blake3_avx2_x86-64_unix.S",
-    "c/blake3_avx512_x86-64_unix.S"
+    "c/blake3_avx2_x86-64_unix.S"
 ]
+headers = ["c/*.h"]
+compatible_with = [
+    "prelude//os:linux-x86_64",
+    "prelude//os:macos-x86_64",
+]
```
dtolnay: Fixups.toml `compatible_with` is not a good fit for expressing real-world Cargo `cfg` predicates of any complexity. Please see if you can do something like https://github.com/dtolnay/buck2-rustc-bootstrap/pull/94/files. This is what we use internally for the great big third-party/rust (6000 crates) which has 0 `compatible_with` fixups, only 2 `target_compatible_with` fixups, and supports `buck2 build ...` on all platforms.

The important bits are `platform_compatibility_on_all_targets = true` and `selects.apply(REINDEER_PLATFORMS, lambda p: target_compatible_with if p in platforms else ["prelude//:none"])`.

dtolnay (changes requested):

shim/third-party/rust/fixups/blake3/fixups.toml
lf-: oh, thank you! I've read through that and tried to make sense of it. I pushed a new version using this functionality. Seems to work for me, and it's way nicer. maybe some of this should land in the prelude at some point so it's easier to use out of the box (e.g. cargo.rust_library doesn't set target_compatible_with without a wrapper).

lmk if you want me to `jj rebase` the parts into a separate PR, i have no idea how the fb import tooling works and if it is annoying to have unrelated changes land in the same PR. my commit hygiene is good so this takes 5 seconds.

dtolnay (approved):
Thanks!

PR #1398, opened 2026-07-23
rust-project: allow rust-project to define test runners invocations; less adhoc unit-test mappings

davidbarsky:
This PR does three things:
1. Unit and integration tests are now modeled through explicit rule relationships, such as `testsof` and `unit_test_of`.
2. rust-project now allows overriding the test runnable invocation.
3. Modify `with_rust_workspace` to accept arbitrary globs, not just rooted crates. It also grew exclusions for when paired with globs.

Each commit should work independently to account for dotslash rollouts.


integrations/rust-project/src/target.rs
```diff
@@ -118,6 +121,7 @@ pub(crate) struct TargetInfo {
     /// This is generic metadata about the target.
     pub(crate) labels: Vec<String>,
 
+    #[serde(flatten)]
```
Wilfred, line 124: This needs to be both backwards and forwards compatible: we want rust-project to accept both old and new bxl formats. I can't guarantee exactly what commit a user's bxl will be (and the bxl changes land immediately, whereas the rust-project binary takes a little longer on dotslash).

Wilfred (changes requested):
I have a concern about compatibility, and could you also split these into three PRs? The seem fairly orthogonal.

PR #1399, opened 2026-07-23
prelude rust: allow NIX_DEBUG through env allowlist

lf-:
this allows you to dump the linker/compiler args if you're having a bad time, and diagnose what exactly went into it. it's a really useful capability, independent of nix, but is helpful to figure out what wrapper shenanigans were at fault for problems.


dtolnay (approved):
Thanks!

PR #1406, opened 2026-07-24
Fix let_underscore_drop lint error in dice_tracker

daandemeyer:
OSS CI (`test.py` / `cargo clippy`) is currently failing on every PR and on main itself with:

```
error: non-binding let on a type that has a destructor
   --> app/buck2_server/src/dice_tracker.rs:138:9
```

0578509c0a added the `HydrationFailed` variant with a `String` field to `DiceEvent`, giving it a destructor. That armed the deny-listed `let_underscore_drop` lint (`lint_levels.bzl`) on the pre-existing `let _ =` in `BuckDiceTracker::event`. Bind to `_ignored` instead, matching other call sites that discard `unbounded_send` results.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/buck2_server/src/dice_tracker.rs
+++ b/app/buck2_server/src/dice_tracker.rs
@@ -135,6 +135,6 @@ impl BuckDiceTracker {
 
 impl DiceEventListener for BuckDiceTracker {
     fn event(&self, event: DiceEvent) {
-        let _ = self.event_forwarder.unbounded_send(event);
+        let _ignored = self.event_forwarder.unbounded_send(event);
     }
 }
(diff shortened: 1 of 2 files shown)


ndmitchell (approved):

daandemeyer:
The macos failure here is an unrelated timing flake: `epoch::worker::tests::test_cancellation_sync_correct_context_does_not_cache` panicked with `elapsed_ms=650 but expected < 650` — exactly on the boundary on a slow runner. The same test passed on linux and windows in this run, and the identical two commits went green on all three platforms (macos included) in the CI run for #1364, which stacks on top of this fix.

ndmitchell:
Thanks!

PR #1408, opened 2026-07-24
Widen timing margin in cancellation caching tests

daandemeyer:
The upper-bound assert measures a window that includes the full CANCELLATION_WAIT_MS (500ms) sleep, leaving only ~150ms of slack for sleep overshoot, scheduling, and the recompute itself before tripping the 650ms limit. On loaded CI runners (macos especially) that boundary is hit in practice: elapsed_ms=650 with limit 650.

A genuinely broken cancellation would wait out the first compute and take ~FIRST_COMPUTE_MS (2500ms), so the limit can be far looser without weakening the check. Allow half of FIRST_COMPUTE_MS of slack (limit now 1775ms), which stays well clear of both the healthy (~525ms) and broken (~2500ms) regimes.

### DIFF
--- a/dice/dice/src/epoch/worker/tests.rs
+++ b/dice/dice/src/epoch/worker/tests.rs
@@ -1316,7 +1316,7 @@ async fn _run_cancellation_caching_test(
     // Ensure that the time it took to run the test is expected
     match cancel_type {
         CancelType::Sync | CancelType::ASync => {
-            let limit = (FIRST_COMPUTE_MS as f64) * 0.05 + (CANCELLATION_WAIT_MS as f64) * 1.05;
+            let limit = (FIRST_COMPUTE_MS as f64) * 0.5 + (CANCELLATION_WAIT_MS as f64) * 1.05;
             assert!(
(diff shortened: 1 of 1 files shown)


dice/dice/src/epoch/worker/tests.rs
```diff
@@ -1316,7 +1316,13 @@ async fn _run_cancellation_caching_test(
     // Ensure that the time it took to run the test is expected
     match cancel_type {
         CancelType::Sync | CancelType::ASync => {
-            let limit = (FIRST_COMPUTE_MS as f64) * 0.05 + (CANCELLATION_WAIT_MS as f64) * 1.05;
+            // The measured window includes the full CANCELLATION_WAIT_MS sleep, so
```
8Keep: This comment kinda just describes the change that you're making ("where the previous 5% margin"). It would be better if you can either just have no comment, or describe this math simply.

8Keep (changes requested):

8Keep (approved):

PR #1421, opened 2026-07-26
test: don't print passing test details by default

thoughtpolice:
This makes the default test output much, much easier to parse, navigate, and read; `stdout` and `stderr` for working tests can completely drown things out, and when you have thousands of tests there's no need to see it

The current behavior can be restored with two different flags:

     --print-passing-details

or

     -v=test_passing_details

This is often useful for dialing in a specific test so that you can analyze it. Note: this does not change the _failing_ behavior, which is to still print the output streams, as expected.

Example of before vs after:

---

**Before**

```
a@spark-92da ~/a/t/a/libedsign> buck test --print-passing-details ...
✓ Pass: depot-tilde//aseipp/libedsign:test-fingerprint (0.0s)
---- STDOUT ----
result: OK

---- STDERR ----

✓ Pass: depot-tilde//aseipp/libedsign:test-roundtrip (0.0s)
---- STDOUT ----
result: OK

---- STDERR ----

✓ Pass: depot-tilde//aseipp/libedsign:test-rekey (0.0s)
---- STDOUT ----
result: OK

---- STDERR ----

Build ID: 09771699-0b17-4860-8be9-7657a8d479db
...
```

---

**After**

```
a@spark-92da ~/a/t/a/libedsign> buck test ...
✓ Pass: depot-tilde//aseipp/libedsign:test-roundtrip (0.0s)
✓ Pass: depot-tilde//aseipp/libedsign:test-fingerprint (0.0s)
✓ Pass: depot-tilde//aseipp/libedsign:test-rekey (0.0s)
Build ID: ae93786d-2993-42a9-8711-1140add76160
...
```

### DIFF
--- a/app/buck2_client/BUCK
+++ b/app/buck2_client/BUCK
@@ -40,6 +40,7 @@ rust_library(
         "//buck2/app/buck2_data:buck2_data",
         "//buck2/app/buck2_error:buck2_error",
         "//buck2/app/buck2_event_log:buck2_event_log",
+        "//buck2/app/buck2_event_observer:buck2_event_observer",
         "//buck2/app/buck2_fs:buck2_fs",
         "//buck2/app/buck2_hash:buck2_hash",
         "//buck2/app/buck2_query_parser:buck2_query_parser",
--- a/app/buck2_client/Cargo.toml
+++ b/app/buck2_client/Cargo.toml
@@ -17,6 +17,7 @@ buck2_core.workspace = true
 buck2_data.workspace = true
 buck2_error.workspace = true
 buck2_event_log.workspace = true
+buck2_event_observer.workspace = true
 buck2_fs.workspace = true
 buck2_hash.workspace = true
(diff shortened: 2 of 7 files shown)


8Keep:
Could you include a demo of before and after in the pr description?

thoughtpolice:
> Could you include a demo of before and after in the pr description?

Done, I'll keep that in mind!

lf-:
Maybe a better CLI design in our fork: https://github.com/facebook/buck2/commit/e73f56c22379f13107fbe0affef6b103f622a9b9

`--test-output=all`, `--test-output=error`, `--test-output=none`.

cormacrelf:
Is there a reason this ought to be part of the test runner protocol (both here via verbosity and in mercury fork directly) and not just in the default test runner's CLI args? It already has `--test-arg=... --test-env=VAR=val`. Plausibly a different runner might have a totally different UI... like a TUI, etc.

PR #1422, opened 2026-07-26
test: allow `buck2.test_builds_targets` to build all RunInfo and DefaultInfos by default

thoughtpolice:
https://github.com/facebook/buck2/pull/1044 was landed long ago and added support for building RunInfos and DefaultInfos when you run `buck2 test`. One problem: it was never wired into the buckconfig allowing you to specify a default. That means you had to specify `--build-default-info` every single time which was pretty tedious outside of CI.

But, while it doesn't work forever, `buck2 test $TARGET_PATTERN` being a superset of `buck2 build` is pretty useful for onboarding new users, IMO, so they can rely on one command to handle most of their integration patterns ("will this pass CI"). So, I think finally wiring it in is a nice touch. The default is still false, and as a bonus the currently do-nothing CLI parameters will actually override these settings (they were added but were dead code).


thoughtpolice:
Oh, and this has some Claude'd tests, though you can run some of these tests manually with some fiddling, which is what I did (full e2e does not work).

I can do this (or not) depending on the preference of people upstream, since obviously I don't have visibility into how these run in your harness.

lf-:
I remember @csamak chiming in on this in team discussions in opposition of this idea, the idea basically being that you want to test things that actually matter in some sense (and also adding some level of implicit pressure to write tests). This led to us writing a copy of the skylib `build_test` rule to *explicitly* specify the targets that are intended to be built with `buck test`.

This isn't to say that other cultures around doing this are invalid though, and I think it's worth supporting in buck2, it's just there's alternative ideas on how to go about this, I guess. Cameron might have more specific thoughts.

csamak:
I worry for wide target patterns it would have the potential to increase the workload somewhat surprising ways, for example by reducing deferred materialization opportunities.

As default-off I don't have any concerns, though. I think it's useful to have as an option locally, even if we might choose not to run it in CI.

8Keep (approved):

thoughtpolice:
Thank you a ton Davis!

8Keep:
I don't think we'll be able to turn this on internally for a while. I started the discussion on that. There are a lot of things that would be worse with this on... But maybe long term we'll get there

PR #1425, opened 2026-07-29
Don't kill a daemon pid that is not a buck2 process

bvinc:
buckd.info records a bare pid. If that daemon died without cleaning up, the OS may have reused the pid, so hard_kill_impl could terminate an unrelated process
- including an ancestor of the client, which kills the client itself.

Check identity before killing, using the same rules killall already applies: a buck2 executable, and not an ancestor of this process. Unknown identity still kills, so callers that report the kill to the user keep today's behavior.


8Keep (approved):

PR #1429, opened 2026-07-31
interpreter: load data files with an explicit `?format=` override

daandemeyer:
Loading static data with `load()` required the file to be named `.json` or
`.toml`, because the extension was the only thing selecting a parser. That rules
out the files people actually want to read: `Cargo.lock` is TOML, a
`package.json.tmpl` is JSON, and neither can be renamed.

Add a `?format=` query on the load path (`load(":deps.lock?format=toml", ...)`)
accepting `bzl`, `json`, or `toml`. It cannot be a `load()` kwarg: `format =
"toml"` is already valid syntax meaning "alias the local name `format` to the
exported symbol `toml`", so taking that spelling would need a grammar change and
would silently change the meaning of existing code. The `?` was already reserved
by `ImportPath`.

The format is stored on `ImportPath` and participates in its hash and equality,
so it is part of the DICE key rather than a side channel. That makes the loaded
file a tracked dependency of the parse and gets invalidation for free, and the
extension check is skipped exactly when a format was given. Every site that
dispatched on the extension now asks `ImportPath::load_format()` instead.

A format that merely restates the extension collapses to `None`, so
`:a.bzl?format=bzl` and `:a.bzl` stay one module. Otherwise the same file would
be evaluated under two keys and hand out two copies of its providers and
transitive sets, breaking the pointer-identity checks that `resolve_load`
already warns about.

An unparseable file stays a load-time error, which fails the whole package.
That is inherent: a parse that is not at load time is not a tracked dependency.
Since the extension no longer implies the parser, the error now names the format
and where it came from. The blast radius is documented, with a pointer to
reading the file in an action when failure needs to be scoped to its consumers.

Incidentally fixes prelude `.toml` files being evaluated as Starlark: the
prelude branch special-cased only `.json`, and now shares the format dispatch.

Covered by unit tests for format inference, the relaxed extension check, the
redundant-format collapse and display round-tripping, plus e2e tests for the
three formats, invalidation on edit, and each error path.

Signed-off-by: Daan De Meyer <daan@amutable.com>


cormacrelf:
See https://github.com/facebook/buck2/pull/1287

The hashing part where the ImportPath for `.toml?as=toml` should not hash differently from `.toml` is a good catch, I will have a look.

cormacrelf:
We need elements of both approaches. Do you mind if I integrate what you've done in the other PR?

Things we need from here, roughly
- Deduplication for noop format changes, although won't make as much different without ?as=bzl
- Most of your tests can be folded in
- The improved error messages

Modulo:
- Stick with ?as= syntax
- probably won't support ?as=bzl after all the discussion on that thread. Unless you have a plausible use case? If a file is starlark surely it is called .bzl?

With maybe one improvement in light of the ImportPath normalisation... maybe we need another path type LoadPath that is not normalised. Then we can keep ImportPath fully normalised and prevent mistakes like mine in future.

daandemeyer:
@cormacrelf Please do, I don't have an immediate need for these changes anymore, so I'll close this one.

PR #1430, opened 2026-07-31
Show all constraints when formatting the config name

cbarrete:
This is important for OSS, who cannot add their modifiers to the hardcoded `NAMED_CONSTRAINT_SETTINGS` dict.

### DIFF
--- a/prelude/cfg/modifier/name.bzl
+++ b/prelude/cfg/modifier/name.bzl
@@ -53,6 +53,19 @@ def cfg_name(cfg: ConfigurationInfo) -> str:
         if constraint_name:
             name_list.append(constraint_name)
 
+    # Append any remaining constraints not covered by NAMED_CONSTRAINT_SETTINGS,
+    # sorted by constraint setting for determinism. Prefer the subtarget name
+    # (e.g. `debug` in `:compile_mode[debug]`) when present, since with the
+    # unified `constraint()` rule the target name is the constraint setting.
+    for constraint_setting in sorted(constraints):
+        if constraint_setting in NAMED_CONSTRAINT_SETTINGS:
+            continue
+        label = constraints[constraint_setting].label
+        if label.sub_target:
+            name_list.append(str(label.sub_target[0]))
+        else:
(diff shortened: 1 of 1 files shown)


8Keep:
Please give some better description in the PR, we should update the contribution guidelines for this. I'd want at least the output and what it looks like.

I'd prefer some kind of setup where the user can set this function in their `set_cfg_constructor()`, so it's easy to customize if they want, with good defaults. Or, could be combined somehow with the aliases that exist there, by default?

Probably want @scottcao to think about how he'd want this to be setup too.

prelude/cfg/modifier/name.bzl
```diff
@@ -53,6 +53,19 @@ def cfg_name(cfg: ConfigurationInfo) -> str:
         if constraint_name:
             name_list.append(constraint_name)
 
+    # Append any remaining constraints not covered by NAMED_CONSTRAINT_SETTINGS,
```
lf-, line 56: I'd been meaning to put a replacement of `NAMED_CONSTRAINT_SETTINGS` in `set_cfg_constructor`. Is that something which is possible/more advisable to solve this problem?

8Keep: Yes

cbarrete:
I've had some more time to play with this, and I agree that this is not the right approach.

I think that we need to make `cfg_name` configurable, but it also needs to be called (likely with more arguments) for things like transitions.
It's unclear to me what the idiomatic way of making it configurable would be.

For now, I am going to wrap `cfg_constructor_post_constraint_analysis` to call a custom `cfg_name` on the returned platform, keeping its constraints but formatting a different label.
The big downside is that again, it doesn't handle transitions.

prelude/cfg/modifier/name.bzl
cbarrete: FWIW, the custom `cfg_name` that I ended up writing ended up having quite a bit more logic than is conveniently handled by a list of transformers, so it's unclear to me that just expanding the list/making it configurable is the right way to go.

PR #1434, opened 2026-08-02
file_watcher: rebuild the watch tree when its coverage may be incomplete

martinpitt:
A directory on which the daemon holds no inotify watch is invisible:
nothing under it produces an event, so buck keeps serving whatever it
read last. This had two causes:

 - Dropped events cleared DICE, but the directories created while they
   were being dropped were never watched.

 - A watcher error discarded the events buffered with it and failed the
   next command, after which the daemon carried on as if nothing had
   been missed.

To fix this, treat both as the same thing: clear DICE, as before, and
then register the tree again so that it covers the project. The new
registration is built before the old one is dropped, so nothing goes
unwatched in between, and the overlap only duplicates events.

Paths that cannot be watched at all (like an inaccessible directory) are
remembered, so that they trigger that recovery once rather than on every
command.

This depends on the `notify` crate actually reporting these failures to
us, which needs https://github.com/notify-rs/notify/pull/970. Until that
lands, a failed watch install is not among the errors that reach us.

The new unit test passes with that notify fix; it stays `#[ignore]`d
here until a notify release carrying it is picked up.

Signed-off-by: Martin Pitt <martin@amutable.com>

----

This was several hours of debugging.. there was an unreadable directory appearing during some split second, which cut off an entire branch of the watched tree, and from then on the daemon kept sending old data until I killed it. That effect was very subtle (and hence dangerous!), I just kept wondering why my changes to the tree didn't have any effect and I got outdated file contents.

Warning: This was done in large parts by Claude Opus 5. I fully understand the bug root cause and the [notify fix](https://github.com/notify-rs/notify/pull/970) and the test case, but I don't know about the innards of buck2, so the actual file watcher code changes need an architecture critique, and I'm happy to rework it.

The second commit adds a `buck2 debug watches` command, to make this easier to investigate. It's of course not critical, might just save the next pour soul some hours. Let me know if you want this or not.

### DIFF
--- a/app/buck2_cmd_debug_client/src/lib.rs
+++ b/app/buck2_cmd_debug_client/src/lib.rs
@@ -37,6 +37,8 @@ use crate::set_log_filter::SetLogFilterCommand;
 use crate::thread_dump::ThreadDumpCommand;
 use crate::trace_io::TraceIoCommand;
 use crate::upload_re_logs::UploadReLogsCommand;
+#[cfg(target_os = "linux")]
+use crate::watches::WatchesCommand;
 
 mod allocative;
 mod allocator_stats;
@@ -59,6 +61,8 @@ mod set_log_filter;
 mod thread_dump;
 mod trace_io;
 mod upload_re_logs;
+#[cfg(target_os = "linux")]
+mod watches;
 
 #[derive(Debug, clap::Parser)]
 #[clap(about = "Hidden debug commands useful for testing buck2")]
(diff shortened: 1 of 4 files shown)


martinpitt:
Rebased to fix the conflict. I also added a third commit which fixes an independent regression on main to re-fix the tests. Re-validated against https://github.com/notify-rs/notify/pull/970 .

martinpitt:
Project admins: https://github.com/notify-rs/notify/pull/970 now landed upstream :tada:

PR #1435, opened 2026-08-02
prelude/git: Expose the fetched repository as a `.git` sub-target

martinpitt:
`git_fetch()` declares the `.git` directory as an output, but previously
only handed out the checked-out work tree.

Expose the existing output as a `.git` sub-target next to the
path-projection sub-targets. The name cannot collide: git refuses to
track a path named `.git`, so no requested work-tree projection can
ever claim it.

This is useful for e.g. fetching cargo git repos: cargo's source
replacement (`[source.x] git = "file://..."`) fetches from the object
database, which is how a Rust crate's git dependencies can be served
locally. A work tree cannot be re-initialized as a git repo, as that
would not carry the actual pinned hash.

Flip the previous commit's test accordingly: the sub-target holds the
pinned commit in an object database.

---

Plus a preparatory commit that introduces `git_fetch()` coverage in the first place -- it didn't have any tests before.

### DIFF
--- a/prelude/git/git_fetch.bzl
+++ b/prelude/git/git_fetch.bzl
@@ -63,9 +63,16 @@ def git_fetch_impl(ctx: AnalysisContext) -> list[Provider]:
         allow_cache_upload = ctx.attrs.allow_cache_upload,
     )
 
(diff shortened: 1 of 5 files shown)


8Keep (approved):

PR #1436, opened 2026-08-03
DICE examples: fix build on macOS

lf-:
this is because the jemalloc crate requires a prefix on macOS which then causes build failures when you try to link the unprefixed symbol. it of course ignores the cargo feature saying to not prefix it (lol) and doesn't expose the mallctl api in a way that's agnostic to the prefixing (lol) so you have to cfg it at the use site.


8Keep (approved):

PR #1444, opened 2026-08-05
fix: python bootstrap uses relative paths

### LINKED ISSUE #1135: python_binary "par" file contains absolute path from RE
When doing a remote build for the "//:alpha" target within https://github.com/seesturm/buck2-python-bug it fails.

```
/usr/bin/env: '/exp2/nativelink/work/ea4d39af-1f2d-4cd9-bfcb-391f37b5a9f3/work/buck-out/v2/gen/toolchains/61376d016e6fd92f/__cpython_archive__/cpython_archive/bin/python': No such file or directory
```
The reason appears to be that an absolute path is hardcoded into the "par" file created by "make_py_package".
```bash
head -n2 buck-out/v2/gen/root/2c621926a02f7469/__run__/run.par
#!/usr/bin/env /exp2/nativelink/work/4fca9ae6-800d-4a3f-b99a-9791334e95d6/work/buck-out/v2/gen/toolchains/2c621926a02f7469/__cpython_archive__/cpython_archive/bin/python 
# Copyright (c) Meta Platforms, Inc. and affiliates.
```

### LINKED ISSUE #1268: Python bootstrapper pex file hermeticity
I've encountered an issue with the make_py_package_inplace.py script used in Buck2 for creating .pex files. The script currently includes the absolute path of the Python interpreter in the shebang line. 

This works if you're passing a system binary as an argument when invoking it, but we define the python interpreter to use in our project (using nix).

If the action creating the .pex file is run on a remote executor, the path will reference the python binary inside the temporary build directory:
```
#!/usr/bin/env /build/work/3a1d9a9b-5eba-4272-bd71-ce3e07f3540a/work/buck-out/v2/gen/toolchains/b42aeba648b8c415/__python_bootstrap_interpreter__/out.link/bin/python
```

Any other dependent action using this script will fail with "no such file or directory".

Because of this, these outputs cannot be cached.

I see this was recently (well Aug 2025, but we updated from 2025-06-15 😄) changed by commit 5f5f6927c30b28dbfba5126708345e8260079481

> These paths are clearly relative to something, and so the resulting par would only be runnable when CWD is that something. That's suboptimal.

Yes, but I think using relative paths (relative to the project root) should be the most idiomatic way. I can see that this will not work if you run this from a different directory. But if this is needed, those files could be modified by another action doing something sensible. 

WDYT @zsol

lf-:
This got us good at Mercury and took down some builds when we enabled `default_allow_cache_upload = true`.

The primary bug: make_py_package_inplace writes a bootstrap script like so:

```python
python = str(args.python)
if not os.path.isabs(python) and os.path.sep in python:
    python = os.path.abspath(python)

new_data = data.replace("<PYTHON>", f"/usr/bin/env {python}")
```

Absolutizing the interpreter path causes this to be a cache poisoning bug if ever uploaded. In our case, our python interpreter is an artifact, so it gets made absolute, thus baking our GitHub actions home directory into the uploaded actions.

This bug came from 5f5f6927c3 (D79552056) which was endeavouring to make calling PARs not rely on the working directory anymore which it did successfully but via means it calls a "terrible idea".

Let's do this in a way that's not terrible.

Why didn't this hit Meta: the RE sandbox is always at `/re_cwd`, and the Python interpreter is usually(?) at an absolute path.

This also fixes a bug where `allow_cache_upload` (via `cxx_attrs_get_allow_cache_upload`) got dropped before hitting the PAR actions, so it's not even possible to turn it off on known non-hermetic actions.

The methodology behind this PR was that I told some agents to find how Bazel fixes this Python bootstrapping problem and learn all the footguns they hit.

One of their impls: https://github.com/bazel-contrib/rules_python/blob/16e167a5b269930f2f5000c4d9dc99b72f930f2b/python/private/pypi/venv_shebang_rewriter.sh

Here's a brief list:

1. `!#/bin/bash` is not compatible with NixOS: `#!/usr/bin/env bash` or `#!/bin/sh` is preferable.
2. Apparently Bazel has a bug with their wrappers' $0 where they got symlink handling wrong: it needs to walk component by component.
3. `exec -a` is not POSIX, so `dash` doesn't implement it. We have to use a `sh -c` trick instead.

You know the deal, the tests are vibed, you can throw them out or clean them up as desired, since I can't run them without being at fb.

Fixes: https://github.com/facebook/buck2/issues/1135
Fixes: https://github.com/facebook/buck2/issues/1268

### DIFF
--- a/prelude/python/make_py_package.bzl
+++ b/prelude/python/make_py_package.bzl
@@ -591,7 +591,12 @@ def _make_py_package_impl(
         modules.add(modules_args)
         if package_style == PackageStyle("outplace"):
             modules.add(cmd_args("--copy-files", hidden = runtime_artifacts))
-        ctx.actions.run(modules, category = "par", identifier = "modules{}".format(output_suffix))
+        ctx.actions.run(
+            modules,
+            category = "par",
+            identifier = "modules{}".format(output_suffix),
+            allow_cache_upload = allow_cache_upload,
+        )
 
         bootstrap = cmd_args(python_internal_tools.make_py_package_inplace)
         bootstrap.add(bootstrap_args)
@@ -604,7 +609,12 @@ def _make_py_package_impl(
         for flag in ctx.attrs.interpreter_args:
             bootstrap.add(cmd_args(["--python-interpreter-flags={}".format(flag)]))
 
-        ctx.actions.run(bootstrap, category = "par", identifier = "bootstrap{}".format(output_suffix))
+        ctx.actions.run(
+            bootstrap,
+            category = "par",
+            identifier = "bootstrap{}".format(output_suffix),
+            allow_cache_upload = allow_cache_upload,
+        )
 
     run_args = []
 
--- a/prelude/python/tools/BUCK
+++ b/prelude/python/tools/BUCK
@@ -120,11 +120,13 @@ prelude.python_bootstrap_binary(
 )
 
 # Run the test suite with this command:
-# buck2 run prelude//python/tools:wheel_tests --target-platforms prelude//platforms:default
+# buck2 run prelude//python/tools:tool_tests --target-platforms prelude//platforms:default
 prelude.sh_binary(
-    name = "wheel_tests",
+    name = "tool_tests",
     main = "tests/main.sh",
     resources = [
+        "make_py_package_inplace.py",
+        "run_inplace.py.in",
         "wheel.py",
     ]
     + glob(["tests/**/*.py"]),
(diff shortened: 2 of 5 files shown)


8Keep (approved):

itamaro:
thanks for the PR @lf- !

I used this to "inspire" a slightly modified fix I'm landing internally now, focused on the relative paths issue.

I didn't include the cache upload fix, as it's unrelated to the relative paths fix - we will look into that as a separate fix to follow shortly.

lf-:
thanks for merging this!

8Keep:
I made some changes to the allow_cache_upload here https://github.com/facebook/buck2/commit/009bec0ae8692a68ec70e836d98552a4a906176e

PR #1447, opened 2026-08-05
Make clean stale scan art-anon and art-bxl dirs

bvinc:
Anon target and bxl outputs are written to sibling directories rather than under `art`: their `make_hashed_path` impls append `-anon`/`-bxl` to the path prefix. The scan root list named only `gen` and `art`, so the walk never reached them and artifacts there were never cleaned, either as stale or as untracked.


bvinc:
Is there any reason not to include these directories in a stale-clean?

cormacrelf:
I can't think of any reason why not.

scottcao (approved):

8Keep (approved):

PR #1448, opened 2026-08-06
Do not underflow when retained_event_logs is zero

### LINKED ISSUE #1431: setting `retained_event_logs = 0` skips old-log cleanup
In `app/buck2_event_log/src/file_names.rs`, `remove_old_logs()` subtracts one from `retained_event_logs`:

https://github.com/facebook/buck2/blob/f6056e24ba2e079979eb5029a049401a44cc734a/app/buck2_event_log/src/file_names.rs#L47-L50

`app/buck2_common/src/init.rs` accepts zero without validation:

https://github.com/facebook/buck2/blob/f6056e24ba2e079979eb5029a049401a44cc734a/app/buck2_common/src/init.rs#L684-L690

This gets triggered if the user changes the root `.buckconfig` to `retained_event_logs = 0` while old event logs exist, then runs a Buck2 command that creates an event log:

```ini
[buck2]
retained_event_logs = 0
```

In release builds, `0 - 1` wraps to `usize::MAX`, so the iterator skips every old log and never delete any of them. Old logs can keep accumulating.

To avoid this, consider rejecting zero, or explicitly handle it before subtracting.

Sanjays2402:
Fixes #1431.

`remove_old_logs` does `skip(retained_event_logs - 1)`. The `-1` is there because the count includes the log about to be written, but nothing rejects zero: `buck2.retained_event_logs = 0` in `.buckconfig` parses straight through `init.rs` as `0usize`, and in release builds `0 - 1` wraps to `usize::MAX`. The iterator then skips every existing log, so cleanup silently stops and logs accumulate forever. Exactly backwards from what someone setting zero is asking for.

Swapped the subtraction for `saturating_sub(1)`, so zero means retain none and every old log is deleted.

I went with saturating rather than validating the config at parse time. Zero is a coherent thing to ask for, and the underflow is really a bug in the arithmetic rather than in the value, so fixing it where it happens keeps the config surface unchanged and can't break anyone already passing zero.

Added `test_remove_old_logs_with_zero_retained` alongside the existing test. It creates three logs, calls `remove_old_logs(dir, 0)` and asserts all three are gone. On the old code it fails with all three still present.

```
running 2 tests
test file_names::tests::test_remove_old_logs_with_zero_retained ... ok
test file_names::tests::test_remove_old_logs_with_mix_of_files_and_folders ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 13 filtered out
```

Run on nightly-2026-05-22 per `rust-toolchain`, macOS arm64.

### DIFF
--- a/app/buck2_event_log/src/file_names.rs
+++ b/app/buck2_event_log/src/file_names.rs
@@ -46,7 +46,10 @@ pub(crate) fn get_logfile_name(
 
 pub(crate) async fn remove_old_logs(logdir: &AbsNormPath, retained_event_logs: usize) {
     if let Ok(logfiles) = get_files_in_log_dir(logdir) {
-        futures::stream::iter(logfiles.into_iter().rev().skip(retained_event_logs - 1))
+        // `retained_event_logs` counts the log about to be written, so one fewer
+        // existing log is kept. Saturate rather than wrap: zero means retain none.
+        let retained_existing_logs = retained_event_logs.saturating_sub(1);
+        futures::stream::iter(logfiles.into_iter().rev().skip(retained_existing_logs))
             .then(|file| async move {
                 // The oldest logs might be open from another concurrent build, so suppress error.
                 tokio::fs::remove_file(file).await.ok()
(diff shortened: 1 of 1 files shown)


8Keep (approved):

PR #1453, opened 2026-08-08
builtin file watcher: set lost-events flag for a pathless rescan event

### LINKED ISSUE #1131: Rescan events from the notify crate are not being handled, causing cache inconsistencies

VesnaSpringett:
`inotify` signals a kernel queue overflow by setting a rescan flag on an event, and that event can arrive with no paths attached. `NotifyFileData::process` only checked `need_rescan()` inside its per-path loop, so a pathless rescan was ignored: the lost-events flag never got set, the next sync did not drop the DICE graph, and the daemon kept answering from a graph that no longer matched the file system, with no error and no warning.

This is the stale-cache class reported in #1131. Lost-events detection was introduced by the fix merged for it (#1149, which closed #1131 as completed, commit 0be55e4871), with the `need_rescan()` check inside the per-path loop, so the no-paths case has slipped through since that commit; there is no earlier passing state to regress from, because `missed_events` did not exist before it. The later telemetry change for the graph drop (#1310, commit 38848e455f) did not move the check.

The flag is a property of the event, not of any path in it, so this change moves the check above the loop. This also evaluates `need_rescan()` once per event instead of once per path.

Tests:

- `rescan_event_with_no_paths_sets_missed_events` is the only fail2pass (RED -> GREEN) test in this PR.
- The other tests pin behavior that must not change, and pass both before and after the fix: a rescan event that does carry paths still sets the flag; a sync with the flag set reports a fresh instance and drops the incremental tracker; a sync without it keeps the tracker.

### DIFF
--- a/app/buck2_file_watcher/src/notify.rs
+++ b/app/buck2_file_watcher/src/notify.rs
@@ -84,6 +84,14 @@ impl NotifyFileData {
         let event =
             event.map_err(|e| from_any_with_tag(e, buck2_error::ErrorTag::NotifyWatcher))?;
 
+        // Checked before the path loop: the kernel-overflow rescan signal can arrive with no
+        // paths attached (platform dependent), and a missed rescan means the daemon keeps
+        // answering from a stale graph.
+        if event.need_rescan() {
+            self.missed_events = true;
+            debug!("FileWatcher: File change events were missed");
+        }
+
         for path in &event.paths {
             // Testing shows that we get absolute paths back from the `notify` library.
             // It's not documented though.
@@ -109,11 +117,6 @@ impl NotifyFileData {
                 "FileWatcher: {:?} {:?} (ignore = {})",
                 path, &event.kind, ignore
             );
-
-            if event.need_rescan() {
-                self.missed_events = true;
-                debug!("FileWatcher: File change events were missed");
-            }
 
             if ignore || ignore_event_kind(event.kind) {
                 self.ignored += 1;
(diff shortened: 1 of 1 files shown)


8Keep (approved):

PR #1455, opened 2026-08-10
bug fix: internal_error not returned

lf-:
claude found this one while reading the code. i put a must_use on it to discourage it recurring.


scottcao (approved):

8Keep:
We had source control errors come up on a small subset of machines and had to revert for now. Will have to fix some things internally then try to make this a soft error first before turning it into a hard error

PR #1458, opened 2026-08-12
prelude//cpu/constraints:wasm32: don't regress

lf-:
On the current prelude used by Mercury, we were using prelude//cpu/constraints:wasm32
(https://github.com/MercuryTechnologies/buck2/blob/07f4a07ed01b674102b2db7db1f1348e5afe9fc2/prelude/cpu/constraints/BUCK#L46-L50) to drive a transition (so I don't think we want to use the prelude//cpu:wasm32 config_setting).

The change to use the modernized constraints removed prelude//cpu/constraints:wasm32 as a target. We need it to exist so that we can smoothly migrate (we have a rollout mechanism for new buck2 versions in CI so we need our repo to work across both new and old).

### DIFF
--- a/prelude/cpu/constraints/BUCK
+++ b/prelude/cpu/constraints/BUCK
@@ -14,6 +14,7 @@ constraint_with_aliases(
         "arm64": "arm64",
         "x86_32": "x86_32",
         "x86_64": "x86_64",
+        "wasm32": "wasm32",
     },
     default = "unspecified",
     values = [


8Keep:
Actually, constraint_with_aliases is just a backward compatibility mechanism to move to constraint() but keep the same targets as the old constraint_value() set. This is just the aliases for the constraint for backwards compatibility. We'd like to get rid of all of these aliases. Would it be acceptable to update on your side to :cpu[wasm32] instead of keeping the wasm32 constraint alias around?

I'm not sure the constraint() rfc was ever circled around externally, but we are moving away from constraint_setting/constraint_value toward just having the new constraint() rule everywhere.

lf-:
Yep, I saw the RFC. I actually think it's flawed for a certain class of constraints (cpu for example is a pretty bad one imo; Meta doesn't know about every cpu that a buck user might use) because it's closed-world. It's a great improvement for most constraints though.

The problem we're running into is that our current version of buck doesn't have the new name yet, and there's no transition state where they all exist as aliases. (and we're having a really hard time with upgrading due to mysterious cache related infra problems, which is making it hard to get anything done on upgrades so we're stuck on an even older buck. fml.).

PR #1459, opened 2026-08-13
cxx/python/rust: propagate UnixEnvInfo

lf-:
Motivation: I want to pass shared library dependencies through my binary targets, so that `snowydeer_package` (our nix package builder; see my LixCon 2026 talk and https://github.com/mercurytechnologies/snowydeer) can automatically include them in the `lib/` directory in the output.

Implementation/test plan: honestly this is a *massive* YOLO. This is clauded and I am *absolutely* not sure why this propagation was originally disabled, nor do I have enough tests out of tree to meaningfully exercise it.

A note: "we only want to traverse deps while dynamically linking" is potentially wrong when you have a setup that prefers the static link style but still allows dynamic linking (which is what we're in; we try to statically link first-party but Rust (https://github.com/rust-lang/rust/issues/73632) and other things force dynamic linking).

This is more of a question for why it was off in the first place with a "yolo" button than a normal PR :)

### DIFF
--- a/prelude/cxx/cxx.bzl
+++ b/prelude/cxx/cxx.bzl
@@ -317,9 +317,10 @@ def get_auto_link_group_specs(ctx: AnalysisContext, link_group_info: [LinkGroupI
 
 def cxx_binary_impl(ctx: AnalysisContext) -> list[Provider]:
     link_strategy = to_link_strategy(cxx_attr_link_style(ctx))
+    deps = cxx_attr_deps(ctx)
     link_group_info = get_link_group_info(
         ctx,
-        filter_and_map_idx(LinkableGraph, cxx_attr_deps(ctx)),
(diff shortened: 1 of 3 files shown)


prelude/cxx/cxx.bzl
```diff
@@ -387,9 +388,7 @@ def cxx_binary_impl(ctx: AnalysisContext) -> list[Provider]:
                     ),
                 ],
             ),
-            # TODO(agallagher): We only want to traverse deps when dynamically
-            # linking.
-            # deps = ctx.attrs.deps,
+            deps = deps,
```
christycylee, line 391: This will unconditionally traverse all deps regardless of link style. It's fine for shared links where we want all dependent shared libraries, but for static mode where we only want the main binary, we would end up packaging all transitive shared objects in the environment and bloat the env size. Could you gate this to shared links only?

PR #1461, opened 2026-08-13
buck2_test: don't poison results channel via internal orchestrator drop

### LINKED ISSUE #1479: buck2 test fails at teardown when a target uses InternalRunnerTestInfo
Any `buck2 test` run that includes a target using `InternalRunnerTestInfo` fails at teardown:

    Did not receive all results from executor: BuckTestOrchestrator
    exited before end-of-tests was received

All test results are reported fine before that. The cause is drop order in `test_targets`: the internal orchestrator is dropped too early, and its Drop poisons the results channel.

There is a second bug behind this one. A failing internal-runner test shows `Fail` in the UI, but `buck2 test` still exits 0, so CI goes green on red tests. The internal runner cannot call `end_of_test_results` (the external executor owns that call), so its failures never reach the exit code.

Fix in #1461.

vlad-ivanov-name:
Fixes #1479.

Any `buck2 test` run in which at least one target used InternalRunnerTestInfo fails at teardown with

    Did not receive all results from executor: BuckTestOrchestrator
    exited before end-of-tests was received

even though all per-test results were already reported correctly.

Root cause: drop order in test_targets.

Fix: drop the internal orchestrator after try_fold.

This fix also surfaces other issue so the next fix is the exit code for failing internal-runner tests: they showed `Fail` in the UI but `buck2 test` exited 0 (green CI on red tests).

<details>
<summary>
Reproducer
</summary>

**`.buckconfig`**
```ini
[cells]
root = .
```

**`defs.bzl`**
```python
def _listing(content: str) -> list[dict]:
    return [{"name": n, "filter": n} for n in content.split()]

def _result(stdout: str, stderr: str, exit_code: int) -> list[dict]:
    # empty -> the runner synthesizes PASS/FAIL from the exit code
    return []

def _impl(ctx):
    return [DefaultInfo(), InternalRunnerTestInfo(
        type = "sh",
        listing_command = ["/bin/sh", "-c", "echo alpha beta"],
        command = ["/bin/sh", "-c", "exit 0"],
        parse_test_listing = _listing,
        parse_test_result = _result,
    )]

internal_test = rule(impl = _impl, attrs = {})
```

**`BUCK`**
```python
load(":defs.bzl", "internal_test")

internal_test(name = "t")
```
</details>

<details>
<summary>
Before
</summary>

```
$ buck2 test //:t; echo $?
✓ Pass: alpha (0.0s)
✓ Pass: beta (0.0s)
Command failed:
Test executor exited unexpectedly with status 1.
Stdout:

Stderr:
Command failed: code: 'Unknown error', message: "Failed to report end-of-tests: end_of_tests was received twice (internal error)"
1
```
</details>

<details>
<summary>
After
</summary>

```
$ buck2 test //:t; echo $?
Tests finished: Pass 2. Fail 0. Timeout 0. Fatal 0. Skip 0. Omit 0. Infra Failure 0. Build failure 0
0
```

With a failing command (`exit 1` for test `beta`):

```
$ buck2 test //:t; echo $?
Tests finished: Pass 1. Fail 1. Timeout 0. Fatal 0. Skip 0. Omit 0. Infra Failure 0. Build failure 0
1 TESTS FAILED
  ✗ beta (<unspecified>)
32
```
</details>

### DIFF
--- a/app/buck2_test/src/command.rs
+++ b/app/buck2_test/src/command.rs
@@ -137,14 +137,24 @@ struct TestOutcome {
 impl TestOutcome {
     fn exit_code(&self) -> buck2_error::Result<i32> {
         self.executor_report
-            .exit_code
+            .exit_code()
             .internal_error("Test executor did not provide an exit code")
     }
 }
 
+/// Exit code for a run in which tests failed;
+/// matches `RunVerdict::Fail` in the OSS test runner
+const TESTS_FAILED_EXIT_CODE: i32 = 32;
+
 #[derive(Default)]
 struct ExecutorReport {
-    exit_code: Option<i32>,
+    /// Exit code reported by the external test executor via
+    /// `end_of_test_results`; `None` if the executor exited without
+    /// reporting. Not the run's verdict — see `exit_code()`.
+    executor_exit_code: Option<i32>,
+    /// Whether any target ran in-process. Only then can the executor's
+    /// exit code miss results collected here.
+    used_internal_runner: bool,
     statuses: TestStatuses,
     info_messages: Vec<String>,
 }
@@ -156,11 +166,22 @@ impl ExecutorReport {
                 self.statuses.ingest(res, session);
             }
             ExecutorMessage::ExitCode(exit_code) => {
-                self.exit_code = Some(*exit_code);
+                self.executor_exit_code = Some(*exit_code);
             }
             ExecutorMessage::InfoMessage(message) => {
                 self.info_messages.push(message.clone());
             }
+        }
+    }
+
+    /// Verdict for the whole run. The executor's exit code is authoritative
+    /// unless it reports success while internal-runner tests failed.
+    fn exit_code(&self) -> Option<i32> {
+        match self.executor_exit_code {
+            Some(0) if self.used_internal_runner && self.statuses.has_failures() => {
+                Some(TESTS_FAILED_EXIT_CODE)
+            }
+            code => code,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


vlad-ivanov-name:
Turns out this requires some adjustments in how exit code is derived. I will fix that and mark it ready for review after.

Edit: fixed

app/buck2_test/src/command.rs
```diff
@@ -797,9 +825,9 @@ async fn test_targets(
                     std::mem::replace(&mut driver.build_target_result, BuildTargetResult::new());
                 drop(driver);
 
-                // Drop internal runner resources so their senders don't
-                // keep the results channel open during try_fold.
-                drop(internal_orchestrator);
+                // Drop the internal runner's sender clone so it doesn't keep
+                // the results channel open during try_fold. Orchestrator is not
+                // dropped yet -- see below.
```
k-cross, line 830:
```
// Whether any target ran in-process deciding whether the exit code accounts for every result collected.
let used_internal_runner = internal_orchestrator.initialized();
```

app/buck2_test/src/command.rs
```diff
@@ -816,6 +844,10 @@ async fn test_targets(
                     .await
                     .buck_error_context("Did not receive all results from executor")?;
 
+                // The results channel is closed now, so the internal
```
k-cross, line 847: Before the drop update the test_status:
```
test_statuses.used_internal_runner = used_internal_runner;
```

k-cross (commented):
I'm not sure if the macos failures are from the rust tests but i have very similar changes with these two additions that pass locally on my mac fwiw

vlad-ivanov-name:
Mac failures are unrelated I believe.

app/buck2_test/src/command.rs
vlad-ivanov-name: done

app/buck2_test/src/command.rs
vlad-ivanov-name: done

vlad-ivanov-name:
Updated with suggestions.

k-cross:
When I look at this [comment](https://github.com/facebook/buck2/pull/1461#issuecomment-5280909166), I wonder if your changes reflect their internal management system. I've never worked for FB so I wouldn't know. If not, I don't know if it's worth opening a new PR to get the changes reflected internally.

vlad-ivanov-name:
"Import" stage runs on every push. But if it doesn't get the PR anywhere I'll reopen both issue and PR

k-cross (commented):
I've run this successfully locally on top of my own changes. I don't have approval power but I do approve of the changes!

PR #1462, opened 2026-08-13
Don't pass unconfigured targets to cquery in pyrefly.bxl

cbarrete:
`uquery.owner()` returns unconfigured targets, and passing those into `cquery.deps()` relies on implicit configuration. That usage is flagged by the `bxl_unconfigured_target_in_cquery` soft error, which is promoted to a hard error on open-source builds (`error_on_oss: true`), so the sourcedb query always fails for OSS users unless a target platform is passed explicitly:

```
error: Unconfigured target with label `...` was passed into cquery.
Targets passed into cquery should be configured (recommendation is
to use `ctx.target_universe()`).
```

This can be worked around by passing `--target-platforms`, but we can fix this properly by `ctx.target_universe()` as the error message recommends.

### DIFF
--- a/prelude/python/sourcedb/pyrefly.bxl
+++ b/prelude/python/sourcedb/pyrefly.bxl
@@ -85,7 +85,7 @@ def _append_or_default_if_module(
 def _list_targets(ctx: bxl.Context) -> bxl.ConfiguredTargetSet:
     cquery = ctx.cquery()
     uquery = ctx.uquery()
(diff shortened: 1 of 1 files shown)


cbarrete:
@connernilsen @samwgoldman Hello guys, could I get some eyes on this, when you have a moment? Thank you!

dtolnay (approved):
Thanks!

PR #1465, opened 2026-08-17
java: use exec_dep properly

lf-:
Lots of tools here were broken under cross compilation: they'd get the target-platform python then blow up when executed.


dtolnay (approved):
Thanks!

PR #1466, opened 2026-08-17
python: fix exec platform handling

lf-:
- python_bootstrap_toolchain.interpreter -> exec_dep. Every python_bootstrap_binary in the prelude is a build tool.
- python_toolchain.compile -> exec_dep. It is read by the bytecode compile action in prelude/python/compile.bzl.
- Split host_interpreter out of interpreter. make_py_package already passes them to the packaging tool as separate --python/--host-python flags, so this is a distinction upheld by the rest of the prelude.

  We introduce a macro to retain source compatibility.
- Fix system_python_bootstrap_toolchain to properly consider the exec platform when determining the name of the interpreter, rather than doing it based on target platform.

related to #1465: this sort of independently fixes the python binaries to work correctly (by forcing them to use the exec platform's interpreter, which is the only correct choice for bootstrap binaries), though it's still correct to transition them to exec configuration as that PR does.

### DIFF
--- a/prelude/toolchains/python.bzl
+++ b/prelude/toolchains/python.bzl
@@ -8,6 +8,11 @@
 
 load("@prelude//:prelude.bzl", "native")
 load(
+    "@prelude//os_lookup:defs.bzl",
+    "Os",
+    "OsLookup",
+)
+load(
     "@prelude//python:python_wheel_toolchain.bzl",
     "PythonWheelToolchainInfo",
 )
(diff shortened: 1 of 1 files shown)


lf-:
~~wait i need to eat my hat. brb draft mode.~~

hat uneaten!! i learned something about exec platform resolution today (owie my head. i gotta figure out how to understand this by some means other than just following rules), namely that interpreters are more libc-shaped than they are compiler-shaped ..

dtolnay (approved):
Thanks!

PR #1467, opened 2026-08-18
Honor `buck2.infer_target_names` in queries

9999years:
Honor the `buck2.infer_target_names` setting in uquery, cquery, aquery, and BXL query evaluation. Thread `InferTargetNames` through the shared query literal parser.

NB: This may be a bug aside from the setting; `buck build` and most other commands already interpret `//foo` as `//foo:foo`, so it's a bit odd that `buck uquery` and similar wouldn't.

See 088c75c7e36805df99c3de29062baa95db700b8b (#1004 / Differential Revision D112694484).

### DIFF
--- a/app/buck2_query_impls/src/dice.rs
+++ b/app/buck2_query_impls/src/dice.rs
@@ -47,6 +47,7 @@ use buck2_fs::paths::abs_norm_path::AbsNormPathBuf;
 use buck2_fs::paths::file_name::FileNameBuf;
 use buck2_hash::BuckMutMap;
 use buck2_hash::buck_indexset;
+use buck2_interpreter_for_build::interpreter::context::HasInterpreterContext;
 use buck2_node::load_patterns::MissingTargetBehavior;
 use buck2_node::load_patterns::load_patterns;
 use buck2_node::nodes::configured::ConfiguredTargetNode;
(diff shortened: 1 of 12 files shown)


app/buck2_query_impls/src/dice.rs
```diff
@@ -238,6 +244,29 @@ impl DiceQueryData {
     }
 }
 
+pub(crate) async fn infer_target_names_from_config(
```
lf-: wiggles i think this is a duplication, one min. i have a better one.

lf- (changes requested):

app/buck2_query_impls/src/dice.rs
lf-: pushed it!

8Keep:
I would argue that we don't want this behind the buckconfig, because we already have this syntax for command line and queries are command line. The buckconfig was really just for using the syntax in build files, which we can't enable inside meta right now for historical reasons.

lf-:
hmmm. that's a very interesting concept. what do you think of attrs.query and bxl queries? ie, i was under the impression Meta doesn't like the implicit eponymous target behaviour if it's anywhere in a file that's checked in.

for bxl i can see a strong argument it should nonetheless always infer target names because it may wind up in ui. for attrs.query, i dunno! maybe it's not that important.

PR #1474, opened 2026-08-25
rust: disable uploading incremental compilation build artifacts

davidbarsky:
I enabled incremental compilation with Rust and noticed that incremental compilation artifacts were being uploading, negating (and worse, regressing!) build times due to cache uploads. I then went down a rabbit hole, which... well, you can see in the comment I wrote.

### DIFF
--- a/prelude/rust/build.bzl
+++ b/prelude/rust/build.bzl
@@ -1763,7 +1763,18 @@ def _rustc_invoke(
 
     # None defers the choice to the `buck2.default_allow_cache_upload` config; an
     # explicit False overrides it. Actions without a preference pass None.
-    if allow_cache_upload:
+    if incremental_enabled:
+        # Incremental compilation should not publish any action output to a shared cache:
+        # 1. the incremental compilation state is not useful for any other user. unfortunately,
+        #    there is no mechanism in buck2 that allows for uploading part of an action output
+        #    and not another.
+        # 2. even if there were, the rlib is not byte-for-byte reproducible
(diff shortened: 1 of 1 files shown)


8Keep (approved):

8Keep:
We don't run into this internally because we don't upload any artifacts from user builds, only trunk builds. We consider user builds unsafe to upload because someone could modify outputs before uploading, we can't trust users. User artifacts are uploaded as part of running them on RE, because other users will only get those artifacts if they have the exact same inputs (inputs is part of the key for the remote action). So, user builds populate caches only through remote actions, and trusted CI machines are the only ones who can upload local only actions.

PR makes sense, just wanted to clarify why we do not hit this internally.

davidbarsky:
> We don't run into this internally because we don't upload any artifacts from user builds, only trunk builds. We consider user builds unsafe to upload because someone could modify outputs before uploading, we can't trust users. User artifacts are uploaded as part of running them on RE, because other users will only get those artifacts if they have the exact same inputs (inputs is part of the key for the remote action). So, user builds populate caches only through remote actions, and trusted CI machines are the only ones who can upload local only actions.

Yup, I remember that constraint! I _just_ ordered a Mac Mini for builds, so within a week, I should be able to sidestep this issue entirely and hopefully prevent someone else from hitting this issue! I think the second case (splitting regular rlibs from incremental build artifacts) mid-build would also be solved, by construction, if users can't upload build artifacts.

PR #1478, opened 2026-08-27
re_grpc: retry io errors wrapped in h2::Error

vlad-ivanov-name:
h2::Error does not override source() in its Error impl, so an io::Error inside it never appears in the error chain and is_retryable misses it. tonic also maps these errors to Code::Unknown. A TCP reset during a BatchReadBlobs response body therefore failed the build with zero retries. Downcast to h2::Error and check the inner io::Error directly.


scottcao (approved):

PR #1496, opened 2026-09-09
Sort `dynamic analysis` actions last in `superconsole` output

9999years:
This shows what's actually executing, rather than listing a ton of dynamic analysis actions at the top; we implement this change by marking dynamic analysis actions as 'boring'.


scottcao (approved):

PR #1501, opened 2026-09-10
tui: rename "tasks" to "actions"

cbarrete:
The TUI lists _actions_ being run, "tasks" are not a thing in this context.


cbarrete:
@scottcao I've just realized that this is exactly the opposite of what you did in https://github.com/facebook/buck2/commit/704fe0539afb8dd117f0e0ef5ad35cef8f45b0c8.

Can you clarify why we're not using the term action anymore? What "tasks" can run that aren't actions?

PR #1502, opened 2026-09-11
cargo: allocative in autocargo (#1502)

JakobDegen:
Summary:

D108568489, revived now that the starlark-rust repository no longer copies
this crate. The rule for the dependency specs, which the rest of the
published crates follow too: a dependency is written out with
`workspace = false` where the handwritten manifest asked for fewer features
than fbsource enables (here every optional dependency, and `tokio` with only
`sync`), and inherits from the workspace otherwise. Version floors move to
fbsource's pins, which is what buck builds against anyway; the handwritten
floors had drifted (`parking_lot = "0.11.2"` for a crate built against 0.12).

The dependency on `allocative_derive` is a caret requirement rather than the
`=` pin the handwritten manifest had, since `dependencies_override` cannot
express an exact version today; the two are always released together.

Reviewed By: 8Keep

Differential Revision: D119623889


8Keep (approved):

Issue #1510, opened 2026-09-18
InternalRunnerTestInfo: let a test listing report a result without running the test


k-cross:
## Problem

A rule that returns `InternalRunnerTestInfo` lists its tests, and Buck2 runs an action for every entry returned. Ignored tests have no way to be excluded from running. Leaving skipped tests out of the listing is worse: `buck2 test` shows nothing for it, which is indistinguishable from a test that doesn't exist but if it's listed it costs compute. The runner must run the process to determine the action's result when the listing could have told it directly. Under remote execution that means transferring the test binary for a test that is never run.

## Current behavior

Checked against `4f46864e49`:

- A listing entry is exactly `{name, filter}` — `TestListingEntry` in `app/buck2_test_api/src/data.rs`.
- `parse_test_listing_output` in `app/buck2_build_api/src/interpreter/rule_defs/provider/builtin/internal_runner_test_info.rs` reads those two keys and silently ignores any others.
- `run_internal_test` in `app/buck2_test/src/internal_runner.rs` reports every entry as discovered, then calls `execute2` for every one of them, with no exceptions.
- A result otherwise reaches Buck2 only through `parse_test_result`, after an action has run, or from the statuses the runner synthesizes itself: `TIMEOUT`, `OMITTED` on cancellation, and `FATAL`.

## Proposal

Let a listing entry carry the result Buck2 should report for it, in place of running it:

```python
{"name": "...", "filter": "...", "status": "SKIP", "message": "ignored"}
```

Entries are still reported as discovered and their its result is reported directly. No action runs for it and entries without `status` and `message` remain unaffected.

Since `parse_test_listing_output` ignores unknown keys today, a runner can emit these keys before the change lands and get the current behavior on older Buck2 versions.

## Questions

- Which statuses may a listing preset?
    - `SKIP` is the one needed here. A listing declaring that a test passed or failed without running it seems undesirable. Accepting more later is backwards compatible should it be needed.
- Should `filter` become optional for an entry with a preset result, since nothing uses it?
    - A runner has to send it for older Buck2 versions anyway, and relaxing it later is backwards compatible.

Two things that looked like they might interact, but don't:
- Listing caching: the cache holds the listing command's output, and `parse_test_listing` runs on it every time, so a preset result is never stale.
- Include and exclude filters: `--include` and `--exclude` match labels and apply to the whole target before listing. Test-name arguments after `--` only apply to the external test executor which never see the targets.

## Context

[...]

PR #1511, opened 2026-09-18
Internal listings to remove needless compute

k-cross:
A `parse_test_listing` entry can now carry a result Buck2 should report for it so runners can list tests they know will not run:

```python
{"name": "...", "filter": "...", "status": "SKIP", "message": "ignored"}
```

Entries are still reported as discovered, but its result is reported directly so no action runs for it. Entries without `status` and `message` behave exactly as before so Buck2 versions without this change ignore these keys; runners can emit them today and still get the old behavior.

Only `SKIP` is accepted since a listing declaring that a test passed or failed without running it seems bad. Accepting more statuses later is backwards compatible though. The `filter` stays required.

Why? `#[ignore]`d Rust tests under a `nextest` based runner are listed so that `buck2 test` reports them as skipped rather than hiding them but skipped tests still have compute and network costs that could be avoided.

## Stacked on #1461

The first commit is #1461. Any e2e test that uses `InternalRunnerTestInfo` needs it, since without it `buck2 test` fails while tearing the run down. The second commit is this change and I'll rebase once #1461 lands.

## Change Highlights

- `TestListingEntry` gains an optional result
- `parse_test_listing_output` reads the optional `status` and `message` keys, parsing `status` with `TestStatus::parse` the way `parse_test_result_output` already does
- `run_internal_test` reports entries that carry a result with `report_test_result` and schedules actions only for the rest

## Testing

- New unit tests alongside the existing `parse_test_listing_output` tests
- A new e2e test `tests/core/test/test_internal_runner_listing.py`
- Manually, with a nextest-based runner over a sample project on my local machine

### DIFF
--- a/app/buck2_test/src/command.rs
+++ b/app/buck2_test/src/command.rs
@@ -137,14 +137,24 @@ struct TestOutcome {
 impl TestOutcome {
     fn exit_code(&self) -> buck2_error::Result<i32> {
         self.executor_report
-            .exit_code
+            .exit_code()
             .internal_error("Test executor did not provide an exit code")
     }
 }
 
+/// Exit code for a run in which tests failed;
+/// matches `RunVerdict::Fail` in the OSS test runner
+const TESTS_FAILED_EXIT_CODE: i32 = 32;
+
 #[derive(Default)]
 struct ExecutorReport {
-    exit_code: Option<i32>,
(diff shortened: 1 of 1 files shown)


k-cross:
This resolves issue #1510

